PR #7, opened 2018-02-05
Cleanup leftover NVTs.

kroosec:
Remove NVT metadata from nvticache when it is deleted from the plugins
folder.


jjnicola:
for some reason the check failed. I merge this PR on my local working copy and it did work. I think it is some miss configuration in circleci.

PR #28, opened 2018-03-06
Get the feed version from nvticache.

kroosec:
Also use NVTICACHE_STR for the string that identifies the cache DB.


kroosec:
Depends on https://github.com/greenbone/gvm-libs/pull/35

PR #56, opened 2018-04-09
Add separator variable and modify prepend format to include it.

jjnicola:


jjnicola:
Related to PR62 in gvm-libs

Issue #60, opened 2018-04-10
master: Remove OPENVAS_CACHE_DIR / cache_folder 


cfi-gb:
The commits https://github.com/greenbone/gvm-libs/commit/3cb79063b3b1b56376c6846868e97d9982b94c3a and https://github.com/greenbone/openvas-scanner/commit/358628c11bfafbe40cef096629927bde9c482582 moved the nvti cache from $prefix/var/cache/openvas into redis 

There are some remains left for "cache_folder" and "OPENVAS_CACHE_DIR" which seems to be unused now and could be removed.

cfi-gb:
Fixed with https://github.com/greenbone/openvas-scanner/pull/69 so closing here.

PR #15, opened 2018-02-19
Move full nvticache to Redis 

kroosec:

PR #69, opened 2018-04-16
Remove OPENVAS_CACHE_DIR / cache_folder.

jjnicola:
Solve issue #60 

Remove the remaining "cache_folder" and "OPENVAS_CACHE_DIR" which are not in use since the
nvticache has been moved to redis.


cfi-gb:
Backported to the openvas-scanner-5.1 branch with https://github.com/greenbone/openvas-scanner/pull/178 and to openvas-libraries-9.0 with https://github.com/greenbone/gvm-libs/pull/125 after https://github.com/greenbone/openvas-scanner/pull/15 and https://github.com/greenbone/gvm-libs/pull/25 got backported to both branches as well.

PR #82, opened 2018-04-25
Solve TODO.

jjnicola:
Simplify code using g_strchomp().
Add function documentation.


kroosec:
IMO better simplify the code further, by using a glib function such as g_strchomp()

PR #89, opened 2018-05-05
Updates to the example redis configuration.

cfi-gb:
This PR:

- Moves the two example redis configs into a subdir of doc/
- Added the plain example configs from https://redis.io/topics/config for newer versions
- Updated the cmake files for this change
- Updated the INSTALL for this change
- Added `<install-prefix>` to the redis examples in INSTALL

Once it was decided to accept this PR i will create a follow-up PR for the new 2.8-4.0 redis configs and apply the changes from the 2.4 and 2.6 configs so our modifications within this files can be followed in the commit history.


cfi-gb:
Follow-up PR in #93

PR #106, opened 2018-06-02
Remove redundant struct nasl_threads_args.

kroosec:
Simplify functions prototypes, and fill script_infos accordingly.


janowagner:
The doxygen headers are not all updated to new proto. For example exec_nasl_script().

kroosec:
Fixed function docs changes.

PR #108, opened 2018-06-04
Add scan_id command line option and get scan prefs from redis.

jjnicola:
A new command line option 'scan-start' is added and receive a scan_id.
If no scan_id is passed, for each new task is autogenerated.
The scan_id is used to identify a Redis DB where the scan preferences should be
stored if the scanner has been started with the --scan-start option.
Then, beforethe attack, the preferences are taken from redis and stored in the
globals structure. After that the attack continues as before.

* misc/scanneraux.h (scan_globals): Add scan_id and otp_scan members.

* src/attack.c (attack_start): Add the item "internal/scan_id" into the KB
which stores the scan_id. New global_scan_type global variable.
(is_otp_scan, set_scan_type): New functions.
(error_message_to_client): Check the scan type (OTP or OSP) before trying
to send a message to the client. In OSP type there is no socket, since the
information is forwarded through redis.

* src/hosts.c: (forward, forward_status): Check for the scan type before
forwarding a message.
(host_rm): Check for scan type before deleting the kb.
(hosts_read_client): Only check for a socket if it is an OTP scan type.

* src/ntp.c (__ntp_timestamp_scan_host):  Check for the scan type before
forwarding a message

* src/openvassd.c: (load_scan_preferences): New function. Search the KB
with the corresponding scan_id, get the preferences and store them into
globals structures.
(handle_client): Call load_scan_preferences
(scanner_thread): Check if the scan_id was passed as command line option.
In other case, generate the scan_id.
(main): Add new 'scan-start' command line option. If it is set, start a scan
without global socket and without wait for a client.
(start_single_task_scan): Start a single scan using the preferences from redis.

### DIFF
--- a/src/attack.c
+++ b/src/attack.c
@@ -129,9 +129,10 @@ static void
 error_message_to_client (int soc, const char *msg, const char *hostname,
                          const char *port)
 {
-  send_printf
-   (soc, "SERVER <|> ERRMSG <|> %s <|>  <|> %s <|> %s <|>  <|> SERVER\n",
-    hostname ?: "", port ?: "", msg ?: "No error.");
+  if (is_otp_scan ())
+    send_printf
+      (soc, "SERVER <|> ERRMSG <|> %s <|>  <|> %s <|> %s <|>  <|> SERVER\n",
+       hostname ?: "", port ?: "", msg ?: "No error.");
 }
 
 static void
@@ -595,9 +596,12 @@ attack_start (struct attack_start_args *args)
   kb_t kb = args->host_kb;
   gvm_hosts_t *hosts_allow, *hosts_deny;
   gvm_hosts_t *sys_hosts_allow, *sys_hosts_deny;
+  char key[1024];
 
   nvticache_reset ();
   gettimeofday (&then, NULL);
+
+  kb_item_add_str (kb, "internal/scan_id", globals->scan_id, 0);
 
   /* The reverse lookup is delayed to this step in order to not slow down the
    * main scan process eg. case of target with big range of IP addresses. */
@@ -635,6 +639,9 @@ attack_start (struct attack_start_args *args)
     g_message ("Testing %s [%d]", ip_str, getpid ());
   g_free (hostnames);
   attack_host (globals, &hostip, args->host->vhosts, sched, kb, net_kb);
+
+  snprintf (key, sizeof (key), "internal/%s", globals->scan_id);
+  kb_item_add_str (kb, key, "finished", 0);
 
   if (!scan_is_stopped () && !all_scans_are_stopped ())
     {
--- a/src/hosts.c
+++ b/src/hosts.c
@@ -64,7 +64,6 @@ static struct host *hosts = NULL;
 static int g_soc = -1;
 static int g_max_hosts = 15;
 
-
 /*-------------------------------------------------------------------------*/
 
 static int
@@ -92,6 +91,10 @@ forward_status (struct host *h, int out)
 {
   char *status = NULL, *buf = NULL;
 
+  /* Send the message to the client only if it is a OTP scan. */
+  if (!is_otp_scan ())
+    return 0;
+
   status = kb_item_pop_str (h->host_kb, "internal/status");
   if (!status)
     return 0;
@@ -110,6 +113,10 @@ forward_status (struct host *h, int out)
 static int
 forward (struct host *h, int out)
 {
+  /* Send the message to the client only if it is a OTP scan. */
+  if (!is_otp_scan ())
+    return 0;
+
   forward_status (h, out);
   while (1)
     {
@@ -154,8 +161,10 @@ host_rm (struct host *h)
   if (h->prev != NULL)
     h->prev->next = h->next;
 
+  if (is_otp_scan ())
+    kb_delete (h->host_kb);
+
   g_free (h->name);
-  kb_delete (h->host_kb);
   g_free (h->ip);
   g_free (h);
 }
@@ -360,7 +369,7 @@ hosts_read_client (struct scan_globals *globals)
 int
 hosts_read (struct scan_globals *globals)
 {
-  if (hosts_read_client (globals) < 0)
+  if (hosts_read_client (globals) < 0 && is_otp_scan ())
     {
       hosts_stop_all ();
       g_debug ("Client abruptly closed the communication");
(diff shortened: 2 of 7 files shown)


src/attack.c
```diff
@@ -599,6 +621,8 @@ attack_start (struct attack_start_args *args)
   nvticache_reset ();
   gettimeofday (&then, NULL);
 
+  kb_item_add_str (kb, "internal/scan_id", globals->scan_id, strlen (globals->scan_id));
```
kroosec: No need for strlen(), just 0.

src/attack.c
```diff
@@ -636,6 +660,10 @@ attack_start (struct attack_start_args *args)
   g_free (hostnames);
   attack_host (globals, &hostip, args->host->vhosts, sched, kb, net_kb);
 
+  char *buf = NULL;
```
kroosec: Var declaration goes to the start of code block.

src/attack.c
```diff
@@ -636,6 +660,10 @@ attack_start (struct attack_start_args *args)
...
+  buf = g_strdup_printf ("internal/%s", globals->scan_id);
```
kroosec: Memleak

src/attack.c
```diff
@@ -636,6 +660,10 @@ attack_start (struct attack_start_args *args)
...
+  kb_item_add_str (kb, buf, "finished", 8);
```
kroosec: 0 instead of 8.

src/attack.c
```diff
@@ -832,7 +860,7 @@ check_kb_access (int soc)
   kb_t kb;
 
   rc = kb_new (&kb, prefs_get ("kb_location"));
-  if (rc)
+  if (rc && is_otp_scan())
```
kroosec: is_otp_scan ().

src/attack.c
```diff
@@ -129,9 +150,10 @@ static void
 error_message_to_client (int soc, const char *msg, const char *hostname,
                          const char *port)
 {
-  send_printf
-   (soc, "SERVER <|> ERRMSG <|> %s <|>  <|> %s <|> %s <|>  <|> SERVER\n",
-    hostname ?: "", port ?: "", msg ?: "No error.");
+  if (is_otp_scan())
```
kroosec: is_otp_scan ().

src/attack.c
kroosec: Also, is this needed ? Given that you already call is_otp_scan() in error_message_to_client().

src/openvassd.c
```diff
@@ -377,25 +379,82 @@ reload_openvassd ()
...
+  char *key;
+  kb_t kb;
+  struct kb_item *res = NULL;
+
+  g_debug ("Start loading scan preferences.");
+  if (!scan_id)
+    return -1;
+  key = g_strdup_printf ("%s/scanprefs", scan_id);
+  kb = kb_find (prefs_get ("kb_location"), key);
+  if (!kb)
```
kroosec: Again memleak. Better use buffer on stack ie. something like key[1024].

src/openvassd.c
```diff
@@ -377,25 +379,82 @@ reload_openvassd ()
...
+load_scan_preferences (const char *scan_id)
+{
+  char *key;
+  kb_t kb;
+  struct kb_item *res = NULL;
+
+  g_debug ("Start loading scan preferences.");
+  if (!scan_id)
+    return -1;
+  key = g_strdup_printf ("%s/scanprefs", scan_id);
```
kroosec: IMO better add internal/ prefix for consistency.

src/openvassd.c
```diff
@@ -377,25 +379,82 @@ reload_openvassd ()
...
+  res = kb_item_get_all (kb, key);
+  if (!res)
+    return -1;
+  g_free (key);
+
+  while (res)
+    {
+      gchar **pref = g_strsplit (res->v_str, "|||", 2);
+      if (pref[0])
+        prefs_set (pref[0], pref[1] ? : "");
```
kroosec: ?:

src/openvassd.c
```diff
@@ -404,8 +463,29 @@ scanner_thread (struct scan_globals *globals)
   int opt = 1, soc;
 
   nvticache_reset ();
-  soc = globals->global_socket;
-  proctitle_set ("openvassd: Serving %s", unix_socket_path);
+
+   if (globals->otp_scan == 1 && !global_scan_id)
+    {
+      globals->scan_id = (char *) gvm_uuid_make();
```
kroosec: gvm_uuid_make ().

src/openvassd.c
```diff
@@ -816,6 +888,37 @@ gcrypt_init ()
...
+#ifdef OPENVASSD_GIT_REVISION
+  g_message ("openvassd %s (GIT revision %s) started",
+             OPENVASSD_VERSION,
+             OPENVASSD_GIT_REVISION);
+#else
+  g_message ("openvassd %s started", OPENVASSD_VERSION);
+#endif
+
+  struct scan_globals *globals;
+  int ret = 0;
```
kroosec: Var declarations to the top of code block.

src/openvassd.c
```diff
@@ -928,6 +1034,14 @@ main (int argc, char *argv[])
 
   if (init_openvassd (config_file))
     return 1;
+
+  if (scan_id)
+    {
+      global_scan_id = g_strdup (scan_id);
+      start_single_task_scan();
```
kroosec: start_single_task_scan ();

src/hosts.c
```diff
@@ -64,6 +64,7 @@ static struct host *hosts = NULL;
 static int g_soc = -1;
 static int g_max_hosts = 15;
 
+extern int global_scan_type;
```
kroosec: Why not use is_otp_scan() / set_scan_type() ?

src/ntp.c
```diff
@@ -392,6 +392,7 @@ __ntp_timestamp_scan (int soc, char *msg)
   return 0;
 }
 
+extern int global_scan_type;
```
kroosec: Same.

src/openvassd.c
```diff
@@ -404,8 +463,29 @@ scanner_thread (struct scan_globals *globals)
...
+      if (soc < 0)
+        goto shutdown_and_exit;
+
+      setsockopt (soc, SOL_SOCKET, SO_KEEPALIVE, &opt, sizeof (opt));
+      globals->global_socket = soc;
+
+      if (comm_init (soc) < 0)
+        exit (0);
+    }
+  else
```
kroosec: While at it, remove the useless { }

PR #120, opened 2018-06-13
Unify tcp and udp required ports checking routines

kroosec:
Also fixes cases where NVT has "Services/udp/unknown" in its required
udp ports.

Fix for Mantis #80572

### DIFF
--- a/src/plugs_req.c
+++ b/src/plugs_req.c
@@ -50,25 +50,21 @@ extern int kb_get_port_state_proto (kb_t, int, char *);
  * @return Whether a port in a port list is closed or not.
  */
 static int
-get_closed_ports (kb_t kb, char *ports_list)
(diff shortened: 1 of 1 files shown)


cfi-gb:
How to test the ``Services/udp/unknown`` part:

1. Install OpenVPN on a target system (OpenVPN is not mandatory, could be also done with pdns-server or dnsmasq)
2. Configure OpenVPN to listen on a non-default port like 555/udp
3. Make sure that OpenVPN isn't using a tls-auth ta.key 0
4. Create a Portlist including 555/udp and assign it to a target and the target to a task
5. Enable "Service discovery on non-default UDP ports (slow)" in global_settings.nasl (Global variable settings, OID: 1.3.6.1.4.1.25623.1.0.12288)

OpenVPN should be detected now on the non-standard port 555/udp.

PR #124, opened 2018-06-15
Add default value for test_empty_vhost.

kroosec:
Also document this preference.


cfi-gb:
PR to defined the test_empty_vhost scanner option in GSA as a radio button: https://github.com/greenbone/gsa/pull/732

PR #131, opened 2018-06-29
added systemd units

GIJack:
As requested. Here is my systemd unit for openvas-scanner.

There is a .service file. the default file gets renamed to openvas-scanner and put in either /etc/default or /etc/sysconfig depending on the distro

there is also a file for logrotated if rsyslog is being used on top of this

### DIFF
--- a/systemd/openvas-scanner.default
+++ b/systemd/openvas-scanner.default
@@ -1,0 +1,27 @@
+# Options to pass to the openvassd daemon
+OPTIONS=""
+
+# Set to yes if plugins should be automatically updated via a cron job
+auto_plugin_update=no
+
+# Notify OpenVAS scanner after update by seding it SIGHUP?
+notify_openvas_scanner=yes
+
+# Method to use to get updates. The default is via rsync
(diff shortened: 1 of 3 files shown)


cfi-gb:
Just to note that the above doesn't look unanswered: This is currently discussed in https://github.com/greenbone/gvm/pull/142

shibumi:
Hi, I would like to get this merged. I am the Arch Linux Maintainer for OpenVAS. I really don't think that this is just a 'packaging aspect'. One reason why the linux distributions have adopted `systemd` was that init scripts would be the same across all distributions. Furthermore upstream is now able to manage these systemd service files. If you include systemd service files in your project you can:

1. r**educe bug reports**, because you know that OpenVAS is started on all systems the same way. If something non default happens, you can close the bug report and refer to the systemd service files that are supported by upstream.
2. **harden your services**. You are able to harden your security application against malicious behavior using cgroups and namespaces (both supported by the linux kernel). If all distributions would do harden the service files by themselves you can expect more work for your team, because of more bug reports in the future regarding system startup or weird behavior of your application.

PR #132, opened 2018-07-02
Fix issue reported by Valgrind.

jjnicola:
Fix "Invalid write of size 1" allocating one byte more for the terminating nul.


kroosec:
The whole code in this file stinks. :/ I will send a PR to clean it up a bit.

PR #161, opened 2018-09-06
Remove unused nasl_file_*() function.

kroosec:
Only nasl_file_stat() is used by some plugins.


kroosec:
Haven't seen that usage in the mentioned .inc file. Closing the PR, however:

> I'm a bit skeptical to remove functions which might be used / needed in the future again.

In cases where said functions are hundreds of LOCs of old GPLv2-only code which weren't used in ages, my opinion is that it is pretty sensible to remove them.

PR #171, opened 2018-09-18
Fix parameter pollution in certain nasl functions.

kroosec:
Some NASL functions use get_*_var_by_name() instead of
get_*_local_var_by_name(), which "climbs" in the context when searching
for a variable ie. if a parameter 'foo' is not provided, the function
would use the variable 'foo' if found.

Unify the behaviour by replacing all get_*_local_var_by_name()
with get_*_var_by_name() calls, and change the later to not climb
in the nasl context when searching for the variable.

To reproduce:
------------------------
hostname = 'foo';
source = 'bar';
add_host_name();
display(get_host_name() + '|' + get_host_name_source());
------------------------


cfi-gb:
Direct ref to the gvm-libs PR: https://github.com/greenbone/gvm-libs/pull/121

PR #174, opened 2018-09-21
Improve message log.

jjnicola:
Print the nvt name in case of "unsupported transport layer".

### DIFF
--- a/misc/network.c
+++ b/misc/network.c
@@ -1001,7 +1001,7 @@ open_stream_connection_ext (struct script_infos *args, unsigned int port,
 
     default:
       g_message ("open_stream_connection_ext(): unsupported transport"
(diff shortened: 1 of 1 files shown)


misc/network.c
```diff
@@ -1001,7 +1001,7 @@ open_stream_connection_ext (struct script_infos *args, unsigned int port,
 
     default:
       g_message ("open_stream_connection_ext(): unsupported transport"
-                 " layer %d", transport);
+                 " layer %d executing %s", transport, args->name);
```
cfi-gb: As the "transport" layer variable is passed by the nvt and not executed by the NVT maybe replace

"executing"

with

"passed by". Besides this the change looks fine to me.

PR #184, opened 2018-09-29
Add expand_vhosts preference.

kroosec:
In order to control whether the target host's vhosts list is expanded
through reverse-lookups and nasl add_host_name() calls.

### DIFF
--- a/src/attack.c
+++ b/src/attack.c
@@ -640,7 +640,8 @@ attack_start (struct attack_start_args *args)
 
   /* The reverse lookup is delayed to this step in order to not slow down the
    * main scan process eg. case of target with big range of IP addresses. */
-  gvm_host_add_reverse_lookup (args->host);
+  if (prefs_get_bool ("expand_vhosts"))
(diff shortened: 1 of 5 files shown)


src/attack.c
```diff
@@ -640,7 +640,8 @@ attack_start (struct attack_start_args *args)
 
   /* The reverse lookup is delayed to this step in order to not slow down the
    * main scan process eg. case of target with big range of IP addresses. */
-  gvm_host_add_reverse_lookup (args->host);
+  if (prefs_get_bool ("expand_vhosts"))
```
cfi-gb: The pref probably should be added / checked in 
https://github.com/greenbone/openvas-scanner/blob/b7cc8aa1d2fdfd8464110cb89dd99222cd064b3c/nasl/nasl.c#L344 as well

kroosec: Agree with the suggested changes to the doc file, and that openvas-nasl should match openvassd behavior when possible. Updated the PR accordingly.

PR #187, opened 2018-10-07
Use SIGUSR2 to check for new vhosts.

kroosec:
Instead of SIGUSR1, which is also used to stop scan host processes. SIGUSR2
is used for main scan processes only. Also use updated vhosts list adequately.


kroosec:
@bjoernricks This fixes the scan stopping issue you reported earlier.

src/ntp.c
```diff
@@ -100,6 +100,7 @@ ntp_parse_input (struct scan_globals *globals, char *input)
           break;
 
         case CREQ_STOP_WHOLE_TEST:
+          g_warning ("EEEEEEEEEEEEEEE");
```
cfi-gb: Debugging output

PR #196, opened 2018-10-18
Cleanup default prefs values.

kroosec:
These are now set in prefs_init() to be shared by both openvassd and
openvas-nasl.

Depends on https://github.com/greenbone/gvm-libs/pull/135

### DIFF
--- a/src/openvassd.c
+++ b/src/openvassd.c
@@ -127,28 +127,18 @@ static openvassd_option openvassd_defaults[] = {
   {"be_nice", "no"},
   {"log_whole_attack", "no"},
   {"log_plugins_name_at_load", "no"},
-  {"cgi_path", "/cgi-bin:/scripts"},
   {"optimize_test", "yes"},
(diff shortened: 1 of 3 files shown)


src/openvassd.c
```diff
@@ -138,17 +136,10 @@ static openvassd_option openvassd_defaults[] = {
   {"auto_enable_dependencies", "yes"},
   {"nasl_no_signature_check", "yes"},
```
cfi-gb: This could be a good chance to move this to the prefs.c as well so that the -X in openvas-nasl (https://github.com/greenbone/openvas-scanner/blob/v6.0%2Bbeta1/nasl/nasl.c#L207-L208) isn't required by default anymore. Not sure if this could make sense in a separate PR.

Rationale:
The most common use-case for openvas-nasl is the development of new VTs where no signature are available at all.

kroosec: @cfi-gb Agree. Updated the PRs accordingly.

PR #199, opened 2018-10-24
Don't parse or insert NVT copyright.

kroosec:
(Work in Progress / Waiting for Feature Specification validation.)

Depends on https://github.com/greenbone/gvm-libs/pull/139

### DIFF
--- a/src/comm.c
+++ b/src/comm.c
@@ -175,7 +175,7 @@ void
 send_plug_info (int soc, const char *oid)
 {
   int category;
-  char *name = NULL, *copyright = NULL, *family = NULL;
+  char *name = NULL, *family = NULL;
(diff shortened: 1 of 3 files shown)


src/comm.c
```diff
@@ -225,8 +219,8 @@ send_plug_info (int soc, const char *oid)
     }
 
   send_printf
-   (soc, "%s <|> %s <|> %d <|> %s <|> %s <|> %s <|> %s <|> %s <|> %s\n",
```
mattmundell: Hey Hani, how about removing the field entirely, so we don't have to parse the dummy field on Manager side?

In other words, remove the "<|>" too.

People aren't supposed to be using OTP directly, so only Manager would need to change, right?

kroosec: Hey matt, I agree. Updated the PR to remove it.

PR #205, opened 2018-11-08
Enable compatibility mode in GnuTLS priority string

cfi-gb:
The scanner is currently failing to connect to mis-configured devices via a SSL/TLS connection:

https://censys.io/ipv4?q=%22Energy+management%22+and+cisco

```
openvas-nasl -X -B -d -t $ find_service.nasl --kb="Ports/tcp/443=1" --debug-tls=10

[30215] (1) Peer's certificate does not allow digital signatures. Key usage violation detected.
*snip*
lib  misc-Message: 07:56:16.516: set key Transports/TCP/443 -> 1
lib  misc-Message: 07:56:16.551: replace key FindService/tcp/443/get_http -> HTTP/1.1 400 Bad Request
```

Running gnutls-cli shows the same reason:

```
gnutls-cli -p 443 $IP --insecure

|<1>| Peer's certificate does not allow digital signatures. Key usage violation detected.
*** Fatal error: Key usage violation in certificate has been detected.
*** handshake has failed: Key usage violation in certificate has been detected.
```

Specifying the %COMPAT keyword described below makes this to work:

```
gnutls-cli -p 443 80.193.70.19 --insecure --priority "NORMAL:%COMPAT"

- Handshake was completed

- Simple Client Mode:
```

> will enable compatibility mode. It might mean that violations of the protocols are allowed as long as maximum compatibility with problematic clients and servers is achieved. More specifically this string will tolerate packets over the maximum allowed TLS record, and add a padding to TLS Client Hello packet to prevent it being in the 256-512 range which is known to be causing issues with a commonly used firewall (see the %DUMBFW option).
> https://gnutls.org/manual/html_node/Priority-Strings.html

### DIFF
--- a/misc/network.c
+++ b/misc/network.c
@@ -1041,7 +1041,7 @@ open_stream_connection (struct script_infos *args, unsigned int port,
                         int transport, int timeout)
 {
   return open_stream_connection_ext (args, port, transport, timeout,
-                                     "NORMAL:+ARCFOUR-128");
+                                     "NORMAL:+ARCFOUR-128:%COMPAT");
 }
 
 /* Same as open_stream_auto_encaps but allows to force auto detection


cfi-gb:
Some additional notes on %COMPAT from https://gitlab.com/gnutls/gnutls/blob/master/NEWS:

> * Version 3.5.0 (released 2016-05-09)
> ** libgnutls: No longer tolerate certificate key usage violations for
   TLS signature verification, and decryption. That is GnuTLS will fail
   to connect to servers which incorrectly use a restricted to signing certificate
   for decryption, or vice-versa. This reverts the lax behavior introduced
   in 3.1.0, due to several such broken servers being available. The %COMPAT
   priority keyword can be used to work-around connecting on these servers.


> * Version 3.1.2 (released 2012-09-26)
> ** libgnutls: Prints warning on certificates that contain keys of
an insecure level. If the %COMPAT priority flag is not specified
the TLS connection fails.
> ** libgnutls: The %COMPAT keyword, if specified, will tolerate
key usage violation errors (they are far too common to ignore).

> * Version 2.11.1 (released 2010-09-14)
> ** libgnutls: The %COMPAT flag now allows larger records that violate the
TLS spec.

> * Version 2.7.6 (released 2009-02-27)
> ** gnutls-serv: No longer disable MAC padding by default.
Use --priority NORMAL:%COMPAT to disable MAC padding again.

cfi-gb:
Backport to openvas-libraries-9.0 in https://github.com/greenbone/gvm-libs/pull/149

Issue #166, opened 2018-09-12
5.1.3 Update Broke My Scanner


grayspacedef:
### Steps to reproduce
1. Not sure, just not working on my system...
2.
3.

### GVM versions
**gsa:** (gsad --version)
7.0.3
**gvm:** (gvmd --version)
7.0.3
**openvas-scanner:** (openvassd --version)
5.1.3
**gvm-libs:**
db revision 184?
**openvas-smb:**

Redis-server v4.0.11

### Environment
**Operating system:**
Kali 2.0
**Installation method / source:** (packages, source installation)
apt-get from Kali repositories
### Logfiles

```
OpenVAS_Logs.txt 1 / 17
[Thu Aug 30 15:26:23 2018][5880] Test complete
[Thu Aug 30 15:26:23 2018][5880] Total time to scan all hosts :
76865 seconds
[Thu Aug 30 16:31:40 2018][2762] Failed to initialize nvti cache.
[Thu Aug 30 16:33:08 2018][2819] openvassd 5.1.2 started
[Thu Aug 30 16:37:54 2018][2819] Received the Terminated signal
[Wed Sep 5 16:18:35 2018][2666] Failed to initialize nvti cache.
[Wed Sep 5 16:20:16 2018][2723] openvassd 5.1.2 started
[Thu Sep 6 01:28:59 2018][2723] Received the Terminated signal
[Tue Sep 11 15:13:56 2018][18437] openvassd 5.1.3 started
[Tue Sep 11 16:32:42 2018][18437] Received the Terminated signal
[Tue Sep 11 16:32:51 2018][29507] openvassd 5.1.3 started
SIGSEGV occured !
openvassd: Serving /var/run/openvassd.sock(sighand_segv+0x81)
[0x5618cd000e11]
/lib/x86_64-linux-gnu/libc.so.6(+0x35fc0)[0x7fb6508f1fc0]
/lib/x86_64-linux-gnu/libc.so.6(+0x3a850)[0x7fb6508f6850]
/usr/lib/x86_64-linux-gnu/libopenvas_base.so.
9(nvticache_get_category+0x30)[0x7fb65114e050]
openvassd: Serving /var/run/openvassd.sock(send_plug_info+0x1d)
[0x5618ccffb1bd]
openvassd: Serving /var/run/openvassd.sock(comm_send_nvt_info+0x18b)
[0x5618ccffb8ab]
[...]

PR #212, opened 2018-11-20
Handle case in which there is no nvt category set for an NVT.

jjnicola:
This can happen only with corrupted nvti cache. Therefore this error
is considered fatal and stop the running scan. Also, send an error message
to the manager.

Depends on greenbone/gvm-libs#151

### DIFF
--- a/src/attack.c
+++ b/src/attack.c
@@ -996,6 +996,14 @@ attack_network (struct arglist *globals, kb_t *network_kb)
   sched = plugins_scheduler_init
            (prefs_get ("plugin_set"), prefs_get_bool ("auto_enable_dependencies"),
             network_phase);
+  if (sched == NULL)
+    {
+      error_message_to_client (global_socket,
+                               "Failed to initialize the plugins scheduler.",
+                               NULL, NULL);
+      return;
+    }
+
 
   max_hosts = get_max_hosts_number ();
   max_checks = get_max_checks_number ();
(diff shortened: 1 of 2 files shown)


jjnicola:
Fix greenbone/openvas-scanner#166 for openvas-scanner-5.1

cfi-gb:
Related community report: https://community.greenbone.net/t/nvt-cache-table-in-db-not-getting-generated-with-openvas9-7-0-3-2bionic/878

src/attack.c
```diff
@@ -996,6 +996,17 @@ attack_network (struct arglist *globals, kb_t *network_kb)
...
            (prefs_get ("plugin_set"), prefs_get_bool ("auto_enable_dependencies"),
             network_phase);
+  if (sched == NULL)
+    {
+      error_message_to_client (global_socket,
+                               "An NVT oid has not category assigned. This is "
+                               "considered a fatal error, since the NVTI Cache "
+                               "structure stored in Redis is out dated or "
+                               "corrupted.", NULL,
+                               NULL);
```
kroosec: This error message is too specific to one error case (missing nvt category.)

Something like "Failed to initialize the plugins scheduler." is better fit IMO.

jjnicola: I agree. Changed.

src/pluginscheduler.c
```diff
@@ -500,6 +500,15 @@ plugins_scheduler_fill (plugins_scheduler_t sched)
       int category;
 
       category = nvticache_get_category (element->data);
+      if (category < 0)
+        {
+          log_write ("The NVT with oid %s has not category assigned. This is "
```
rgrewe: The "not" should be a "no"?

Issue #218, opened 2018-12-05
openvas-scanner-6.0-beta2 don't compile


tuxmaster5000:
### Expected behavior
Working compiler run
### Current behavior
the build fails

### GVM versions

**gvm-libs:**
gvm-libs 1 beta2

### Environment
**Operating system:**
CentOS 7.6
 
Build error:
In file included from /usr/include/arpa/inet.h:22:0,
                 from /builddir/build/BUILD/openvas-scanner-6.0-beta2/nasl/nasl_builtin_synscan.c:34:
/builddir/build/BUILD/openvas-scanner-6.0-beta2/nasl/nasl_builtin_synscan.c: In function 'extractack':
/builddir/build/BUILD/openvas-scanner-6.0-beta2/nasl/nasl_builtin_synscan.c:429:23: error: 'struct tcphdr' has no member named 'th_ack'
  ret = htonl(ntohl(tcp->th_ack) - 1);
                       ^
In file included from /usr/include/bits/byteswap.h:35:0,
                 from /usr/include/endian.h:60,
                 from /usr/include/bits/waitstatus.h:64,
                 from /usr/include/stdlib.h:42,
                 from /builddir/build/BUILD/openvas-scanner-6.0-beta2/nasl/nasl_builtin_synscan.c:33:
/builddir/build/BUILD/openvas-scanner-6.0-beta2/nasl/nasl_builtin_synscan.c: In function 'extractsport':
/builddir/build/BUILD/openvas-scanner-6.0-beta2/nasl/nasl_builtin_synscan.c:446:18: error: 'struct tcphdr' has no member named 'th_sport'
  return ntohs(tcp->th_sport);
                  ^
/builddir/build/BUILD/openvas-scanner-6.0-beta2/nasl/nasl_builtin_synscan.c: In function 'issynack':
/builddir/build/BUILD/openvas-scanner-6.0-beta2/nasl/nasl_builtin_synscan.c:462:13: error: 'struct tcphdr' has no member named 'th_flags'
   return tcp->th_flags == (TH_SYN | TH_ACK);
             ^
/builddir/build/BUILD/openvas-scanner-6.0-beta2/nasl/nasl_builtin_synscan.c:462:28: error: 'TH_SYN' undeclared (first use in this function)
   return tcp->th_flags == (TH_SYN | TH_ACK);
                            ^
[...]

PR #219, opened 2018-12-05
Fix build

tuxmaster5000:
Fix for https://github.com/greenbone/openvas-scanner/issues/218


tuxmaster5000:
_SVID_SOURCE will disable the BSD stlylee for /usr/include/netinet/tcp.h
But the BSD style is needed for nasl_builtin_synscan.c

janowagner:
IIRC, _SVID_SOURCE is meanwhile deprecated with newer glibc.

PR #226, opened 2018-12-07
Use a stored user agent instead of getting it from redis.

jjnicola:
The user agent does not change during a scan. It stores a global user agent (if there is one already set in redis kb). This is done just once after the ACT_SETTING VTs, location of the global_settings.nasl. 
If no user agent is set in the redis kb, it uses an user agent depending on
the vendor, or if there is no vendor set a default OpenVAS with the current
nasl lib version.


jjnicola:
Replaced with openvas-scanner#229 and openvas-scanner#231

PR #230, opened 2018-12-10
Replace kb_item_add_*() calls.

kroosec:
Use kb_item_add_*_unique() or kb_item_set_*() where needed.


jjnicola:
Depends on greenbone/gvm-libs#160

PR #231, opened 2018-12-12
Use vendor version in user-agent.

jjnicola:
If no user agent is set in preferences, it uses an user agent depending on
the vendor, or if there is no vendor set a default OpenVAS with the current
nasl lib version.

### DIFF
--- a/nasl/nasl_http.c
+++ b/nasl/nasl_http.c
@@ -25,6 +25,7 @@
 #include <gvm/util/kb.h>         /* for kb_item_get_str */
 
 #include "../misc/plugutils.h"  /* plug_get_host_fqdn */
+#include "../misc/vendorversion.h" /* for vendor_version_get */
 
 #include "nasl_tree.h"
 #include "nasl_global_ctxt.h"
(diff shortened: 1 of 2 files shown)


nasl/nasl_http.c
```diff
@@ -123,7 +124,13 @@ _http_req (lex_ctxt * lexic, char *keyword)
       if (!ua || strlen (g_strstrip (ua)) == 0)
         {
           g_free (ua);
-          ua = g_strdup ("Mozilla/5.0 [en] (X11, U; OpenVAS)");
+          vendor = g_strdup (vendor_version_get ());
```
kroosec: There is no need to g_strdup()/g_free() vendor.

nasl/nasl_http.c
```diff
@@ -123,7 +124,13 @@ _http_req (lex_ctxt * lexic, char *keyword)
...
+          if (!vendor || *vendor == '\0')
+            ua = g_strdup_printf ("Mozilla/5.0 [en] (X11, U; OpenVAS-VT %s)",
+                                         OPENVAS_NASL_VERSION);
```
kroosec: Code formatting while at it. OPENVAS_NASL_VERSION should start under "Mozilla...

PR #242, opened 2019-01-10
Update greenbone-nvt-sync.in

argusb:
The synchronization scripts do two rsyncs in quick succession, this leads to frequent IP blocking by the rsync server.

Fixed by adding a short sleep before the second (synchronization) rsync.


kroosec:
Thanks for the PR, but we prefer to fix this in the server, rather than adding sleeps to the script. Will discuss this issue with the feed team.

xBytez:
Any update on this @kroosec ?

Issue #211, opened 2018-11-19
Migrate remaining "DEBUG" calls / defines to new logging


cfi-gb:
With the final commit https://github.com/greenbone/openvas-scanner/commit/9240a44597ceee15bd33acbdb8efea0a50c8bee7#diff-6a755e7d228a2b491321dbae0a7a248b the scanner got a new logging mechanism similar to the other modules.

It seems there are a few DEBUG defines like ``NASL_DEBUG``, ``DEBUG``, ``DEBUG_SSL`` and ``TCPIP_DEBUGGING`` which needs to be defined on compile time. It could make sense to migrate them to the new logging functionality via openvassd_log.conf to make them configurable without requiring to compile the whole scanner to enable/disable them.

PR #245, opened 2019-01-17
Use g_debug() instead of DEBUG defines

kroosec:
Some of the debug messages wrapped in various DEBUG defines was replaced with adequate g_debug() calls. Others were removed (never used, are too noisy, provide no value etc,.)

Fixes https://github.com/greenbone/openvas-scanner/issues/211


kroosec:
@cfi-gb you can configure Glib logging with per-domain values (logging level (level=128 for debug), file path to use etc,.) in the openvassd_log.conf file (See the G_LOG_DOMAIN defines at the start of each file (eg. "lib  misc"))

Even with these logging domains separation and multiple removals in this PR, there are still too many useless messages to get much useful information.

Given that these were never used over the past 10 years (many included syntax errors, undefined variables etc,.), I doubt that we would start using them overnight. So better keep on cleaning the code, and continue adding our own debug messages where deemed to be useful as we currently do.

PR #246, opened 2019-01-17
Cleanup

kroosec:
be_nice doesn't provide any flexibility, and nicing the scanner if needed is better done from outside.

report_host_details is to be moved to host_details.nasl as a script preference.


cfi-gb:
Remark for the report_host_details, we might need to think about some sort of migrators for the move to the script preference of host_details.nasl or at least a warning in the log that it got moved:

If one has set report_host_details to "no" in the scanner configuration after the removal the host details will be sent again as the host_details.nasl preference will default to "yes".

PR #252, opened 2019-01-24
avoid unitialized access in nasl parsing

brada4:


brada4:
Sorry, cant fill file with 10 char extra comments.

PR #256, opened 2019-01-28
Switch advanced log to use host details.

kroosec:
Remove http and not-launched entries, as they dramatically increase the
amout of logging, without providing much value.

### DIFF
--- a/src/pluginlaunch.c
+++ b/src/pluginlaunch.c
@@ -122,10 +122,14 @@ update_running_processes (kb_t kb)
 
               if (prefs_get_bool ("advanced_log"))
                 {
-                  char buf[2048], buf2[2048];
-                  snprintf (buf, sizeof (buf), "log/launched/%s/end", oid);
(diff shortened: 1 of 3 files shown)


src/pluginlaunch.c
```diff
@@ -137,7 +141,17 @@ update_running_processes (kb_t kb)
                     g_message ("%s (pid %d) is slow to finish - killing it",
                                oid, processes[i].pid);
                   if (prefs_get_bool ("advanced_log"))
-                    kb_item_add_str (kb, "log/timedout", oid, 0);
+                    {
+                      char buffer[2048];
+
+                      snprintf (buffer, sizeof (buffer),
+                                "LOG||| |||general/Host_Details||| |||<host>"
+                                "<detail><name>nvt_end_time</name><value>>"
```
cfi-gb: ``<value>>`` -> ``<value>``

kroosec: @cfi-gb Thanks for spotting it. PR updated.

PR #257, opened 2019-01-28
Add function to set and get the current launched/running plugin.

jjnicola:
Some functions do not have access to the current running plugin,
what makes it hard to find the source of the error message during
a scan. With this PR it is possible to set and get the nasl
script filename to improve the error message.

Set the plugin name for each launched plugin.
This is useful to improve log messages, where the plugin name is unknown to be logged.

Improve error message in network.c ("Severe bug! Unhandled transport layer ...")
Add the filename of the plugin and the function name.

### DIFF
--- a/misc/network.c
+++ b/misc/network.c
@@ -49,6 +49,8 @@
 #include <gvm/util/serverutils.h> /* for load_gnutls_file */
 #include <gvm/util/kb.h>          /* for kb_item_get_str() */
 
+#include "../nasl/nasl_debug.h"          /* for nasl_*_filename */
+
 #ifdef __FreeBSD__
 #include <netinet/in.h>
 #define s6_addr32 __u6_addr.__u6_addr32
@@ -737,7 +739,8 @@ socket_negotiate_ssl (int fd, openvas_encaps_t transport,
   if (open_SSL_connection (fp, cert, key, passwd, cafile, hostname) <= 0)
     {
       g_free (hostname);
-      g_message ("socket_negotiate_ssl: SSL connection failed.");
+      g_message ("%s: socket_negotiate_ssl: SSL connection failed.",
+                 nasl_get_plugin_filename ());
       release_connection_fd (fd, 0);
       return -1;
     }
@@ -1220,8 +1223,10 @@ read_stream_connection_unbuffered (int fd, void *buf0, int min_len, int max_len)
 
     default:
       if (fp->transport || fp->fd != 0)
-        g_message ("Severe bug! Unhandled transport layer %d (fd=%d)",
-                   fp->transport, fd);
+        g_message ("Severe bug! Unhandled transport layer %d (fd=%d). "
+                   "Function %s called from %s.",
+                   fp->transport, fd, nasl_get_plugin_filename (),
+                   nasl_get_function_name ());
       else
         g_message ("read_stream_connection_unbuffered: "
                    "fd=%d is closed",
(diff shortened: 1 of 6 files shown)


nasl/nasl_debug.c
```diff
@@ -40,9 +41,32 @@ extern FILE *nasl_trace_fp;
 
 static char *debug_filename = NULL;
 static char *debug_funname = NULL;
+static char debug_plugin_filename[1024];
```
kroosec: Small pedantic change: Use PATH_MAX, instead of 1024.

src/attack.c
```diff
@@ -340,6 +341,10 @@ launch_plugin (struct scan_globals *globals, struct scheduler_plugin *plugin,
       return ERR_HOST_DEAD;
     }
 
+  name = nvticache_get_filename (oid);
+  nasl_set_plugin_filename (name);
```
kroosec: It would be better to call this from exec_nasl_script(), which already has the filename:
- No extra Redis command for every launched plugin.
- Would apply to openvas-nasl-* as well, if needed ?

jjnicola:
Thanks for the review @kroosec. Now works also for openvas-nasl*

misc/network.c
```diff
@@ -1220,8 +1223,10 @@ read_stream_connection_unbuffered (int fd, void *buf0, int min_len, int max_len)
 
     default:
       if (fp->transport || fp->fd != 0)
-        g_message ("Severe bug! Unhandled transport layer %d (fd=%d)",
-                   fp->transport, fd);
+        g_message ("Severe bug! Unhandled transport layer %d (fd=%d). "
+                   "Function %s called from %s.",
+                   fp->transport, fd, nasl_get_plugin_filename (),
+                   nasl_get_function_name ());
```
cfi-gb: nasl_get_plugin_filename () and nasl_get_function_name () seems to have the wrong order

misc/network.c
```diff
@@ -1399,8 +1404,10 @@ write_stream_connection4 (int fd, void *buf0, int n, int i_opt)
...
```
cfi-gb: nasl_get_plugin_filename () and nasl_get_function_name () seems to have the wrong order

nasl/nasl_ssh.c
```diff
@@ -328,8 +334,9 @@ nasl_ssh_connect (lex_ctxt *lexic)
 
       if (ssh_options_set (session, SSH_OPTIONS_PORT, &my_port))
         {
-          g_message ("Failed to set SSH port for '%s' to %d: %s",
-                     ip_str, port, ssh_get_error (session));
+          g_message ("%s: Failed to set SSH port for '%s' to %d: %s",
+                 nasl_get_plugin_filename (), ip_str, port,
```
cfi-gb: Indentation

nasl/nasl_ssh.c
```diff
@@ -724,8 +732,9 @@ nasl_ssh_set_login (lex_ctxt *lexic)
       if (username && *username &&
           ssh_options_set (session, SSH_OPTIONS_USER, username))
         {
-          g_message ("Failed to set SSH username '%s': %s",
-                     username, ssh_get_error (session));
+          g_message ("%s: Failed to set SSH username '%s': %s",
+                     username, ssh_get_error (session),
+                     nasl_get_plugin_filename ());
```
cfi-gb: nasl_get_plugin_filename () should be in front?

PR #275, opened 2019-02-12
Change NVT prefs format

kroosec:
Send and receive NVT preferences in OID:PrefType:PrefName instead of NVTName[PrefType]:PrefName.

This allows NVT plugins to change their names without breaking existing scan configs.

Relevant gvmd PR https://github.com/greenbone/gvmd/pull/394


kroosec:
@cfi-gb Added the check in script_add_preference() to see if a preference with the same name exists already.

PR #277, opened 2019-02-14
Update INSTALL documentation

bjoernricks:
An issue with the build process should be reported as a new ticket at github. Therefore don't point users to the community forum


cfi-gb:
As discussed via chat i'm still not sure if this could lead to more reports here at the issue tracker where just some dependencies / prerequisites are missing (e.g. libssh not installed).

bjoernricks:
I am also unsure about the best approach. But IMHO what's currently stated in the INSTALL.md is in contrast to the README.md. Therefore I would apply this PR. But it's up to @kroosec to decide.

PR #282, opened 2019-03-04
Add ID to NVT preferences.

kroosec:
Extend script_add_preference() accordingly. Adjust format in which NVT
preferences are sent to gvmd.


jjnicola:
Depends on greenbone/gvm-libs#192

PR #286, opened 2019-03-14
Fix OPENVASLIB_VERSION

jjnicola:
Set openvas_nasl and openvas_lib version to 11.0.0.
Use nasl_version as OPENVASLIB_VERSION.

### DIFF
--- a/misc/CMakeLists.txt
+++ b/misc/CMakeLists.txt
@@ -94,8 +94,8 @@ set (CMAKE_C_FLAGS              "${CMAKE_C_FLAGS} -Wall -Wextra -fPIC")
 ## Project version
 # The following three variables should be set through the project command once
 # we require CMake >= 3.0
-set (MISC_VERSION_MAJOR 9)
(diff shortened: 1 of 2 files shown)


wiegandm:
> Set openvas_nasl and openvas_lib version to 10.0.0.

Should not the value be `11.0.0` for the `master` branch (i.e., what will become GVM-11) and `10.0.0` for the `openvas-scanner-6.0` branch (i.e. GVM-10)?

bjoernricks:
Would it be possible to rename OPENVAS_VERSION to GVM_VERSION without much effort on all sides?

cfi-gb:
I had already a short discussion about this with @jjnicola and he mentioned some possible licensing issues which might prevent such a renaming.

But basically OPENVAS_VERSION is the version of the NASL libraries (previously openvas-libraries, now openvas-scanner) so calling it GVM_VERSION might be even more confusing as it is now (returns 10.0.0 for openvas-scanner 6.0.0).

PR #287, opened 2019-03-17
Bump Redis supported versions to 3.2 and higher.

kroosec:


janowagner:
Very welcome update. The older redis version are not tested by us anymore with old redis-versions,
so better not guess we support them and rather have 3.2 as minimum.

PR #302, opened 2019-04-12
Don't restrict send_packet() and send_v6packet() to target host.

kroosec:


nasl/nasl_packet_forgery.c
```diff
@@ -1437,27 +1433,6 @@ nasl_send_packet (lex_ctxt *lexic)
       bzero (&sockaddr, sizeof (struct sockaddr_in));
       sockaddr.sin_family = AF_INET;
       sockaddr.sin_addr = sip->ip_dst;
-      if (sockaddr.sin_addr.s_addr != inaddr.s_addr)
```
cfi-gb: I'm currently voting to not remove this restriction completely and only use broadcast / multicast addresses in addition to the current target IP:

While VTs having other means of accessing external systems not within the target scope by e.g. calling external tools like nmap, curl and similar we should have restrictions like this where possible to avoid mistakes.

cfi-gb:
Superseded/replaced by #388

PR #314, opened 2019-04-25
Use /var/tmp and not /tmp/

tuxmaster5000:
On modern Linux /tmp is only an tmpfs file system, so use /var/tmp to save memory.

### DIFF
--- a/tools/greenbone-nvt-sync.in
+++ b/tools/greenbone-nvt-sync.in
@@ -197,7 +197,7 @@ if [ -z "$TMPDIR" ]; then
   SYNC_TMP_DIR=/tmp
   # If we have mktemp, create a temporary dir (safer)
   if [ -n "`which mktemp`" ]; then
-    SYNC_TMP_DIR=`mktemp -t -d greenbone-nvt-sync.XXXXXXXXXX` || { echo "ERROR: Cannot create temporary directory for file download" >&2; exit 1 ; }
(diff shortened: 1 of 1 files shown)


janowagner:
This saves memory indeed, but tmpfs is faster and we can be sure data are deleted with a reboot.
I am undecided what weights more.

bjoernricks:
Feel free to submit patches that allow to configure the temp directory.

tuxmaster5000:
As far as I know, /var/tmp will cleaned by systemd regularly via systemd-tmpfiles-clean.timer/systemd-tmpfiles-clean.service

jjnicola:
Hi @tuxmaster5000,
I would like to know if you have intention to update this PR with a configuration option for the temp directory. Otherwise I would close this PR, since it will not be merged as it is. I appreciate your collaboration.
Regards

tuxmaster5000:
I have no idea how to fix the cmake file.

mattmundell:
I don't think this patch or a configuration option is a good idea.  It's already possible to achieve this by setting $TMPDIR.

From mktemp manpage:

       -t     interpret  TEMPLATE  as  a single file name component, relative to a directory: $TMPDIR, if set; else the directory specified via -p; else /tmp [depre‐
              cated]

mattmundell:
Closing due to lack of activity.

PR #318, opened 2019-04-30
Initialize the host kb from hosts_new().

jjnicola:
If there is no more availabe redis kb, it will wait until a host finishes
and release a db for the next one.
This avoid to hanging in the kb initialization and keeps working the results forwarding.


jjnicola:
Depends on PR #226

jjnicola:
Close in favor of PR #340

PR #328, opened 2019-05-15
Separate each scan plugin process into its own process group.

kroosec:
This makes cleaning-up children easier, and fixes issue of forgotten children of children.

Backport of parts from https://github.com/greenbone/openvas-scanner/pull/324


jjnicola:
With this PR, it takes longer to finish the scan.

PR #332, opened 2019-05-24
Each main scan process run with a niceness of 10.

jjnicola:
This commit complement the PR #313, which removed the be_nice scanner preference.
All scan processes and their children run with a niceness  of 10.


jjnicola:
The niceness will be set from outside. Closing here in favor of PR greenbone/ospd-openvas#78

PR #360, opened 2019-07-17
Preparing for release openvas-scanner-6.0.1

jjnicola:


CMakeLists.txt
```diff
@@ -66,8 +66,6 @@ set (PROJECT_BETA_RELEASE 0)
 
 if (GIT_REVISION)
   set (PROJECT_VERSION_GIT "${GIT_REVISION}")
-else (GIT_REVISION)
-  set (PROJECT_VERSION_GIT "~git")
```
bjoernricks: @jjnicola did you remove both lines by intention?

jjnicola: Yes, it is intentional. In Release versions (if there is no .git folder), the string "~git" must not be added to the lib object filenames.

bjoernricks: Hmm I suppose this works in  gsa without having to change the cmake file...

CHANGES
```diff
@@ -4,8 +4,37 @@ SUMMARY OF RELEASE CHANGES FOR OPENVAS-SCANNER
...
+* Do not launch the scan if the nvticache is corrupted or an error is detected
+  during the plugin schedule process.
+* Issues in building process have been addressed.
+* An issue which caused the manager to consider a scan as finished when it was
+  actually stopped has been addressed.
+* An issue which caused possible null IP values in OTP results has been
+  addressed.
+* An issue which caused forgotten children of children processes has been
+  addressed.
+* Advanced log feature has been removed.
```
cfi-gb: Maybe also mention "incomplete" or "unfinished" here as without this the line might be read as something useful was removed.

PR #370, opened 2019-08-12
Redis config

bjoernricks:
Add one single config for redis to `config/redis-openvas.conf`


doc/redis_config.txt
```diff
@@ -20,7 +20,7 @@ On the redis side, use the following directives:
 
   port 0 # prevent redis from listening on a TCP socket
   unixsocket /tmp/redis.sock
```
jjnicola: The path should be updated also in the doc file.

```suggestion
   unixsocket /run/redis-openvas/redis.sock
```

PR #377, opened 2019-08-29
Wait for zombie process in case of timed out nvts.

jjnicola:
Before, if an NVT timedout, the process is terminated and removed
from the running process struct, and the parent process does not wait for the child.
This leaved a zombie process until the end of the scan. The zombies proc become orphan proc
and removed for the system.
Now the timed out nvts are terminated, but a check is performerd to know
if the process was waited, before removing it from the running process struct.
In case it was not cleaned, it is kept in the running process table until is waited
for the parent and finally removed from the struct.


jjnicola:
Closing in favor of PR#379

PR #383, opened 2019-09-05
Remove unneeded includes and document remaining.

janowagner:


janowagner:
I do not think this deserves a CHANGELOG entry.

PR #303, opened 2019-04-12
Don't restrict send_packet() and send_v6packet() to target host.

kroosec:
Backport of https://github.com/greenbone/openvas-scanner/pull/302

PR #388, opened 2019-09-12
Allow send_packet() and send_v6packet() to send packets to broadcast/multicast

jjnicola:
For ipv4, the especial broadcast address 255.255.255.255 is allowed as dst to broadcast packet to the network you're connected to.
For ipv6, the multicast address fe02::1, being the closer option in ipv6 to the special broadcast address.
This PR replaces PRs #302 / #303


nasl/nasl_packet_forgery.c
```diff
@@ -1437,7 +1438,16 @@ nasl_send_packet (lex_ctxt *lexic)
       bzero (&sockaddr, sizeof (struct sockaddr_in));
       sockaddr.sin_family = AF_INET;
       sockaddr.sin_addr = sip->ip_dst;
-      if (sockaddr.sin_addr.s_addr != inaddr.s_addr)
+
+      if (allow_broadcast)
+        {
+          if (setsockopt (soc, SOL_SOCKET, SO_BROADCAST, &i, sizeof (i)) < 0)
```
mattmundell: I think it would be better to use a variable in this block instead of i, which was defined way up above and is not very descriptive.

nasl/nasl_packet_forgery.c
```diff
@@ -1437,7 +1438,16 @@ nasl_send_packet (lex_ctxt *lexic)
...
+            perror ("setsockopt ");
+          if (sockaddr.sin_addr.s_addr != INADDR_BROADCAST)
+            allow_broadcast = 0;
```
mattmundell: I don't know the context very well, so I am wondering why you would set the socket to broadcast above, if you're going to disable broadcasting here.

nasl/nasl_packet_forgery_v6.c
```diff
@@ -1688,7 +1689,26 @@ nasl_send_v6packet (lex_ctxt *lexic)
       bzero (&sockaddr, sizeof (struct sockaddr_in6));
       sockaddr.sin6_family = AF_INET6;
       sockaddr.sin6_addr = sip->ip6_dst;
-      if (dstip != NULL && !IN6_ARE_ADDR_EQUAL (&sockaddr.sin6_addr, dstip))
+
+      if (allow_multicast)
+        {
+          struct sockaddr_in6 multicast;
+
+          if (setsockopt (soc, SOL_SOCKET, SO_BROADCAST, &offset,
```
mattmundell: Similar to nasl_send_packet, this is not an offset, but an on/off optval, so the variable naming is slightly confusing.

nasl/nasl_packet_forgery.c
jjnicola: allow_broadcast will be disabled if the allow broadcast flag was enabled but the destination address differs from the broadcast address. 
In the next code line, it will check if the the dst ip differs from the target ip. In that case it will return immediately, except that the dst ip is the broadcast address and the allow_broadcast is set to 1.

mattmundell: OK thanks, I think I get it.

PR #412, opened 2019-10-22
Use nvti_get_tag() to avoid internal syntax

janowagner:
The pipe-separated syntax should not be exposed / expected.
So use the new api command nvti_get_tag() added with
https://github.com/greenbone/gvm-libs/pull/285

### DIFF
--- a/src/pluginscheduler.c
+++ b/src/pluginscheduler.c
@@ -70,7 +70,7 @@ plugin_add (plugins_scheduler_t sched, GHashTable *oids_table,
   /* Check if the plugin is deprecated */
   nvti = nvticache_get_nvt (oid);
   tag_value = nvti_get_tag (nvti, "deprecated");
-  if (tag_value && ! strcmp (tag_value, "1"))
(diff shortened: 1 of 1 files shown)


src/pluginscheduler.c
```diff
@@ -62,14 +62,15 @@ plugin_add (plugins_scheduler_t sched, GHashTable *oids_table,
...
+
   if (g_hash_table_lookup (oids_table, oid))
     return 0;
 
   /* Check if the plugin is deprecated */
   nvti = nvticache_get_nvt (oid);
-  if (nvti_tag (nvti)
-      && (g_str_has_prefix (nvti_tag (nvti), "deprecated=1")
-          || strstr (nvti_tag (nvti), "|deprecated=1")))
+  tag_value = nvti_get_tag (nvti, "deprecated");
```
jjnicola: Check for null, or apply the suggested change in https://github.com/greenbone/gvm-libs/pull/285#discussion_r338421370

janowagner: gvm-libs is fixed.

PR #416, opened 2019-10-29
[7.0] Improve handling of invalid or existent ids of nvt's preference id.

jjnicola:
It detects if the id is repeted.
Also detects if the id 0 is beeing used, which is invalid because id 0 is reserved
for the timeout preferences. Also detects if an id has a non-integer value.

### DIFF
--- a/nasl/nasl_scanner_glue.c
+++ b/nasl/nasl_scanner_glue.c
@@ -458,8 +458,15 @@ script_add_preference (lex_ctxt *lexic)
 
   if (!script_infos->nvti)
     return FAKE_CELL;
-  if (id <= 0)
+  if (id < 0)
     id = nvti_pref_len (script_infos->nvti) + 1;
+  if (id == 0)
+    {
+      nasl_perror (lexic,
+                   "Invalid id or not allowed id value in the call to %s()\n",
+                   __func__);
+      return FAKE_CELL;
+    }
   if (!name || !type || !value)
     {
       nasl_perror (lexic,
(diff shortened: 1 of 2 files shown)


nasl/nasl_scanner_glue.c
```diff
@@ -458,8 +458,15 @@ script_add_preference (lex_ctxt *lexic)
...
   if (!script_infos->nvti)
     return FAKE_CELL;
-  if (id <= 0)
+  if (id < 0)
     id = nvti_pref_len (script_infos->nvti) + 1;
+  if (id == 0)
+    {
+      nasl_perror (lexic,
+                   "Invalid id or not allowed id value in the call to %s()\n",
+                   __func__);
```
mattmundell: I think we should switch to `__func__` everywhere, instead of mixing it with the existing uses of `__FUNCTION__`.

nasl/nasl_scanner_glue.c
```diff
@@ -473,6 +480,11 @@ script_add_preference (lex_ctxt *lexic)
           nasl_perror (lexic, "Preference '%s' already exists\n", name);
           return FAKE_CELL;
         }
+      if (id == nvtpref_id (nvti_pref (script_infos->nvti, i)))
+        {
+          nasl_perror (lexic, "Invalid or already existent preferences id\n");
```
mattmundell:
```suggestion
          nasl_perror (lexic, "Invalid or already existent preference id\n");
```

PR #432, opened 2019-11-25
Increase size of buffer for preferences to allow for upto 105K NVTs

jjnicola:


CHANGES
```diff
@@ -18,6 +18,7 @@ Main changes compared to openvas-scanner 6.0.1:
 * An issue which could have caused a truncated string in register_service()
   has been fixed.
 * Improve signal handling when update vhosts list.
+* Increase size of buffer for preferences to allow for upto 105K NVTs.
```
cfi-gb:
```suggestion
* Increase size of buffer for preferences to allow for up to 105K NVTs.
```

src/ntp.c
```diff
@@ -179,7 +180,7 @@ ntp_read_prefs (int soc)
           g_debug ("Empty data string -- closing comm. channel");
           exit (0);
         }
-
+      buffer_size_count =  buffer_size_count + n;
```
cfi-gb:
```suggestion
      buffer_size_count = buffer_size_count + n;
```

PR #433, opened 2019-11-25
[5.1] Increase size of buffer for preferences to allow for upto 105K NVTs.

jjnicola:
Backport of https://github.com/greenbone/openvas/pull/432

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -8,6 +8,11 @@ $ git clone https://github.com/greenbone/openvas-scanner.git
 $ cd openvas-scanner && git checkout openvas-scanner-5.1 && git log
 
 
+openvas-scanner 5.1.3 (unreleased)
+
+* Increase size of buffer for preferences to allow for up to 105K NVTs.
+
+
 openvas-scanner 5.1.3 (2018-08-29)
 
 This is the third maintenance release of the openvas-scanner 5.1 module
--- a/src/ntp.c
+++ b/src/ntp.c
@@ -163,8 +163,9 @@ static int
 ntp_read_prefs (int soc)
 {
   char *input;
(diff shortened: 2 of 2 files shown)


src/ntp.c
```diff
@@ -209,6 +211,17 @@ ntp_read_prefs (int soc)
...
 
+  if (buffer_size_count >= input_sz) /* finished = 1; */
+    {
+      log_write ("%s: The received plugin list is longer than expected.\n",
+                 __func__);
+    }
+
+#if DEBUGMORE
+  log_write ("%s: Received %d chars from the client as plugin list.\n",
+￼            __func__, buffer_size_count);
```
bjoernricks: Could it be possible that the line contains some strange character at the beginning?

jjnicola: @bjoernricks,
I couldn't see the char, but removing all the spaces in that line and adding it again, there was a difference. So, pushing again with the invisible char.

src/ntp.c
```diff
@@ -179,6 +180,7 @@ ntp_read_prefs (int soc)
           exit (0);
         }
 
+      buffer_size_count =  buffer_size_count + n;
```
cfi-gb:
```suggestion
      buffer_size_count = buffer_size_count + n;
```

CHANGES
```diff
@@ -8,6 +8,11 @@ $ git clone https://github.com/greenbone/openvas-scanner.git
 $ cd openvas-scanner && git checkout openvas-scanner-5.1 && git log
 
 
+openvas-scanner 5.1.3 (unreleased)
+
+* Increase size of buffer for preferences to allow for upto 105K NVTs.
```
cfi-gb:
```suggestion
* Increase size of buffer for preferences to allow for up to 105K NVTs.
```

PR #436, opened 2019-11-28
Add gvm libs version to openvas --version output

bjoernricks:
Print gvm libs version in openvas --version output.

Requires https://github.com/greenbone/gvm-libs/pull/301


bjoernricks:
@wiegandm could you update the docker image build-env-openvas-scanner-master-debian-stretch-gcc-core to include libxml2-dev? otherwise the ci can't build this pr.

PR #439, opened 2019-12-02
Perform a scan even if there are missing plugins.

jjnicola:


bjoernricks:
Would it be possible to raise an error result in this case too? It would be good to display an error in the report for this issue. not only in the logs of the machine.

src/attack.c
```diff
@@ -1074,15 +1074,32 @@ attack_network (struct scan_globals *globals, kb_t *network_kb)
...
 
+  if (plugins_init_error > 0)
+    {
+      char buf[96];
+      int i = atoi (prefs_get ("ov_maindbid"));
+      kb_t main_kb = NULL;
+
+      sprintf (buf,
+               "%d errors were found during the plugin scheduling. "
+               "Some plugins will not be launch.",
```
bjoernricks: It suppose it should be "Some plugins will not be launched." or "Some plugins have not been launched."

PR #447, opened 2020-02-10
Don't detect MongoDB as a HTTP service.

cfi-gb:
Also updated the CHANGELOG.md because https://keepachangelog.com/en/1.0.0/ mentions that "Changed" should be used, not "Changes".


bjoernricks:
Looks good to me

PR #453, opened 2020-02-20
Send message to the client if the port list is not valid.

jjnicola:
Also set the status to finished.


jjnicola:
> Not sure if we should just add an error result, ignore the port and scan instead.

Yes. I thought about that, but this will need a change in gvm-lib/networking.c:validate_port_range()

PR #459, opened 2020-03-05
Fix init nvticache init.

jjnicola:
Add a new function to initialize the nvticache without loading the plugins.


jjnicola:
> Code wise it looks straight forward. Personally I would like to know the background of this PR :-)
With PR #458 I avoid to load plugins when a scan starts, since the plugins load up must be done calling openvas with --update-nvt-feed option. During this plugins load at the init of openvas, it was also done the nvticache init, necessary for the scan.

With this PR, I split the function and now I openvas can init the nvticache when a new start scan, but without loading up the plugins.

PR #489, opened 2020-04-29
Drop HTTP sync (7.0)

timopollmeier:
This removes wget and curl options for feed download.
These methods always download the entire feed instead of just the delta like rsync does.
The command line options --rsync, --curl and --wget are removed.

Backport of #478


bjoernricks:
Timo could you add https://github.com/greenbone/openvas/pull/488/commits/dea2ad463dc61ea07c79d13e83ccd456731d4563 ?

PR #491, opened 2020-04-29
Prevent deadlock, ensure that no locks are being held at the time of forking

jjnicola:
During testing, the child deadlock was reproduce, right after fork()'ing, when logging the message
"Testing xxx.xxx.xxx.xxx (Vhost...."

Before, since alive detection is running in a thread, it can lock a mutex when the main process is forking a new child. This mutex is copied into the child process in its locked state, and any child that tries to lock the mutex waits forever.

Also, add log domain for alive detection.


jjnicola:
Depends on PR greenbone/gvm-libs#328

PR #493, opened 2020-04-29
Drop network scan

janowagner:
Remove the network scan mode and all related code.

The experimental feature of the network scan mode of OpenVAS was developed to force scanning steps into parallel processing, especially for wrapped external tools like Nmap. The concept did not work out as expected but it helped to find advanced concepts. Wrapping tools in VTs did not deliver the desired seamless integration into the scan process. Greenbone rather started into micro-service architecture to bring bottlenecks into parallel seamless integration. Now it is time to finally drop the network scan mode.


hazcod:
@janowagner : can you elaborate why this is being removed?

bjoernricks:
@hazcod because it got never beyond of the experimental status and isn't supported at all.

PR #495, opened 2020-05-01
 Activate boreas alive scanner for explicit ICMP.

janowagner:
In case the target configuration explicitely selects
ICMP as alive detection, run the boreas alive detection
even if not configured to be always active.


janowagner:
A non-selective approach to activate Boreas is preferred.
Therfor closing this PR -  we do not need it anymore.

PR #496, opened 2020-05-04
Add aligned summary to log at scan end

janowagner:
Add a summary about duration and number of alive and total hosts to the log when the scan finishes. Have one for each, the alive scanner and the vulnerability scanner.
Also add the scan id to allow to distinguish when there are many concurrent scans.

### DIFF
--- a/src/alivedetection.c
+++ b/src/alivedetection.c
@@ -1326,7 +1326,7 @@ send_dead_hosts_to_ospd_openvas (void)
         {
           g_string_append (chunked_hosts, host_str);
           hosts_in_chunk++;
-          count_dead_ips ++;
+          count_dead_ips++;
 
           if (hosts_in_chunk == 1000)
             {
--- a/src/alivedetection.c
+++ b/src/alivedetection.c
@@ -1286,14 +1286,15 @@ send_arp (__attribute__ ((unused)) gpointer key, gpointer value,
  * All hosts which are not identified as alive are sent to ospd-openvas. This is
  * needed for the calculation of the progress bar for gsa in ospd-openvas.
  *
- * @return 0 on success, <0 on failure.
+ * @return number of dead IPs, or -1 in case of an error.
  */
-static void
+static int
 send_dead_hosts_to_ospd_openvas (void)
 {
   kb_t main_kb = NULL;
   int maindbid;
   int hosts_in_chunk = 0;
+  int count_dead_ips = 0;
 
   GHashTableIter target_hosts_iter;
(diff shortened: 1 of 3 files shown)


src/attack.c
```diff
@@ -1421,8 +1421,15 @@ attack_network (struct scan_globals *globals, kb_t *network_kb)
   plugins_scheduler_free (sched);
 
   gettimeofday (&now, NULL);
-  g_message ("Total time to scan all hosts : %ld seconds",
-             now.tv_sec - then.tv_sec);
+  if (alive_hosts_list)
```
ArnoStiefvater, line 1424: If we use the new alive detection and 0 hosts are alive in the target range we won't print the new info because `alive_hosts_list` is NULL. 
I would just use `if (test_alive_hosts_only)`.

src/attack.c
```diff
@@ -1421,8 +1421,15 @@ attack_network (struct scan_globals *globals, kb_t *network_kb)
...
+    g_message ("Vulnerability scan %s finished in %ld seconds: "
+               "%d alive hosts of %d",
+               globals->scan_id, now.tv_sec - then.tv_sec,
+               gvm_hosts_count (alive_hosts_list), gvm_hosts_count (hosts));
```
ArnoStiefvater, line 1428: If consider alive was chosen and e.g. 10 hosts are in the target range. This will print `... :  10 alive hosts of 10` even though they might all be dead. Maybe for the logs the message is good enough though.

src/alivedetection.c
```diff
@@ -1295,6 +1296,7 @@ send_dead_hosts_to_ospd_openvas (void)
         {
           g_string_append (chunked_hosts, host_str);
           hosts_in_chunk++;
+          count_dead_ips++;
```
ArnoStiefvater, line 1299: I think its good enough for now but keep in mind that we might have edge cases here when `max_alive_hosts` is used because of the way the progress bar and resume scan works. Alive hosts might be counted as dead hosts because they were not allowed to be considered as alive.

src/attack.c
janowagner: I adopted your proposal.

src/attack.c
janowagner: This is surely debatable. The "consider alive" option is special anyway. I suggest
to leave as is.

src/alivedetection.c
janowagner: OK. But nothing to address in this particular PR I guess.

PR #503, opened 2020-05-12
Prepare relase 6.0.2

jjnicola:
Version bump for OpenVAS and libs
Update Changelog.


CHANGES
```diff
@@ -21,6 +21,7 @@ Main changes compared to openvas-scanner 6.0.1:
 * Increase size of buffer for preferences to allow for up to 105K NVTs.
 * Perform the sca even if there are missing plugins in the nvticache.
```
cfi-gb, line 22:
```suggestion
* Perform the scan even if there are missing plugins in the nvticache.
```

PR #505, opened 2020-05-13
Add some tests

ArnoStiefvater:
Some tests for `routethrough()`.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -175,6 +175,8 @@ target_link_libraries (openvas openvas_nasl_shared ${LIBOPENVAS_NASL_LDFLAGS}
 
 ## Tests
 
+set (ALIVEDETECTION_TEST_LINKER_WRAP_OPTIONS
+     "-Wl,-wrap,socket,-wrap,setsockopt")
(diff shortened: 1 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -184,7 +184,8 @@ target_link_libraries (alivedetection-test openvas_nasl_shared cgreen
                        ${LIBGVM_BASE_LDFLAGS}
                        ${GLIB_LDFLAGS}
                        ${PCAP_LDFLAGS}
-                       ${LINKER_HARDENING_FLAGS} ${CMAKE_THREAD_LIBS_INIT})
+                       ${LINKER_HARDENING_FLAGS} ${CMAKE_THREAD_LIBS_INIT} 
+                       -Wl,-wrap,socket,-wrap,setsockopt)
```
jjnicola, line 188: Instead of using the flags directly here, it would be nice to have something like
```
set (LINKER_WRAP_OPTIONS     "-Wl,-wrap,socket,-wrap,setsockopt")
```
and then 
```suggestion
                       ${LINKER_WRAP_OPTIONS})
```

PR #507, opened 2020-05-14
Use flock to lock the feed lock file.

jjnicola:


jjnicola:
Thanks  @mattmundell !

PR #510, opened 2020-05-20
Fix host_deny

jjnicola:
Finished the denied hosts and let openvas to finish a scan. Otherwise they never ended and openvas keep looking for the host to finished, which has been never started.

### DIFF
--- a/src/attack.c
+++ b/src/attack.c
@@ -170,11 +170,12 @@ comm_send_status (kb_t kb, char *hostname, int curr, int max)
 }
 
 static void
-error_message_to_client2 (kb_t kb, const char *msg, const char *port)
+error_message_to_client2 (kb_t kb, const char *msg, const char *ip_str, const char *port)
 {
   char buf[2048];
 
-  sprintf (buf, "ERRMSG||| |||%s||| |||%s", port ?: " ", msg ?: "No error.");
+  sprintf (buf, "ERRMSG|||%s|||%s||| |||%s", ip_str ?: "",
+           port ?: " ", msg ?: "No error.");
   kb_item_push_str (kb, "internal/results", buf);
 }
 
@@ -994,7 +995,7 @@ attack_network (struct scan_globals *globals)
       connect_main_kb (&main_kb);
       error_message_to_client2 (
         main_kb, "Invalid port list. Ports must be in the range [1-65535]",
-        NULL);
+        NULL, NULL);
       kb_lnk_reset (main_kb);
       g_warning ("Invalid port list. Ports must be in the range [1-65535]. "
                  "Scan terminated.");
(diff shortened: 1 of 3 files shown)


src/attack.c
```diff
@@ -591,15 +593,24 @@ check_host_authorization (gvm_host_t *host, const struct in6_addr *addr,
   hosts_deny = gvm_hosts_new (prefs_get ("hosts_deny"));
   if (!host_authorized (host, addr, hosts_allow, hosts_deny))
     {
-      error_message_to_client2 (kb, "Host access denied.", NULL);
+      char ip_str[INET6_ADDRSTRLEN];
+
+      addr6_to_str (addr, ip_str);
+      error_message_to_client2 (kb, "Host access denied.", ip_str, NULL);
```
ArnoStiefvater, line 599: We could let the caller handle the error reporting. We already print a warning in `attack_start` if access denied. We could send the message there so that `check_host_authorization` is only responsible for checking and nothing more. This way we do not need to do `addr6_to_str` two times.

src/attack.c
```diff
@@ -591,15 +593,24 @@ check_host_authorization (gvm_host_t *host, const struct in6_addr *addr,
...
   sys_hosts_deny = gvm_hosts_new (prefs_get ("sys_hosts_deny"));
   if (!host_authorized (host, addr, sys_hosts_allow, sys_hosts_deny))
     {
+      char ip_str[INET6_ADDRSTRLEN];
+
+      addr6_to_str (addr, ip_str);
       error_message_to_client2 (
-        kb, "Host access denied (system-wide restriction.)", NULL);
+        kb, "Host access denied (system-wide restriction.)", ip_str, NULL);
+      kb_item_set_str (kb, "internal/host_deny", "True", 0);
```
ArnoStiefvater, line 612: This is done for every host (in `struct host *hosts`). I think it would be better to include this in the host directly and not via Redis. Maybe something like `host->host_deny` or better `host->access_allowed` (so its harder to missunderstand).

E.g. in `hosts_new()` do `h->access_allowed = FALSE;` and then in `attack_start()` do something like
```
  if (check_host_authorization (args->host, &hostip, kb))
    {
      error_message_to_client2...
      return;
    }
  else 
      hosts_set_access_allowed(name);
```

src/hosts.c
```diff
@@ -240,14 +240,20 @@ hosts_read_data (void)
...
           h->ip = kb_item_get_str (h->host_kb, "internal/ip");
           if (h->ip)
             host_set_time (h->host_kb, "internal/start_time");
+          else
+            /* internal/host_deny is set during check_host_authorization() */
+            host_deny = kb_item_get_str (h->host_kb, "internal/host_deny");
         }
-      if (h->ip)
+
+      if (h->ip || host_deny)
```
ArnoStiefvater, line 256: Maybe `if (h->ip ||  !(host->access_allowed))`

See previous comment for context.

src/attack.c
jjnicola: This is not possible, because the check is done from the host process and the modified struct will not be change the data in the accessed by host_read(), which is called for the parent process. Therefore, the communication is done through redis.

PR #513, opened 2020-05-29
Simplify TCP ping

ArnoStiefvater:
* Do not calculate checksum for IP header manually.
* Do not use `routethrough()` for getting the source address. Open udp socket, `connect()` and use `getsockname()` for acquiring the src addr instead.
* Add tests


src/alivedetection.c
```diff
@@ -1008,36 +1008,27 @@ send_tcp_v6 (int soc, struct in6_addr *dst_p, uint8_t tcp_flag)
...
-
-      g_debug ("%s: Use %s as source IP for IPv4 pings.", __func__, addr_str);
-    }
-
   /* No ports in portlist. */
   if (scanner.ports->len == 0)
     return;
 
+  /* Get source address for TCP header. */
+  if ((error = get_source_addr_v4 (&scanner.udpv4soc, dst_p, &src)) != 0)
```
mattmundell, line 1035: Just a minor note, there's no need to embed the error assignment in the `if` here.  It would be easier to read if the assignment was done before the `if`, and it would be less prone to = vs == confusion.

```suggestion
  error = get_source_addr_v4 (&scanner.udpv4soc, dst_p, &src);
  if (error)
```

PR #517, opened 2020-06-03
Remove dependency

ArnoStiefvater:
* Use alternative function for getting src addr and delete now unused include


src/alivedetection.c
```diff
@@ -19,11 +19,9 @@
 
 #include "alivedetection.h"
 
-#include "../misc/pcap.c" /* routethrough functions */
```
jjnicola, line 22: :)

PR #530, opened 2020-06-15
Add nasl function to get the host kb index.

jjnicola:
The functions returns a integer, the redis db index of the host kb
related to the host running the plugin.

### DIFF
--- a/nasl/nasl_scanner_glue.c
+++ b/nasl/nasl_scanner_glue.c
@@ -763,6 +763,33 @@ get_kb_item (lex_ctxt *lexic)
   return retc;
 }
 
+/**
+ * @brief Get the kb index of the host running the current script.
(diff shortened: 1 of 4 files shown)


nasl/nasl_scanner_glue.c
```diff
@@ -763,6 +763,33 @@ get_kb_item (lex_ctxt *lexic)
   return retc;
 }
 
+/**
+ * @brief Get the kb index of the host running the current script.
+ *
+ * @param[in] lexic     NASL lexer.
+ *
+ * @return lex cell containing thee host kb index value as positive integer.
```
ArnoStiefvater, line 771:
```suggestion
 * @return lex cell containing the host kb index value as positive integer.
```

PR #534, opened 2020-06-19
Sleep between checks for last plugin to finish before changing to another category

jjnicola:


src/attack.c
```diff
@@ -506,6 +506,8 @@ attack_host (struct scan_globals *globals, struct in6_addr *ip, GSList *vhosts,
         }
       else if (plugin == NULL)
         break;
+      else if (plugin != NULL && plugin == PLUG_RUNNING)
+        usleep (50000);
```
bjoernricks, line 510: Could you add a comment about the time length? Maybe in miliseconds or seconds. it is not obvious how long this is

PR #546, opened 2020-07-10
Fix calculation of cell size in get_tcp_element()

ArnoStiefvater:


nasl/nasl_packet_forgery.c
```diff
@@ -570,9 +570,9 @@ get_tcp_element (lex_ctxt *lexic)
   else if (!strcmp (element, "data"))
     {
       retc = alloc_typed_cell (CONST_DATA);
-      retc->size = UNFIX (ip->ip_len) - ntohl (tcp->th_off) * 4;
+      retc->size = UNFIX (ip->ip_len) - tcp->th_off * 4;
       retc->x.str_val = g_malloc0 (retc->size);
-      bcopy (tcp + ntohl (tcp->th_off) * 4, retc->x.str_val, retc->size);
+      bcopy (tcp + tcp->th_off * 4, retc->x.str_val, retc->size);
```
cfi-gb, line 575: In nasl/nasl_packet_forgery.c the following is used:

```
bcopy ((char *) tcp + tcp->th_off * 4, retc->x.str_val, retc->size);
```

I can't tell about the difference between with / without `(char *)` but might worth a look.

Besides the above i start doing some testing now.

ArnoStiefvater: You are right. Nice catch. It has to be `(char *)tcp`. Else it would be incremented in steps of 20 bytes (tcphdr size).

cfi-gb: Could this also be responsible that "crap" data is included in the response? I'm asking because on a `get_tcp_element` to `data` i'm getting something like e.g. the following back on a TCP packet without any data field in the TCP packet:

```
lib  nasl-Message: 10:22:10.396: 0x00:  00 00 00 00 71 00 00 00 00 00 00 00 20 56 F8 7B    ....q....... V.{
0x10:  CD 55 00 00                                        .U.. 
```

cfi-gb: Confirmed that 3e1a3db295f0b649bdef22868fdb148513e9f2af solves the data extraction.

nasl/nasl_packet_forgery.c
```diff
@@ -570,9 +570,9 @@ get_tcp_element (lex_ctxt *lexic)
   else if (!strcmp (element, "data"))
     {
       retc = alloc_typed_cell (CONST_DATA);
-      retc->size = UNFIX (ip->ip_len) - ntohl (tcp->th_off) * 4;
+      retc->size = UNFIX (ip->ip_len) - (tcp->th_off + ip->ip_hl) * 4;
```
cfi-gb, line 573: I guess a similar fix for this done in 3e1a3db295f0b649bdef22868fdb148513e9f2af is required for `get_tcp_v6_element` here as well?

https://github.com/greenbone/openvas/blob/d589114edb38c082e5ee25b9d364f24a3a8cc34d/nasl/nasl_packet_forgery_v6.c#L592

At least when getting the data element with `get_tcp_v6_element()` (see internal tracker for a reproducer) i'm getting the following message:

```
lib  nasl-Message: 13:51:28.877: [70564](tcp_data.nasl:59) get_tcp_v6_element: Erroneous tcp header offset 0
```

ArnoStiefvater: I think the size calculation is correct in this case. But data of size 0 was considered as an error. So I think it was just a wrong error message. Should be fixed now.

cfi-gb: Indeed this should do the trick. The "header offset 0" was a little bit misleading in this case.

PR #548, opened 2020-07-10
Backport fix of icmp checksum calculation

ArnoStiefvater:
This also fixes an issue of not being able to ICMP ping IPv6 addresses.


CHANGES
```diff
@@ -22,6 +22,7 @@ Main changes compared to openvas-scanner 6.0.1:
 * Perform the scan even if there are missing plugins in the nvticache.
 * Drop HTTP sync.
 * Use new URL for GCF rsync.
+* Fix icmp checksum calculation in openvas-nasl.
```
cfi-gb, line 25: This should have been added to a new `openvas-scanner 6.0.3 (unreleased)` item instead of to the already released 6.0.2. Fix is available in #554

PR #549, opened 2020-07-10
Fix wrong proto and byte order in dump_ipv6_packet

ArnoStiefvater:


nasl/nasl_packet_forgery_v6.c
```diff
@@ -330,22 +330,22 @@ dump_ipv6_packet (lex_ctxt *lexic)
...
-          printf ("\tip6_tc: %d\n", (ip6->ip6_flow >> 20) & 0xff);
-          printf ("\tip6_fl: %d\n", (ip6->ip6_flow) & 0x3ffff);
+          printf ("\tip6_v  : %d\n", ntohl (ip6->ip6_flow) >> 28);
+          printf ("\tip6_tc: %d\n", (ntohl (ip6->ip6_flow) >> 20) & 0xff);
+          printf ("\tip6_fl: %d\n", ntohl (ip6->ip6_flow) & 0x3ffff);
           printf ("\tip6_plen: %d\n", UNFIX (ip6->ip6_plen));
-          printf ("\tip6_nxt : %d\n", ntohs (ip6->ip6_nxt));
-          printf ("\tip6_hlim : %d\n", ntohs (ip6->ip6_hlim));
-          switch (ip6->ip6_ctlun.ip6_un1.ip6_un1_nxt)
+          printf ("\tip6_nxt : %d\n", ip6->ip6_nxt);
```
cfi-gb, line 337: I think this could be dropped because it is already included in the `switch` below.

PR #552, opened 2020-07-13
[20.08] Fix IPPROTO_IP string in nasl_init.c

cfi-gb:
No NASL script is currently calling this variable so we just can fix it right away.


cfi-gb:
Backports in #553 and #554

PR #556, opened 2020-07-14
Use correct key for retrieving auth for HTTP

ArnoStiefvater:


cfi-gb:
This is a fix for a regression introduced in #502 (dcfb9b287e074ee506364ccb6d000108e4ae2e01). Basically the string like `11` saved in the KB Key `http/$port` ended up in the HTTP GET request crafted by `http_get()`.

PR #387, opened 2019-09-09
Fix mis-identification of Sphinx Search service

jjnicola:
It was detected as MySQL in nasl_builtin_find_service.c

PR #559, opened 2020-07-28
Add a missing space to a log message.

cfi-gb:
Minor improvement for #387 as the log message was showing up like the following without a space:

```
A Sphinx search server (MySQL listener)seems to be running on this port
```


cfi-gb:
Backport to the openvas-7.0 branch in #564

PR #565, opened 2020-08-03
[20.08] Drop wget and curl mentions from the install file.

cfi-gb:
#478 had dropped the HTTP sync via wget and curl so it can be dropped here as well.


cfi-gb:
Backports to the other active branches done in #566 and #567

PR #569, opened 2020-08-07
Fix syntax of openvas manpage.

cfi-gb:
Previous to this change the man page related part looked like this:

```
              https://community.greenbone.net ⟨⟩ (Community site)
              https://github.com/greenbone/ ⟨⟩ (Development site)
              https://www.openvas.org/ ⟨⟩ (Traditional home site)
```

Now it is using the correct syntax as explained in https://man7.org/linux/man-pages/man7/man.7.html and is showing up like:

```
              Community site ⟨https://community.greenbone.net⟩
              Development site ⟨https://github.com/greenbone⟩
              Traditional home site ⟨https://www.openvas.org⟩
```

The final manpage can be verified with something like e.g.:

```
man doc/openvas.8.in
```

### DIFF
--- a/doc/openvas.8.in
+++ b/doc/openvas.8.in
@@ -221,20 +221,17 @@ The canonical places where you will find more information
 about OpenVAS are: 
 
 .RS
-.UR
-https://community.greenbone.net
+.UR https://community.greenbone.net
+Community site
 .UE
-(Community site)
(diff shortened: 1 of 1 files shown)


cfi-gb:
Backports to the other active branches in #570 and #571

PR #576, opened 2020-08-13
Fork vhosts before creating the socket.

jjnicola:
This solves the issue in which a parent process creates a socket
and child processes inheritates the socket but are not allow to use it.
Now, each child uses its own socket.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [20.08.1] (unreleased)
+
+### Fixed
+- Fork vhosts before creating the socket.[#576](https://github.com/greenbone/openvas/pull/576)
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [20.08.1] (2020-08-11)
```
cfi-gb, line 7:
```suggestion
## [20.08.1] (unreleased)
```

jjnicola: ouch!

jjnicola:
> I can confirm that this fixes scanning SSL/TLS service for me.
> 
> Unfortunately i cannot judge on the C-Code changes (e.g. if `hostname` needs to be freed as well), this needs to be reviewed and approved by a @greenbone/gvm-dev member.

Thanks a lot for testing @cfi-gb

cfi-gb:
Backport to the openvas-7.0 branch which is also affected in #579

PR #581, opened 2020-08-19
Check if another forked child has already added the vhost.

jjnicola:
In some cases, in which expand_vhosts and test_empty_vhost options
are enabled, a plugin process fork()s one child for vhost.
These children inherit the current state of vhosts list, but it does not
check if the concurrent child plugin has already added a new found vhost.
Now, it checks not only if the vhost has been added by another plugin before
(and already exists in the vhosts list), but also checks if a concurrent child
added to the internal/vhosts key in redis.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 - Fork vhosts before creating the socket.[#576](https://github.com/greenbone/openvas/pull/576)
+- Check if another forked child has already added the same vhost. [#581](https://github.com/greenbone/openvas/pull/581)
 
 [20.08]: https://github.com/greenbone/openvas/compare/v20.8.0...openvas-20.08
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 - Fork vhosts before creating the socket.[#576](https://github.com/greenbone/openvas/pull/576)
+- Check if another forked child has already added the a vhost. [#581](https://github.com/greenbone/openvas/pull/581)
```
cfi-gb, line 11:
```suggestion
- Check if another forked child has already added a vhost. [#581](https://github.com/greenbone/openvas/pull/581)
```

cfi-gb (commented):
Tested for regressions of this change with some larger scans. No additional problems found, scans are still working as expected.

I wasn't able to reproduce the initial problem though and can't judge on the code changes itself, a final approval is required by @greenbone/gvm-dev

ArnoStiefvater (approved):

cfi-gb:
Backport to the openvas-7.0 branch in #584

PR #591, opened 2020-08-28
Handle the stop scan out of the interruption rutine

jjnicola:
During interruption, just set a variable. This variable is checked later
and the main process does the cleanup if the scan stop signal was received.
Signal handlers must be lean. Malloc is a non-async-signal
funtion which can generates deadlocks if it is called from inside a signal handler.
Redis related functions (kb_item_get_str()) use malloc.
Therefore this avoid deadlock.

**What**:

During interruption, just set a variable. This variable is checked later
and the main process does the cleanup if the scan stop signal was received.

**Why**:

Signal handlers must be lean. Malloc is a non-async-signal
funtion which can generates deadlocks if it is called from inside a signal handler.
Redis related functions (kb_item_get_str()) use malloc.
Therefore this avoid deadlock.

**How**:
Not always is reproducible a deadlock. But you can try starting and stopping a task.

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/attack.c
+++ b/src/attack.c
@@ -208,10 +208,13 @@ fork_sleep (int n)
 }
 
 int global_scan_stop = 0;
+static void scan_stop_cleanup (void);
 
 static int
 scan_is_stopped (void)
 {
+  if (global_scan_stop == 1)
+    scan_stop_cleanup ();
   return global_scan_stop;
 }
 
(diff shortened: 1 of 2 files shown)


jjnicola:
Duplicated for master. It was intended to be created for 20.08.

PR #593, opened 2020-08-28
Only send the signal if the pid is a positive value.

jjnicola:
**What**:
Only send the signal if the pid is a positive value.

**Why**:
Since kb_item_get_int() will return -1 if the key does
not exist. killing with -1 as pid will send the signal system wide.
Apparently the stop signal is received twice, being the second one
sent by a child process, but received after the kb was already cleaned up.
Therefore -1 is received and the kill SIGUSR1 signal reaches other process.

**How**:
You can test this starting 2 or 3 parallel scans with many hosts. This produces a big amount child process which during stops scans, spread the kill signal to other process. You should be able to stop the scans without problem if this patch is applied.

**Checklist**:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fork vhosts before creating the socket.[#576](https://github.com/greenbone/openvas/pull/576)
 - Check if another forked child has already added the same vhost. [#581](https://github.com/greenbone/openvas/pull/581)
 - Send duplicated hosts as dead hosts to ospd, to adjust scan progress calculation. [#586](https://github.com/greenbone/openvas/pull/586)
+- Only send the signal if the pid is a positive value. [#593](https://github.com/greenbone/openvas/pull/593)
 
 [20.08]: https://github.com/greenbone/openvas/compare/v20.8.0...openvas-20.08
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fork vhosts before creating the socket.[#576](https://github.com/greenbone/openvas/pull/576)
 - Check if another forked child has already added the same vhost. [#581](https://github.com/greenbone/openvas/pull/581)
 - Send duplicated hosts as dead hosts to ospd, to adjust scan progress calculation. [#586](https://github.com/greenbone/openvas/pull/586)
+- Only send the signal if the pid is a positive value.[#593](https://github.com/greenbone/openvas/pull/593)
```
cfi-gb, line 13:
```suggestion
- Only send the signal if the pid is a positive value. [#593](https://github.com/greenbone/openvas/pull/593)
```

src/openvas.c
```diff
@@ -408,6 +408,13 @@ stop_single_task_scan (void)
     exit (1);
 
   pid = kb_item_get_int (kb, "internal/ovas_pid");
+
+  /* Only send the signal if the pid is a positive value.
+     since kb_item_get_int() will return -1 if the key does
```
cfi-gb, line 413:
```suggestion
     Since kb_item_get_int() will return -1 if the key does
```

cfi-gb (changes requested):

ArnoStiefvater (approved):

ArnoStiefvater (approved):

cfi-gb (approved):

PR #599, opened 2020-09-14
￼Consider csv files for checksum check and upload in redis cache.

jjnicola:
**What**:
￼
Consider csv files during the plugins load up and perform the checksum check if the signature check is enabled.

**Why**:
First step for preparing the scanner to handle a csv with metadata of many vulnerability tests.

**How**:
Add .csv in the plugins folder. Then, check in redis if a new entry filename:file.csv has been stored.
For checking with the signature check, you have to add the sha256 checksum of the file in the sha256sums file and generate the signature file again with a valid private key, having the public one in the gpg keyring.

**Checklist**:

- [ ] Tests N/A
- [x ] [CHANGELOG](https://github.com/greenbone/openvas/blob/master/CHANGELOG.md) Entry

### DIFF
--- a/nasl/nasl_global_ctxt.h
+++ b/nasl/nasl_global_ctxt.h
@@ -36,6 +36,9 @@ typedef struct
 int
 init_nasl_ctx (naslctxt *, const char *);
 
+int
+init_notus_vt_list (const kb_t, const char *, int);
+
 void
 nasl_clean_ctx (naslctxt *);
 
--- a/nasl/nasl_global_ctxt.h
+++ b/nasl/nasl_global_ctxt.h
@@ -19,6 +19,7 @@
 #ifndef _NASL_CTX_H
 #define _NASL_CTX_H
 
+#include "../nasl/nasl_lex_ctxt.h" /* for tree_cell */
 /* for FILE */
 #include <gvm/util/kb.h>
 #include <stdio.h>
@@ -37,7 +38,7 @@ int
 init_nasl_ctx (naslctxt *, const char *);
 
 int
-init_notus_vt_list (const kb_t, const char *, int);
+csv_vt_list_checksum_check (const kb_t, const char *, int);
 
 void
 nasl_clean_ctx (naslctxt *);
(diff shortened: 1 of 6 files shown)


nasl/nasl_global_ctxt.h
```diff
@@ -36,6 +37,9 @@ typedef struct
 int
 init_nasl_ctx (naslctxt *, const char *);
 
+int
+init_notus_vt_list (const kb_t, const char *, int);
```
ArnoStiefvater, line 41: I would not use the name notus. Why not something like init_csv_vt_list ?
Or even just init_csv_file or init_csv_file_check.

nasl/nasl_grammar.y
```diff
@@ -694,6 +695,108 @@ file_checksum (const char *filename, int algorithm)
   return result;
 }
 
+/**
+ * @brief .
```
ArnoStiefvater, line 699: missing description

nasl/nasl_grammar.y
```diff
@@ -694,6 +695,108 @@ file_checksum (const char *filename, int algorithm)
...
+  if (!inc_dirs)
+    add_nasl_inc_dir ("");
+
+  while (inc_dir != NULL)
+    {
+      if (full_name)
+        g_free (full_name);
+      full_name = g_build_filename (inc_dir->data, name, NULL);
+
+      if ((g_file_get_contents (full_name, &buffer, &flen, NULL)))
```
ArnoStiefvater, line 727: `buffer` is never freed. Probably needs to be freed same as `full_name`

src/pluginload.h
```diff
@@ -53,6 +53,9 @@ total_loading_plugins (void);
 int
 nasl_plugin_add (char *, char *);
 
+int
+check_notus_vt_list (char *, char *);
```
ArnoStiefvater, line 57: Same as above. I would not use the name notus. Why not something like check_csv_vt_list ?
Or even just check_csv_file.

ArnoStiefvater (approved):

jjnicola:
@ArnoStiefvater Thanks for the review!

PR #601, opened 2020-09-24
Fix passphrase for snmp v3

jjnicola:
**What**:
Pass the passphrase correctly

**Why**:

**How**:
Attack an snmpv3 target with the right credential (SHA, AES).

**Checklist**:

- [ ] Tests


jjnicola:
Closed here since it does not fix the issue.

PR #606, opened 2020-10-08
Send message to client with the amount of host to be scanned.

jjnicola:
**What**:
Send message to client with the amount of host to be scanned.

**Why**:
The host count can differ from the original one. For a correct progress calculation, a message is sent to the client.

**How**:

**Checklist**:

- [ ] Tests


jjnicola:
Depends on greenbone/ospd-openvas#343 and greenbone/ospd#332

ArnoStiefvater (approved):

PR #610, opened 2020-10-21
Use metric when finding best route

ArnoStiefvater:
**What**:

Currently for a destination address the last entry in the routes table which matches will be taken even if the last entry has the same mask and worse metric.

Now metric is also taken into account. 

**Why**:

Wrong interface might be chosen otherwise.

**How**:

**Checklist**:

- [ ] Tests


cfi-gb (commented):
Did a short test and this fixes my issue without introducing any regressions.

Final code review and approval needs to be done by a @greenbone/gvm-dev member.

jjnicola (approved):

PR #613, opened 2020-10-23
Consistent naming of functions and parameters and byte order fix

ArnoStiefvater:
**What**:

* Make naming of functions consistent. E.g. `forge_ipv6_packet` -> `forge_ip_v6_packet`
* Rename func parameter from `ipv6` to `ip6`
* Fix byte order when retrieving ip6 elements in `get_ipv6_element`

**Why**:

* Easier use of nasl functions.
* Correct value of ip6 elements.

**How**:

Check backwards compatibility
```
# create some ip6 packet and store it in ip6_packet
# All of the following calls should work and output the same
version = get_ipv6_element(ipv6:ip6_packet, element: "ip6_v");
display ("ipv6 " + version);
version = get_ipv6_element(ip6:ip6_packet, element: "ip6_v");
display ("ip6 " + version);
version = get_ip_v6_element(ipv6:ip6_packet, element: "ip6_v");
display ("ipv6 " + version);
version = get_ip_v6_element(ip6:ip6_packet, element: "ip6_v");
display ("ip6 " + version);
```

Check output of retrieved elements

```
# Create ip6 packet with some  (maybe nonsensical) data
IP6_v = 6;
IP6_P = 42;
IP6_HLIM = 42;
IP6_PLEN = 42;
IP6_TC = 42;
IP6_FL = 42;
ip6_packet = forge_ipv6_packet( ip6_v:IP6_v,
                                ip6_p:IP6_P,
                                ip6_plen:IP6_PLEN,
                                ip6_tc:IP6_TC,
                                ip6_fl:IP6_FL,
                                ip6_hlim:IP6_HLIM,
                                ip6_src:myhost,
                                ip6_dst:get_host_ip());

# Get and display the elements
version = get_ip_v6_element(ipv6:ip6_packet, element: "ip6_tc");
display ("ip6_tc " + version);
version = get_ip_v6_element(ipv6:ip6_packet, element: "ip6_fl");
display ("ip6_fl " + version);
version = get_ip_v6_element(ipv6:ip6_packet, element: "ip6_plen");
display ("ip6_plen " + version); # payload len is inserted automatically and is probably 0 even though we specified it above.
```

**Checklist**:

- [ ] Tests


nasl/nasl_init.c
```diff
@@ -193,16 +193,24 @@ static init_func libfuncs[] = {
...
   {"get_ipv6_element", get_ipv6_element},
-  {"set_ip_elements", set_ip_elements},
   {"set_ipv6_elements", set_ipv6_elements},
-  {"insert_ip_options", insert_ip_options},
   {"insert_ipv6_options", insert_ipv6_options},
-  {"dump_ip_packet", dump_ip_packet},
   {"dump_ipv6_packet", dump_ipv6_packet},
+
+  {"forge_ip_packet", forge_ip_packet},
+  {"forge_ip_v6_packet", forge_ipv6_packet},
```
cfi-gb, line 205: Any strong reasons against renaming the internal functions as well because we currently still have `forge_tcp_v6_packet` vs. `forge_ipv6_packet` for the internal functions.

ArnoStiefvater: I checked if these functions are called internally and it does not seem like it. So yes, good point. The internal function names can be changed too.

nasl/nasl_init.c
```diff
@@ -193,16 +193,24 @@ static init_func libfuncs[] = {
   {"sleep", nasl_sleep},
   {"isnull", nasl_isnull},
   {"defined_func", nasl_defined_func},
+
+  /* Following 5 entries needed for backwars compatibility */
```
cfi-gb, line 197:
```suggestion
  /* Following 5 entries needed for backwards compatibility */
```

cfi-gb (changes requested):

jjnicola (approved):

cfi-gb (changes requested):
Base branch might be "master" again to avoid introducing new symbols into a stable branch.

ArnoStiefvater:
> Base branch might be "master" again to avoid introducing new symbols into a stable branch.

This should not be a problem. According the @jjnicola this was only an issue in the past because of binary incompatibility. Now nasl/misc libs are build together with openvas.

cfi-gb (commented):
Thanks for clarification. I'm currently not able to do any tests on this but changes looks good from a short look.

cfi-gb (approved):

PR #624, opened 2020-11-13
Correct example parameter / names in function description.

cfi-gb:
Correct two log messages.

Not much to say about that, just check the function names and their parameters to verify.

### DIFF
--- a/nasl/nasl_crypto2.c
+++ b/nasl/nasl_crypto2.c
@@ -607,7 +607,7 @@ ret:
 /**
  * nasl function
  *
- *    DH_compute_key(p:mpi_p, g:mpi_g, dh_server_pub:mpi_server_pub,
+ *    dh_compute_key(p:mpi_p, g:mpi_g, dh_server_pub:mpi_server_pub,
  *                   pub_key:mpi_client_pub, priv_key:mpi_client_priv)
  *
  * Computes the Diffie-Hellman shared secret key from the shared
@@ -761,7 +761,7 @@ strip_pkcs1_padding (tree_cell *retc)
 /**
  * nasl function
  *
- *  rsa_public_encrypt(data:data, e:mpi_e, n:mpi_n, padd:<TRUE:FALSE>)
+ *  rsa_public_encrypt(data:data, e:mpi_e, n:mpi_n, pad:<TRUE:FALSE>)
  *
  * Encrypt the provided data  with the public RSA key given by its parameters e
  * and n. The return value is the encrypted data.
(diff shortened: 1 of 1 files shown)


nasl/nasl_crypto2.c
```diff
@@ -1617,7 +1618,8 @@ encrypt_stream_data (lex_ctxt *lexic, int cipher)
 
   if (!data || datalen <= 0)
     {
-      nasl_perror (lexic, "Syntax: encrypt_data: Missing data or key argument");
+      nasl_perror (lexic, "Syntax: encrypt_stream_data (called from "
+                          "rc4_encrypt): Missing data argument");
```
jjnicola, line 1622: Same here

nasl/nasl_crypto2.c
```diff
@@ -1547,7 +1547,8 @@ nasl_open_stream_cipher (lex_ctxt *lexic, int cipher, int mode)
 
   if (!key || keylen <= 0)
     {
-      nasl_perror (lexic, "Syntax: encrypt_data: Missing data or key argument");
+      nasl_perror (lexic, "Syntax: open_stream_cipher (called from "
+                          "open_rc4_cipher): Missing key argument");
```
jjnicola, line 1551: Despite this function currently is being used only with RC4, the message should not include the cipher name hardcoded, since the function can handle other stream cipher than RC4. If you need some more info, I suggest to include the cipher name in the message, which will give you a hint about the function making the call.
```suggestion
nasl_perror (lexic, "Syntax: open_stream_cipher (called with cipher: %d): "
                    "Missing key argument", gcry_cipher_algo_name (cipher));
```

jjnicola (dismissed):

nasl/nasl_crypto2.c
cfi-gb: Is it possible to somehow pass the name of the calling function to this function here? IMHO the `called with cipher` is far from being optimal because we still don't know which function was exactly called here and can only guess it's name from the cipher then.

jjnicola: Yes, it is possible. You can modify the function signature to receive a new parameter with the caller function name.

nasl/nasl_crypto2.c
```diff
@@ -1530,7 +1530,8 @@ nasl_bf_cbc_decrypt (lex_ctxt *lexic)
  * @return Returns the ID of the cipher handler on success. Otherwise NULL.
  */
 static tree_cell *
-nasl_open_stream_cipher (lex_ctxt *lexic, int cipher, int mode)
+nasl_open_stream_cipher (lex_ctxt *lexic, int cipher, int mode,
```
ArnoStiefvater, line 1535: Add documentation for caller_func param.

nasl/nasl_crypto2.c
```diff
@@ -1602,7 +1606,7 @@ nasl_open_stream_cipher (lex_ctxt *lexic, int cipher, int mode)
  * @return Returns the encrypted data on success. Otherwise NULL.
  */
 static tree_cell *
-encrypt_stream_data (lex_ctxt *lexic, int cipher)
+encrypt_stream_data (lex_ctxt *lexic, int cipher, const char *caller_func)
```
ArnoStiefvater, line 1613: Same. Add doc for caller_func param.

ArnoStiefvater (changes requested):

ArnoStiefvater (approved):

PR #625, opened 2020-11-20
End the scanner in a better way if the host list is malformed.

jjnicola:
**What**:
End the scanner in a better way if the host list is malformed and send a message to the client in case of invalid target list

**Why**:
If the target host list is not empty but malformed, the hosts struct is NULL. Then gvm_hosts_resolve() segfaults

**How**:
Start a scan with gvm-cli (gsa already checks for malformed target) wiht 192.168.0.1/32 as host target.

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/attack.c
+++ b/src/attack.c
@@ -1074,8 +1074,18 @@ attack_network (struct scan_globals *globals)
   max_checks = get_max_checks_number ();
 
   hosts = gvm_hosts_new (hostlist);
-  unresolved = gvm_hosts_resolve (hosts);
+  if (hosts == NULL)
+    {
+      connect_main_kb (&main_kb);
+      message_to_client (
+        main_kb, "Invalid target list.", NULL, NULL, "ERRMSG");
+      kb_lnk_reset (main_kb);
+      g_warning ("Invalid target list. Scan terminated.");
+      set_scan_status ("finished");
+      goto stop;
+    }
 
+  unresolved = gvm_hosts_resolve (hosts);
   while (unresolved)
     {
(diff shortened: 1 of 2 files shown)


bjoernricks (commented):
Would it be possible to include the invalid target list in the error message for the client too?

src/attack.c
```diff
@@ -1074,8 +1074,18 @@ attack_network (struct scan_globals *globals)
   max_checks = get_max_checks_number ();
 
   hosts = gvm_hosts_new (hostlist);
-  unresolved = gvm_hosts_resolve (hosts);
+  if (hosts == NULL)
+    {
+      sprintf (buf, "Invalid target list: %s.", hostlist);
```
ArnoStiefvater, line 1079: Buffer is very small (only 96 bytes) and `hostlist` can get very big. Use `snprinf` or `g_snprintf()` and a bigger buffer if its likely that hostlist will exceed 96 bytes length. You can use `strlen` to get the needed lenght and check it agains the buffer size.
Do we even limit the lenght of a hoststring?

src/attack.c
```diff
@@ -1074,8 +1074,18 @@ attack_network (struct scan_globals *globals)
...
+      connect_main_kb (&main_kb);
+      message_to_client (main_kb, buf, NULL, NULL, "ERRMSG");
```
ArnoStiefvater, line 1081: `message_to_client` has internal buffer with len 2048. Make sure that no buffer overflow can happen in this function  too.

ArnoStiefvater (changes requested):

ArnoStiefvater (approved):
g_strdup_printf is a good solution.

PR #631, opened 2020-12-11
Add timeout argument to ssh_connect() nasl function

jjnicola:
**What**:
Add timeout argument to ssh_connect() nasl function to set the connection timeout in seconds

**Why**:
Some ssh server take too long to return the prompt for the password

**How**:
As I didn't have a ssh server with the issue, I have used the paramiko server demo and added some sleep() before returning the prompt. That simulates the issue in the server side. Then, I used the next nasl script:

` openvas-nasl -X -B -d -i /home/jnicola/install/var/lib/openvas/plugins -t localhost ssh_timeout.nasl`

```
login = "robey";
password = "foo";
port = 2200;

#Sockets
  soc = open_sock_tcp( port );
  if( ! soc ) exit( 0 );
  display ("socket opened:",soc);

#Use the new timeout option. It must be greather than the sleep set in the demo server
  sess = ssh_connect( socket:soc, timeout:15 );
  display("session: ", sess);

if( ! sess )
  return 0;

if( ssh_userauth( sess, login:login, password:password ) ) {
  ssh_disconnect( sess );
  last_sess = 0;
  return 0;
}
```

**Checklist**:

- [ ] Tests

### DIFF
--- a/nasl/nasl_ssh.c
+++ b/nasl/nasl_ssh.c
@@ -242,6 +242,7 @@ nasl_ssh_connect (lex_ctxt *lexic)
   unsigned int tbl_slot;
   int verbose = 0;
   int forced_sock = -1;
+  long timeout; // in seconds
 
   sock = get_int_var_by_name (lexic, "socket", 0);
   if (sock)
@@ -262,6 +263,18 @@ nasl_ssh_connect (lex_ctxt *lexic)
                  nasl_get_function_name (), nasl_get_plugin_filename ());
       return NULL;
     }
+
+  timeout = get_int_var_by_name (lexic, "timeout", 0);
+  if (timeout > 0)
+    if (ssh_options_set (session, SSH_OPTIONS_TIMEOUT, &timeout))
+      {
+        g_message ("Function %s called from %s: "
+                   "Failed to set SSH Connection Timeout to %ld seconds: %s",
+                   nasl_get_function_name (), nasl_get_plugin_filename (), timeout,
+                   ssh_get_error (session));
+        ssh_free (session);
+        return NULL;
+      }
 
   if ((s = getenv ("OPENVAS_LIBSSH_DEBUG")))
     {
--- a/nasl/nasl_ssh.c
+++ b/nasl/nasl_ssh.c
@@ -222,6 +222,9 @@ extern int lowest_socket;
  *
  * - @a scciphers SSH server-to-client ciphers.
  *
+ * - @a timeout Set a timeout for the connection in seconds. Defaults to 10
+ * seconds (defined by libssh internally) if not given.
+ *
  * @naslret An integer to identify the ssh session. Zero on error.
  *
  * @param[in] lexic Lexical context of NASL interpreter.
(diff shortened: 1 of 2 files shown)


nasl/nasl_ssh.c
```diff
@@ -263,6 +264,18 @@ nasl_ssh_connect (lex_ctxt *lexic)
       return NULL;
     }
 
+  timeout = get_int_var_by_name (lexic, "timeout", 0);
```
cfi-gb, line 267: Documentation missing.

Is there also any information available which default SSH connection timeout libssh is used if no `timeout` (SSH_OPTIONS_TIMEOUT) was given?

nasl/nasl_ssh.c
```diff
@@ -263,6 +264,18 @@ nasl_ssh_connect (lex_ctxt *lexic)
...
+  if (timeout > 0)
+    if (ssh_options_set (session, SSH_OPTIONS_TIMEOUT, &timeout))
+      {
+        g_message ("Function %s called from %s: "
+                   "Failed to set SSH Connection Timeout to %ld seconds: %s",
```
cfi-gb, line 272:
```suggestion
                   "Failed to set the SSH connection timeout to %ld seconds: %s",
```

cfi-gb (changes requested):

ArnoStiefvater (approved):

nasl/nasl_ssh.c
```diff
@@ -222,6 +222,8 @@ extern int lowest_socket;
  *
  * - @a scciphers SSH server-to-client ciphers.
  *
+ * - @a timeout Set a timeout for the connection in seconds
```
cfi-gb, line 225:
```suggestion
 * - @a timeout Set a timeout for the connection in seconds.
```
As also asked in https://github.com/greenbone/openvas/pull/631/files#r541548160:

Is there also any information available which default SSH connection timeout libssh is used if no timeout (SSH_OPTIONS_TIMEOUT) was given?

cfi-gb (changes requested):

nasl/nasl_ssh.c
```diff
@@ -222,6 +222,8 @@ extern int lowest_socket;
  *
  * - @a scciphers SSH server-to-client ciphers.
  *
+ * - @a timeout Set a timeout for the connection in seconds.
```
cfi-gb, line 225:
```suggestion
 * - @a timeout Set a timeout for the connection in seconds. Defaults to 10 seconds (defined by libssh internally) if not given.
```

Found the following in https://github.com/libssh/libssh-mirror/blob/cd15043656280e92714c139d820ef271d00876a4/src/misc.c#L1372-L1373:
```
		res = 10 * 1000; /* use a reasonable default value in case
				* SSH_OPTIONS_TIMEOUT is not set in options. */
```

cfi-gb (commented):
Approving the text / description changes.

cfi-gb (approved):

ArnoStiefvater (approved):

PR #641, opened 2021-01-20
Update example redis config

tuto193:
**What**:

`redis-openvas.conf` has been updated and brought to the newest standards, since some settings were either not present at all, or worded wrongly.
**Why**:
Update

**How**:
As stated in the `git log` history:
1. _While comparing this `ridis.conf` against `redis-openvas.conf`_ Settings which were missing (actually missing and not commented out in the `redis-openvas.conf`) were first added.
2. The rewording of some settings (actually just renaming from `slave` to `replica` was done.
3. Check at this step, which of the settings that are active in both differ:
```
Following settings commented out on the redis.conf 5.0
---------------------
Setting         Value(s)
--------------  ---------------------------------
unixsocket      ['/run/redis-openvas/redis.sock']
unixsocketperm  ['770']
syslog-enabled  ['yes']
maxclients      ['10000']

The follwing settings differ in both config files
Setting    redis.conf                   redis-openvas.conf
---------  ---------------------------  ---------------------------------------
port       ['6379']                     ['0']
daemonize  ['no']                       ['yes']
pidfile    ['/var/run/redis_6379.pid']  ['/run/redis-openvas/redis-server.pid']
databases  ['16']                       ['1025']
```
4. Replaced the `redis-openvas.conf` with the `redis.conf` text.
5. Changed/added the settings from `3.` so original behavior is not affected.

**Checklist**:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   [#607](https://github.com/greenbone/openvas/pull/607)
 - Rename some nasl functions and func parameters for consistency and fix byte order issue in get_ipv6_element. [#613](https://github.com/greenbone/openvas/pull/613)
 - Change log level from debug to message to show max_host and max_scan during scan start. [#626](https://github.com/greenbone/openvas/pull/626)
+- Changed the redis-openvas.conf, so that it is compliant with the 5.0+ version(s) of redis. [#641](https://github.com/greenbone/openvas/pull/641)
 
 ### Fixed
 - Fork vhosts before creating the socket.[#576](https://github.com/greenbone/openvas/pull/576)
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -20,7 +20,7 @@ Prerequisites:
 * pkg-config
 * libpcap
 * libgpgme >= 1.1.2
-* redis >= 3.2.0
+* redis >= 5.0.3
 * libssh >= 0.6.0
 * libksba >= 1.0.7
 * libgnutls >= 3.6.4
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   [#607](https://github.com/greenbone/openvas/pull/607)
 - Rename some nasl functions and func parameters for consistency and fix byte order issue in get_ipv6_element. [#613](https://github.com/greenbone/openvas/pull/613)
 - Change log level from debug to message to show max_host and max_scan during scan start. [#626](https://github.com/greenbone/openvas/pull/626)
+- Changed the redis-openvas.conf, so that it is compliant with the 5.0+ version(s) of redis. [#641](https://github.com/greenbone/openvas/pull/641)
```
cfi-gb, line 24: Is there any info about backwards compatibility of these changes available?

If not this might require an update of the INSTALL.md because currently Redis 3.2+ is listed there as a requirement.

tuto193: I am personally in favor of updating the `INSTALL.md`, since we have for the build requirements Debian **Buster**, and in the [debian package archive](https://packages.debian.org/buster/database), `redis 5.0+` is listed as the available version.

jjnicola (changes requested):
Hi @tuto193,
The redis-openvas.conf configuration file looks good. Only that it should be push against the master branch, as raising the required redis version for the stable branch is not desirable.
So, please raise the version in the INSTALL.md file, but also set "master" as target branch.
Thanks in advance.

INSTALL.md
```diff
@@ -20,7 +20,7 @@ Prerequisites:
 * pkg-config
 * libpcap
 * libgpgme >= 1.1.2
-* redis >= 3.2.0
+* redis >= 5.0
```
ArnoStiefvater, line 23:
```suggestion
* redis >= 5.0.3
```

tuto193:
PR moved to https://github.com/greenbone/openvas-scanner/pull/668 so changes can be merged easily in master.

Issue #651, opened 2021-02-09
[20.8.1] gcc 9.3.1 can't build the source code


tuxmaster5000:
The build on CentOS7 fails with:
```
cd /builddir/build/BUILD/openvas-20.8.1/my_build/misc && /usr/lib64/ccache/cc -DOPENVAS_DATA_DIR=\"/usr/share/openvas\" -DOPENVAS_FEED_LOCK_PATH=\"/run/greenbone/feed-update.lock\" -DOPENVAS_STATE_DIR=\"/var/lib/greenbone/lib/openvas\" -DOPENVAS_SYSCONF_DIR=\"/etc/openvas\" -Dopenvas_misc_shared_EXPORTS -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include -I/usr/include/gvm -I/usr/include/gvm-helper  -O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1  -m64 -mtune=generic -D_FILE_OFFSET_BITS=64 -DLARGEFILE_SOURCE=1 -Wall -D_BSD_SOURCE -D_ISOC99_SOURCE -D_SVID_SOURCE -D_DEFAULT_SOURCE -Wall -Wextra -fPIC -g  -Werror -fPIC   -o CMakeFiles/openvas_misc_shared.dir/strutils.c.o   -c /builddir/build/BUILD/openvas-20.8.1/misc/strutils.c
In file included from /usr/include/string.h:638,
                 from /usr/include/glib-2.0/glib/gtestutils.h:30,
                 from /usr/include/glib-2.0/glib.h:82,
                 from /builddir/build/BUILD/openvas-20.8.1/misc/scanneraux.h:28,
                 from /builddir/build/BUILD/openvas-20.8.1/misc/network.h:29,
                 from /builddir/build/BUILD/openvas-20.8.1/misc/pcap.c:22:
In function 'strncpy',
    inlined from 'getinterfaces' at /builddir/build/BUILD/openvas-20.8.1/misc/pcap.c:461:7:
/usr/include/bits/string3.h:120:10: error: '__builtin_strncpy' output may be truncated copying 63 bytes from a string of length 10239 [-Werror=stringop-truncation]
[...]

PR #652, opened 2021-02-09
Build fix for gcc9

tuxmaster5000:
Fix for https://github.com/greenbone/openvas-scanner/issues/651


jjnicola:
Hi @tuxmaster5000,
While the first warnings seems to be a false positive, a case that the compiler can't handle, the other cases can be solved adding the null character '\0' after calling strncpy. The false positive case, I solved it replacing the strncpy with memcpy.
So, closing this PR in favor of #655 
Thank you very much for reporting this.
Best regards.

PR #655, opened 2021-02-11
Fix gcc-9 adn gcc-10 warnings

jjnicola:
**What**:
Add the '\0' after the strncpy call.
For one case, despite the '\0' char is added after strncpy, the compiler raises a warning.
Therefore strncpy is replace with memcpy.

**Why**:
Fix gcc-9 adn gcc-10 warnings

**How**:

**Checklist**:

- [ ] Tests


jjnicola:
Fix for #651

bjoernricks (approved):
Looks good to me

PR #665, opened 2021-02-18
Replace bogus data with a better message and the vendor.

jjnicola:
**What**:
Replace bogus data with a better message and the vendor.

**Why**:
The message "are you dead ?" was used during a scan and it was not
well interpreted.

**How**:
Run a test against a printer and disabled safe checks.

**Checklist**:

- [ ] Tests

### DIFF
--- a/nasl/nasl_misc_funcs.c
+++ b/nasl/nasl_misc_funcs.c
@@ -23,9 +23,10 @@
 
 #include "nasl_misc_funcs.h"
 
-#include "../misc/ftp_funcs.h" /* for ftp_log_in */
-#include "../misc/network.h"   /* read_stream_connection_min */
-#include "../misc/plugutils.h" /* plug_get_host_open_port */
+#include "../misc/ftp_funcs.h"     /* for ftp_log_in */
+#include "../misc/network.h"       /* read_stream_connection_min */
+#include "../misc/plugutils.h"     /* plug_get_host_open_port */
+#include "../misc/vendorversion.h" /* for vendor_version_get */
 #include "byteorder.h"
 #include "exec.h"
 #include "nasl_debug.h"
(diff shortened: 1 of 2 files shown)


nasl/nasl_misc_funcs.c
```diff
@@ -270,8 +272,10 @@ nasl_end_denial (lex_ctxt *lexic)
       if (soc > 0)
         {
           /* Send some data */
-#define BOGUS "are you dead ?"
-          if ((nsend (soc, BOGUS, sizeof (BOGUS) - 1, 0)) >= 0)
+          snprintf (bogus_data, sizeof (bogus_data),
+                    "Network Security Scan In Progress run by %s",
```
cfi-gb, line 276:
```suggestion
                    "Network Security Scan by %s in progress.",
```

Not absolutely sure but that might sound a little bit better.

nasl/nasl_misc_funcs.c
```diff
@@ -243,6 +244,7 @@ nasl_end_denial (lex_ctxt *lexic)
   int to = lexic->recv_timeout;
   struct script_infos *script_infos = lexic->script_infos;
   tree_cell *retc = NULL;
+  char bogus_data[64];
```
cfi-gb, line 247: I'm not really deep in C but isn't this limiting the amount of chars in that string which also means that the message could be stripped if the "vendor" string is really long?

jjnicola: Therefore I leave the string at the end of the message. snprintf() will truncate the vendor string and add the '\0' char at the end. Do you think it would be necessary a bigger/dynamic buffer for the vendor?

jjnicola: solved with https://github.com/greenbone/openvas-scanner/pull/665/commits/b81c402492b0b4bf72f038bcb6affdbdb1e0700d

cfi-gb: Thanks, that looks good. :+1:

Approving the text change but a full code review and a merge needs to be done by a @greenbone/gvm-dev member.

bjoernricks (approved):

PR #668, opened 2021-02-24
New redis config

tuto193:
**What**:

`redis-openvas.conf` has been updated and brought to the newest standards, since some settings were either not present at all, or worded wrongly.
**Why**:
Update

**How**:
As stated in the `git log` history:
1. _While comparing this `ridis.conf` against `redis-openvas.conf`_ Settings which were missing (actually missing and not commented out in the `redis-openvas.conf`) were first added.
2. The rewording of some settings (actually just renaming from `slave` to `replica` was done.
3. Check at this step, which of the settings that are active in both differ:
```
Following settings commented out on the redis.conf 5.0
---------------------
Setting         Value(s)
--------------  ---------------------------------
unixsocket      ['/run/redis-openvas/redis.sock']
unixsocketperm  ['770']
syslog-enabled  ['yes']
maxclients      ['10000']

The follwing settings differ in both config files
Setting    redis.conf                   redis-openvas.conf
---------  ---------------------------  ---------------------------------------
port       ['6379']                     ['0']
daemonize  ['no']                       ['yes']
pidfile    ['/var/run/redis_6379.pid']  ['/run/redis-openvas/redis-server.pid']
databases  ['16']                       ['1025']
```
4. Replaced the `redis-openvas.conf` with the `redis.conf` text.
5. Changed/added the settings from `3.` so original behavior is not affected.

**Checklist**:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   [#607](https://github.com/greenbone/openvas/pull/607)
 - Rename some nasl functions and func parameters for consistency and fix byte order issue in get_ipv6_element. [#613](https://github.com/greenbone/openvas/pull/613)
 - Change log level from debug to message to show max_host and max_scan during scan start. [#626](https://github.com/greenbone/openvas/pull/626)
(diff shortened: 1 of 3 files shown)


jjnicola (approved):

jjnicola:
Thanks @tuto193 for updating this file.

PR #670, opened 2021-03-04
Improve log message for WMI connect failed or missing WMI support.

jjnicola:
**What**:
Both, a WMI connect failed or the WMI stub function return NULL.
The message suggest that the connect failed can be produce because
missing WMI support.

**Why**:
To let now the user that the libopenvas_wmiclient is missing. See [openvas-smb](https://github.com/greenbone/openvas-smb)

**How**:
Build the scanner from scratch, but remove/rename the openvas-smb libs, so you don't have WMI support. Run a scan and check the log.

**Checklist**:

- [ ] Tests


cfi-gb (approved):
LGTM :+1:

ArnoStiefvater (approved):

PR #676, opened 2021-03-11
Don't log all target hosts in a normal message

bjoernricks:
**What**:

Only log target hosts of a scan as debug message.

**Why**:

The list of target hosts can be rather big and is more a debug
information the a normal log message.

**How**:

**Checklist**:

- n/a Tests


jjnicola:
Closing here. It will not merge because this information is considered necessary for each scan, and it is not considered as debug information.

PR #681, opened 2021-03-15
PoC alternative way to scan vhosts

jjnicola:
**What**:
This a PoC to implement an alternative way to scan vhosts. The main idea is to detect vhosts first, identify which plugins must be run against vhosts and create a shorter plugin list. 
Currently, plugins are forked for each vhosts. This means that vhosts are scanned at plugin level, and the results are stored in the host kb.
This PoC present an scan alternative for vhosts, in which the vhosts are scanned at host process level. Once the plugins list for vhosts is done and the vhosts were detected, the host kb is duplicated for the vhosts, the shorter list of plugins is launched against the vhosts. Results are stored/sent to ospd, kb items are stored only in vhost kb, and once the vhost scan finished the vhost kb is delete and redis memory is release

**Why**:
Some host have many vhosts. With current scanner, all data generated for vhosts is stored in the same host kb, and memory is released once the host and its vhosts have been scanned.

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/misc/plugutils.c
+++ b/misc/plugutils.c
@@ -28,7 +28,6 @@
 #include "network.h" // for OPENVAS_ENCAPS_IP
 
 #include <errno.h>               // for errno
-#include <gvm/base/hosts.h>      // for g_vhost_t
 #include <gvm/base/networking.h> // for port_protocol_t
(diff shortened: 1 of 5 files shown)


jjnicola:
Will not be merge.

PR #682, opened 2021-03-15
Add support for volatile keys

ArnoStiefvater:
**What**:

Add support for volatile keys. Add timeout in seconds for keys to expire.
Depends on https://github.com/greenbone/gvm-libs/pull/460.

**Why**:

Can be used in combination with redis settings such as `maxmemory` and `maxmemory-policy` to have a more options in handling memory issues in regards to redis.

**How**:

Stop ospd-openvas if running. 
Do a flushall in redis. `sudo redis-cli -s /run/redis-openvas/redis.sock  FLUSHALL`)
Start nasl script which places keys with expire in redis. `sudo openvas-nasl -X -i <installpath>/var/lib/openvas/plugins/ -t 127.0.0.1 add_volatile_keys.nasl`

```
include("misc_func.inc"); # for rand_str()

i = 0;
key  = "VOLATILE";
value = rand_str(length: 100000);
for (; i<100; i++) {
    set_kb_item(name: key+i, value:value+i, expire:1000000);
}

sleep(10000);
exit(0);
```

Check number of keys in db. `sudo redis-cli -s /run/redis-openvas/redis.sock -n 1 DBSIZE`. Should be 100.
Set maxmemory `config set maxmemory 15728640` in redis.
Set maxmemory-policy `Config set maxmemory-policy volatile-lru` in redis.
Start VT2 which places keys without expire set in other terminal. `sudo openvas-nasl -X -i <installpath>/var/lib/openvas/plugins/ -t 127.0.0.1 persistent_keys.nasl`.

```
include("misc_func.inc"); # for rand_str()

i = 0;
key  = "PERSISTENT";
value = rand_str(length: 100000);
for (; i<100; i++ ) {
    set_kb_item(name: key+i, value:value);
}

sleep(10000);
```

When maxmemory is hit the keys with expire set will get removed by redis to make space for new keys. Check number of keys in db 1 and 2. db 2 should have 100 keys while db 1 should have less.
This way we can make sure which keys can be removed when memory limits are reached.

**Checklist**:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [21.10] (unreleased)
 
 ### Added
+
+- Add support for volatile keys. [#682](https://github.com/greenbone/openvas/pull/682)
+
 ### Changed
 ### Removed
 ### Fixed
--- a/misc/plugutils.c
+++ b/misc/plugutils.c
@@ -684,6 +684,57 @@ plug_set_key (struct script_infos *args, char *name, int type,
   plug_set_key_len (args, name, type, value, 0);
 }
 
+/**
+ * @brief Set volatile key with expire.
+ *
+ * @param args    Script infos.
+ * @param name    Key name.
+ * @param type    Key type.
+ * @param value   Key value.
+ * @param expire  Key expire in seconds.
+ * @param len     Len of value.
+ */
+void
+plug_set_key_len_volatile (struct script_infos *args, char *name, int type,
+                           const void *value, int expire, size_t len)
+{
+  kb_t kb = plug_get_kb (args);
+  int pos = 0; // Append the item on the right position of the list
+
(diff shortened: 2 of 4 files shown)


jjnicola (approved):

jjnicola:
LGTM. 
Do you think it makes sense to set some default values for maxmemory and maxmemory-policy in config/redis-openvas.conf.
Also, it would be nice to document that there is some kind of support/usage of those redis option in doc/redis_config.md

ArnoStiefvater:
> LGTM.
> Do you think it makes sense to set some default values for maxmemory and maxmemory-policy in config/redis-openvas.conf?
> Also, it would be nice to document that there is some kind of support/usage of those redis option in doc/redis_config.md

I would not set a default for maxmemory because we have no possibility knowing the limitation of the machines. This has to be decided individually for the machine at hand. maxmemory-policy does not make sense without maxmemory.

Will add documentation in doc/redis_config.md in a different PR.

PR #695, opened 2021-04-01
￼ Send the a message to the client on iface issues.

jjnicola:
**What**:
￼Send the a message to the client on iface issues. Also set log message to *warning* level.

**Why**:
When an invalid interface is used as source interface for a scan, the task is set as interrupterd without useful information about the problem

**How**:
Create a task and set an invalid iface as source interface. Run the task and check the logs. Now you should see a message:
`sd   main:WARNING:2021-04-01 10h17.28 utc:15037: source_iface: Error with eth0 interface.`

Also, send the message to the client. You should get a message on the client side too.

**Checklist**:

- [ ] Tests
- [X] [CHANGELOG](https://github.com/greenbone/openvas/blob/master/CHANGELOG.md) Entry

### DIFF
--- a/src/attack.c
+++ b/src/attack.c
@@ -875,7 +875,7 @@ apply_source_iface_preference (void)
     {
       gchar *msg =
         g_strdup_printf ("Erroneous source interface: %s", source_iface);
-      g_debug ("source_iface: Error with %s interface.", source_iface);
+      g_warning ("source_iface: Error with %s interface.", source_iface);
 
       g_free (msg);
       return -2;
--- a/src/attack.c
+++ b/src/attack.c
@@ -888,7 +888,6 @@ apply_source_iface_preference (void)
 
       g_free (ipstr);
       g_free (ip6str);
-      ret = 0;
     }
 
   if (msg)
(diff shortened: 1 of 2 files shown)


src/attack.c
```diff
@@ -890,8 +888,20 @@ apply_source_iface_preference (void)
 
       g_free (ipstr);
       g_free (ip6str);
-      return 0;
+      ret = 0;
```
ArnoStiefvater, line 891:
```suggestion
```
Not needed. Ret is already 0.

src/attack.c
```diff
@@ -890,8 +888,20 @@ apply_source_iface_preference (void)
...
     }
+
+  if (msg)
```
ArnoStiefvater, line 894:
```suggestion
  if (ret < 0 && msg)
```
Maybe also use the ret value here.

src/attack.c
jjnicola: Fixed with https://github.com/greenbone/openvas-scanner/pull/695/commits/e36f8545effc55b769d772c936d38bff1a520909

src/attack.c
jjnicola: Not necessary. It would be redundant, as msg will be != NULL, if and only if ret is less than 0.

ArnoStiefvater (approved):

PR #701, opened 2021-04-11
Fix greenbone-nvt-sync on (at least) FreeBSD

stynoo:
Two error messages when running greenbone-nvt-sync on FreeBSD.  (default shell, fresh install in an iocage)
- "chmod: invalid file mode: +660"
        chmod uses or numeric values or the +/- sign and symbolic notation.  
- "exit: greenbone-nvt-sync.vX4hC9RFi9: not found"
        mktemp [-d] [-t prefix] should be used

After these changes the script does not return errors and "md manage:WARNING:2021-04-11 17h41.07 UTC:85430: /usr/local/bin/greenbone-nvt-sync returned a non-zero exit code." is no longer present in the log files.

### DIFF
--- a/tools/greenbone-nvt-sync.in
+++ b/tools/greenbone-nvt-sync.in
@@ -202,7 +202,7 @@ if [ -z "$TMPDIR" ]; then
   SYNC_TMP_DIR=/tmp
   # If we have mktemp, create a temporary dir (safer)
   if [ -n "`which mktemp`" ]; then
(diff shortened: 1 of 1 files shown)


jjnicola:
It seems this is an issue only on FreeBSD. I tried the PR in my environment and it does not affect the script behavior in a Debian reference system.
Only that in Debian the `+` adds permission to the current setup, while your changes replace the current with 660. I don't see a problem here.
Please add a changelog entry

jjnicola (changes requested):
Missing changelog entry

y0urself:
@stynoo will you provide a Changelog entry for your changes?

PR #715, opened 2021-04-26
Fix hosts count when reverse_lookup_only preferences is enabled.

jjnicola:
**What**:
Fix hosts count when reverse_lookup_only preferences is enabled.

**Why**:
When this option is enabled, hosts without a hostname are removed from the host list, which changes the host count.
As this differs from the host expected by ospd-openvas, finishes the scan as INTERRUPTED.

**How**:
Run a scan (I tested via gvm-tools) with reverse_lookup_only enabled. Also, I added an excluded host, as this part of the code was modified as well. Ensure that some hosts in the target list can be reversed-lookup'ed.

gvm-cli --protocol OSP --timeout 120 socket --socketpath=/home/jnicola/install/var/run/ospd/openvas.sock --xml "<start_scan scan_id='829097a9-85d5-4bb8-bac0-e64c362b2836'><targets><target><hosts>192.168.0.0/24</hosts><exclude_hosts>192.168.0.10</exclude_hosts><finished_hosts/><ports>22</ports><credentials><credential port='22' service='ssh' type='up'><username>test</username><password>test</password></credential></credentials><reverse_lookup_only>1</reverse_lookup_only></target></targets><scanner_params></scanner_params><vt_selection><vt_single id='1.3.6.1.4.1.25623.1.0.14259'/></vt_selection></start_scan>" |xmlstarlet fo

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/attack.c
+++ b/src/attack.c
@@ -731,13 +731,30 @@ attack_start (struct attack_start_args *args)
 }
 
 static void
-apply_hosts_preferences (gvm_hosts_t *hosts)
+apply_hosts_excluded (gvm_hosts_t *hosts)
 {
-  const char *ordering = prefs_get ("hosts_ordering"),
-             *exclude_hosts = prefs_get ("exclude_hosts");
+  const char *exclude_hosts = prefs_get ("exclude_hosts");
+
+  /* Exclude hosts ? */
+  if (exclude_hosts)
+    {
+      /* Exclude hosts, resolving hostnames. */
+      int ret = gvm_hosts_exclude (hosts, exclude_hosts);
+
+      if (ret > 0)
+        g_message ("exclude_hosts: Skipped %d host(s).", ret);
+      if (ret < 0)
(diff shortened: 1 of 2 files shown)


src/attack.c
```diff
@@ -731,13 +731,30 @@ attack_start (struct attack_start_args *args)
...
+      /* Exclude hosts, resolving hostnames. */
+      int ret = gvm_hosts_exclude (hosts, exclude_hosts);
+
+      if (ret > 0)
+        g_message ("exclude_hosts: Skipped %d host(s).", ret);
+      if (ret < 0)
+        g_message ("exclude_hosts: Error.");
+    }
 
   if (hosts == NULL)
```
ArnoStiefvater, line 750: Does not seem useful here. We return nonetheless. Should probably be at the top of the function.
This check might be needed at the top of the other new functions as well?

ArnoStiefvater (approved):

PR #730, opened 2021-05-10
Remove source iface preferences

ArnoStiefvater:
**What**:

Remove source iface preferences.
source_iface preferences can still be set for nasl standalone for development purposes.

**Why**:

Remove rarely used option.

**How**:

Add options sys_ifaces_allow and sys_ifaces_deny to the openvas conf file and start a small scan via gvm-cli with options:

```
        <scanner_params>\
            <source_iface>enp0s8</source_iface>\
            <ifaces_deny>enp0s8</ifaces_deny>\
            <ifaces_allow>foo</ifaces_allow>\
        </scanner_params>\
```

In the results the following error result will appear: `The following provided settings are deprecated since the 21.10 release and will be ignored: source_iface (task setting) ifaces_
allow (user setting) ifaces_deny (user setting) sys_ifaces_allow (scanner only setting)sys_ifaces_deny (scanner only setting)`

**Checklist**:

- [ ] Tests


src/attack.c
```diff
@@ -627,6 +627,47 @@ vhosts_to_str (GSList *list)
   return g_string_free (string, FALSE);
 }
 
+/**
+ * @brief Check if any deprecated prefs are in pref table and print warning.
+ *
+ * @return True if deprecated prefs are in preference table. Else False.
+ */
+gboolean
```
jjnicola, line 635: Shouldn't this be static? Shouldn't be just a void function? the return value is not used, and the message is sent directly from the function.

jjnicola (changes requested):

jjnicola (approved):

PR #742, opened 2021-05-18
Sentry integration

jjnicola:
Depends on greenbone/gvm-libs#502

**What**:
Integrate Sentry

**Why**:

**How**:

**Checklist**:

- [ ] Tests


timopollmeier (commented):
I think you forgot to commit the debug_utils files here.
Also, the message about sentry in `init_openvas` looks like it's missing a check whether Sentry is actually enabled.

Kraemii (dismissed):

Kraemii (approved):

PR #806, opened 2021-07-16
Notus integration mqtt

jjnicola:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/attack.c
+++ b/src/attack.c
@@ -28,6 +28,7 @@
 #include "../misc/network.h"        /* for auth_printf */
 #include "../misc/nvt_categories.h" /* for ACT_INIT */
 #include "../misc/pcap_openvas.h"   /* for v6_is_local_ip */
+#include "../misc/plugutils.h"      /*for make_table_driven_lsc_info_json_str */
 #include "../nasl/nasl_debug.h"     /* for nasl_*_filename */
 #include "hosts.h"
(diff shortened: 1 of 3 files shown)


ArnoStiefvater (changes requested):
Just a comment so nobody merges accidentally before I could look at it.

src/attack.c
```diff
@@ -339,6 +341,89 @@ check_new_vhosts (void)
...
+/**
+ * @brief Publish the necessary data to start a Table driven LSC scan.
+ *
+ * If the gather-package-list.nasl plugin was launched, and it generated
+ * a valid package list for a supported OS, the table driven LSC scan
+ * which is subscribed to the topic will perform a scan an publish the
+ * the results to be handle by the sensor/client.
+ */
+static void
+run_table_driven_lsc (const char *scan_id, kb_t kb, const char *ip_str,
```
ArnoStiefvater, line 353: I would move all the json handling to plugutils.c. Maybe the function could be calles something like `scanner_get_table_driven_lsc_start_str(...);` and be placed at the top where the other `scanner_..` function is. /misc already has json support, so no changes to the cmake needed.

The  `run_table_driven_lsc` function can stay in attack.c I think.

ArnoStiefvater (approved):

PR #807, opened 2021-07-21
[master] Refactor sending of dead host status 

ArnoStiefvater:
**What**:

Refactor sending of dead host status. Its only a first small improvement. This will be improved more in the future.

**Why**:

Separate sending of progress status and sending of
this special status case.

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/attack.c
+++ b/src/attack.c
@@ -156,22 +156,69 @@ set_scan_status (char *status)
 }
 
 /**
- * @brief Sends the status of a host's scan.
+ * @brief Send status to the client that the host is dead
+ *
+ * Originally the progress status is of the format
+ * "current_host/launched/total". Current host is the ip_str of the current host
+ * which is vulnerability tested. Launched is the number of plguins(VTs) which
+ * got already started. Total is the total number of plugins which will be
+ * started for the current host. But here we use the format "current_host/0/-1"
+ * for implicit singalling that the host ist dead.
+ *
+ * @param main_kb Kb to use
+ * @param ip_str str representation of host ip
+ *
+ * @return 0 on success, -1 on failure.
  */
 static int
-comm_send_status (kb_t main_kb, char *hostname, int curr, int max)
+comm_send_status_host_dead (kb_t main_kb, char *ip_str)
 {
-  char buffer[2048];
+  // implicit status code. Originally launched/total plugins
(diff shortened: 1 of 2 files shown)


src/attack.c
```diff
@@ -156,22 +156,69 @@ set_scan_status (char *status)
...
+ * started for the current host. But here we use the format "current_host/0/-1"
+ * for implicit singalling that the host ist dead.
+ *
+ * @param main_kb Kb to use
+ * @param hostnae ip_str representation of host
+ *
+ * @return 0 on success, -1 on failure.
+ */
+static int
+comm_send_status_host_dead (kb_t main_kb, char *hostname)
```
jjnicola, line 174: I would use ip_str instead of hostname, since we are using the IP. Also, this change would fix the typo in the docstring.
```suggestion
comm_send_status_host_dead (kb_t main_kb, char *ip_str)
```

src/attack.c
```diff
@@ -156,22 +156,69 @@ set_scan_status (char *status)
...
+ *
+ * Originally the progress status is of the format
+ * "current_host/launched/total". Current host is the ip_str of the current host
+ * which is vulnerability tested. Launched is the number of plguins(VTs) which
+ * got already started. Total is the total number of plugins which will be
+ * started for the current host. But here we use the format "current_host/0/-1"
+ * for implicit singalling that the host ist dead.
+ *
+ * @param main_kb Kb to use
+ * @param hostnae ip_str representation of host
```
jjnicola, line 169: use ip_str.
```suggestion
 * @param ip_str ip_str representation of host
```

src/attack.c
```diff
@@ -511,10 +558,7 @@ attack_host (struct scan_globals *globals, struct in6_addr *ip, GSList *vhosts,
                     "<description/><type/><name/></source></detail></host>",
                     ip_str);
 #if (PROGRESS_BAR_STYLE == 1)
```
jjnicola, line 560: I think we can remove this preprocessor directive. It is always set to 1 and other modules does not hanlde other option than this for dead hosts.. The DEFINE is above in the same file.

jjnicola (changes requested):

jjnicola (approved):

PR #809, opened 2021-07-26
Add vt info to results

jjnicola:
DEPENDS on greenbone/gvm-libs#568

**What**:
Results format is changed and now include the fields
- name, the script name
- QoD, quality of detection
- Score, calculated from the vector found in the vt.

Those fields are optional and will be added only if there is an OID

**Why**:
This task is done currently by ospd-openvas and must be moved to openvas.

**How**:
- Use mosquitto_sub to subscribe the topic "scanner/results"
- Run a small scan.
- Check that results contains a severity, qod and name fields.

```
scanner/results {"scan_id":"829097a9-85d5-4bb8-bac0-e64c362b2836","type":"LOG","host_ip":"192.168.123.52","hostname":"192.168.0.1","port":"22/tcp","OID":"1.3.6.1.4.1.25623.1.0.90022","name":"SSH Authorization Check","qod":"80","severity":"0.0","value":"It was possible to login using the provided SSH credentials. Hence authenticated checks are enabled.\n","uri":""}
```

**Checklist**:

- [ ] Tests


misc/plugutils.c
```diff
@@ -405,6 +407,104 @@ make_table_driven_lsc_info_json_str (const char *scan_id, const char *ip_str,
...
+ *
+ * OpenVAS result contains also the nvti name, QoD and score, which need to be
+ * taken from redis and the score must be calculated from the vector
+ *
+ * @param[in]  oid     The oid of the NVT.
+ * @param[out] builder Json builder to add data to.
+ *
+ */
+static void
+add_nvti_info_into_json_builder (const char *oid, JsonBuilder *builder)
```
ArnoStiefvater, line 478:
```suggestion
static JsonBuilder *
json_builder_add_nvti_info (JsonBuilder *builder, const char *oid)
```
Maybe something like that for consistency.

misc/nvt_qod.c
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * @brief GLib logging domain.
+ */
+#define G_LOG_DOMAIN "lib  misc"
+
+static const struct {
+  qod_val val;
+  const char *type;
+} qod_types[] = {
+                 {EXPLOIT, "exploit"},
```
ArnoStiefvater, line 40: Where can I look all these qod_types up?

jjnicola: I was not able to find any documentation for qod_types. I created it first for [ospd-openvas](https://github.com/greenbone/ospd-openvas/blob/86d6637b91e779e1149301cc11c2888e0b3b000f/ospd_openvas/nvticache.py#L42) some time ago, based on [gvmd](https://raw.githubusercontent.com/greenbone/gvmd/master/src/manage_sql.c)

ArnoStiefvater (approved):

ArnoStiefvater (approved):

PR #818, opened 2021-08-02
Fix Workflow: Quote name

y0urself:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Tests


.github/workflows/docker-sched.yml
```diff
@@ -1,4 +1,4 @@
-name: [Scheduled] Build and upload docker images
+name: "[Scheduled] Build and upload docker images"
```
ArnoStiefvater, line 1: Is this only needed because of the `[]`? There are a lot of other `name` which are not quoted.

ArnoStiefvater (approved):

PR #826, opened 2021-08-05
Changed defaults for installation locations

bjoernricks:
**What**:

* SYSCONFDIR is /etc by default now
* LOCALSTATEDIR is /var by default now
* OPENVAS_RUN_DIR is /run/ospd by default now
* OPENVAS_FEED_LOOK_PATH is /var/lib/openvas/feed-update.lock by default
  now

/run/ospd is used because normally openvas is started from ospd-openvas
which by default uses the /run/ospd path in the systemd service file.
This ensures that the directory for the OPENVAS_RUN_DIR exists at
runtime.

/var/lib/openvas/feed-update.lock is used because the sync scripts may
be started when the services are stopped an /run/ospd (or an other
directory in /run) is not available anymore.

**Why**:

Use better and sane defaults.

**How**:

**Checklist**:

- [ ] Tests


CMakeLists.txt
```diff
@@ -161,24 +161,25 @@ if (NOT LIBDIR)
 endif (NOT LIBDIR)
 
 if (NOT LOCALSTATEDIR)
-  set (LOCALSTATEDIR "${CMAKE_INSTALL_PREFIX}/var")
+  set (LOCALSTATEDIR "/var")
 endif (NOT LOCALSTATEDIR)
 
 if (NOT DATADIR)
   set (DATADIR "${CMAKE_INSTALL_PREFIX}/share")
```
y0urself, line 168: Isn't it inconsistent to use the install prefix here, but remove it elsewhere?

y0urself (approved):

PR #832, opened 2021-08-06
Fix interrupted scans

jjnicola:
SC-177

**What**:
Check if the max running process was reached or even exceed, when waiting for running a new plugin. 

**Why**:
Fix an issue which produced interrupted scan, because a host not being scan 100%

**How**:
Not easy to reproduce. It depends on the system, scan config and mainly on the target system (some plugins must be trigger in a especial order). Good luck.

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/attack.c
+++ b/src/attack.c
@@ -56,7 +56,6 @@
 #include <unistd.h>   /* for close() */
 
 #define ERR_HOST_DEAD -1
-#define ERR_CANT_FORK -2
 
 #define MAX_FORK_RETRIES 10
 /**
@@ -346,13 +345,14 @@ check_new_vhosts (void)
  * Does not launch a plugin twice if !save_kb_replay.
  *
  * @return ERR_HOST_DEAD if host died, ERR_CANT_FORK if forking failed,
- *         0 otherwise.
+ *         ERR_NO_FREE_SLOT if the process table is full, 0 otherwise.
  */
 static int
 launch_plugin (struct scan_globals *globals, struct scheduler_plugin *plugin,
                struct in6_addr *ip, GSList *vhosts, kb_t kb, kb_t main_kb)
 {
-  int optimize = prefs_get_bool ("optimize_test"), pid, ret = 0;
+  int optimize = prefs_get_bool ("optimize_test");
+  int launch_error, pid, ret = 0;
   char *oid, *name, *error = NULL, ip_str[INET6_ADDRSTRLEN];
   nvti_t *nvti;
 
@@ -428,14 +428,13 @@ launch_plugin (struct scan_globals *globals, struct scheduler_plugin *plugin,
 
   /* Update vhosts list and start the plugin */
   check_new_vhosts ();
-  pid = plugin_launch (globals, plugin, ip, vhosts, kb, main_kb, nvti);
-  if (pid < 0)
+  launch_error = 0;
+  pid = plugin_launch (globals, plugin, ip, vhosts, kb, main_kb, nvti,
+                       &launch_error);
+  if (launch_error == ERR_NO_FREE_SLOT || launch_error == ERR_CANT_FORK)
     {
       plugin->running_state = PLUGIN_STATUS_UNRUN;
-      if (pid == ERR_NO_FREE_SLOT)
-        ret = ERR_NO_FREE_SLOT;
-      else
-        ret = ERR_CANT_FORK;
+      ret = launch_error;
       goto finish_launch_plugin;
     }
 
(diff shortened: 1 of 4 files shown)


src/pluginlaunch.c
```diff
@@ -402,7 +402,15 @@ plugin_launch (struct scan_globals *globals, struct scheduler_plugin *plugin,
   pluginlaunch_wait_for_free_process (main_kb);
   p = next_free_process (main_kb, plugin);
   if (p < 0)
-    return -1;
+    {
+      g_debug ("There are currently no free slot for running a new plugins, "
```
bjoernricks, line 406: Either `a new plugin` or `new plugins`. Maybe
```suggestion
      g_debug ("There are currently no free slot for starting a new plugin, "
```

src/attack.c
```diff
@@ -519,13 +522,21 @@ attack_host (struct scan_globals *globals, struct in6_addr *ip, GSList *vhosts,
                   kb_item_push_str (main_kb, "internal/results", buffer);
                   goto host_died;
                 }
+              else if (e == ERR_NO_FREE_SLOT)
+                {
+                  g_debug (
+                    "fork() failed for %s. Not free slot to run a plugin.",
```
cfi-gb, line 528:
```suggestion
                    "fork() failed for %s. No free slot to run a plugin.",
```

src/pluginlaunch.c
```diff
@@ -402,7 +402,15 @@ plugin_launch (struct scan_globals *globals, struct scheduler_plugin *plugin,
   pluginlaunch_wait_for_free_process (main_kb);
   p = next_free_process (main_kb, plugin);
   if (p < 0)
-    return -1;
+    {
+      g_debug ("There are currently no free slot for starting a new plugin, "
```
cfi-gb, line 406: Either `are currently no free slots` or `is currently no free slot` and both probably needs an `available` as well.

src/pluginlaunch.c
```diff
@@ -461,8 +469,15 @@ pluginlaunch_wait_for_free_process (kb_t kb)
   update_running_processes (kb);
   /* Max number of processes are still running, wait for a child to exit or
    * to timeout. */
+
+  if (num_running_processes >= max_running_processes)
+    g_debug ("%s. Number of running processes >= maximum running processes (%d "
+             ">= %d). "
+             "Waitting for free slot for processes.",
```
cfi-gb, line 476:
```suggestion
             "Waiting for free slot for processes.",
```

src/pluginlaunch.c
```diff
@@ -402,7 +402,15 @@ plugin_launch (struct scan_globals *globals, struct scheduler_plugin *plugin,
   pluginlaunch_wait_for_free_process (main_kb);
   p = next_free_process (main_kb, plugin);
   if (p < 0)
-    return -1;
+    {
+      g_debug ("There are currently no free slot for starting a new plugin, "
+               "probably because the parallel check is temporally disabled. "
```
cfi-gb, line 407:
```suggestion
               "probably because the parallel check is temporarily disabled. "
```

cfi-gb (changes requested):

src/attack.c
```diff
@@ -432,7 +432,10 @@ launch_plugin (struct scan_globals *globals, struct scheduler_plugin *plugin,
   if (pid < 0)
     {
       plugin->running_state = PLUGIN_STATUS_UNRUN;
-      ret = ERR_CANT_FORK;
+      if (pid == ERR_NO_FREE_SLOT)
```
ArnoStiefvater, line 435: Update `launch_plugin()` documentation.
Maybe `plugin_launch()` should get a new error parameter so we return pid on success and set and error on error.
This way we do not mix pid value and error value.

src/attack.c
```diff
@@ -519,13 +522,21 @@ attack_host (struct scan_globals *globals, struct in6_addr *ip, GSList *vhosts,
                   kb_item_push_str (main_kb, "internal/results", buffer);
                   goto host_died;
                 }
+              else if (e == ERR_NO_FREE_SLOT)
+                {
+                  g_debug (
+                    "fork() failed for %s. No free slot to run a plugin.",
```
ArnoStiefvater, line 528: `fork()` was never tried here. ERR_NO_FREE_SLOT is returned long before `create_process()` is called.
Maybe better debug message would be  `Launch failed for %s. No free slot to run a plugin."`

src/attack.c
```diff
@@ -519,19 +521,32 @@ attack_host (struct scan_globals *globals, struct in6_addr *ip, GSList *vhosts,
                   kb_item_push_str (main_kb, "internal/results", buffer);
                   goto host_died;
                 }
+              else if (e == ERR_NO_FREE_SLOT)
+                {
+                  if (forks_retry < MAX_FORK_RETRIES)
+                    {
+                      forks_retry++;
+                      g_warning ("Launch failed for %s. No free slot available "
```
ArnoStiefvater, line 529:
```suggestion
                      g_warning ("Launch failed for %s. No free slot in the internal process table available "
```
Makes it more explicit what is meant with `slot`.

ArnoStiefvater (approved):

PR #841, opened 2021-08-12
Store the timeout as preference

jjnicola:
Closes SC-354.

Depends on greenbone/gvm-libs#578

**What**:
Store and get the script timeout as preference

**Why**:
The script timeout is now handle as preference and it is stored as another preference, but always with the ID 0.

**How**:
Use an up to date build of middleware from gvmd (send the script preference) and gvm-libs (handle timeout as preference)
Set a timeout to 1second for a script in gsa, one which runs for a long time (find_service.nasl). Run the scanner. Check that the script timeouts. You can check this with "log_whole_attack = yes" 

A redis flushall() is required, since it is a change in the redis nvticache and all plugins must be loaded up again.

**Checklist**:

- [ ] Tests

### DIFF
--- a/nasl/nasl_scanner_glue.c
+++ b/nasl/nasl_scanner_glue.c
@@ -82,6 +82,18 @@ isalldigit (char *str, int len)
  * scanner.
  */
 
+/**
+ * @brief Add timeout preference to VT preferences
+ *
+ * VT timeout is handled as normal VT preference.
+ * Because of backward compatibility issues the timeout preference is always
+ * located at the VT pref location with id NVTPREF_TIMEOUT_ID.
+ *
+ * @param[in] lexic   lexic
+ * @param[in] to      script timeout
+ *
+ * @return FAKE_CELL
+ */
 tree_cell *
 script_timeout (lex_ctxt *lexic)
 {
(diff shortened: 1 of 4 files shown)


nasl/nasl_scanner_glue.c
```diff
@@ -87,11 +87,16 @@ script_timeout (lex_ctxt *lexic)
...
+  nvtpref_t *np;
+  gchar *timeout;
 
   if (to == -65535)
     return FAKE_CELL;
 
-  nvti_set_timeout (nvti, to ? to : -1);
+  timeout = g_strdup_printf ("%d", to);
+
+  np = nvtpref_new (0, "timeout", "entry", timeout);
```
ArnoStiefvater, line 98:
```suggestion
  np = nvtpref_new (SCRIPT_TIMEOUT_PREF_ID, "timeout", "entry", timeout);
```
Add SCRIPT_TIMEOUT_PREF_ID to gvm-libs/base/nvti.c

Add function documentation like
```
/**
 * @brief Add timeout preference to VT preferences
 * 
 * VT timeout is handled as normal VT preference.
 * Because of backward compatibility issues the timeout preference is always
 * located at the VT pref location with id SCRIPT_TIMEOUT_PREF_ID. 
 * 
 * @param[in] lexic   lexic
 * @param[in] to      script timeout
 * 
 * @return FAKE_CELL
 */
```

ArnoStiefvater (approved):

PR #842, opened 2021-08-13
Data loop

nichtsfrei:
This is basically PR838 with a couple of fixes:
- change scan.status => status.scan (unlike in the topic definitions functions is before aggregate within the message type)
- add missing fields to result


src/openvas.c
```diff
@@ -192,82 +195,245 @@ init_signal_handlers (void)
...
+ * prefs from client.
+ *
+ * @param globals Scan ID of globals used as key to find the corresponding KB
+ * where to take the preferences from.
+ *
+ * @return 0 on success, -1 if the kb is not found or no prefs are found in
+ *         the kb.
+ */
+static int
+overwrite_openvas_prefs_with_prefs_from_client (struct scan_globals *globals)
```
ArnoStiefvater, line 391: It would be nice if the original function could not be touched at all at this point.
Just put the functionality in a new function and call that one instead of `overwrite_openvas_prefs_with_prefs_from_client()`.
I know it means dead code but I do think for now it would be nice to have it as reference for how openvas handles the prefs currently because there are still some unknowns here.
You can use `__attribute__ ((unused))` if the compiler complains. 

Alternatively this function could be just touched minimally with `if/else mqtt_is_initialized()` eulabeia switch.

nichtsfrei: Wouldn't it be better to run in those unknown cases explicitly to make them known instead of keeping it around just in case?

ArnoStiefvater: Maybe I was not really clear what I meant. The cases are known.

For example before the changes there is the logic which at the end does `store_file()`. But this PR just deletes this functionality without replacement and without documenting that the functionality was now removed. So the information about what we still need to implement in the future gets lost on code side.
For our start scan example this PR is enough. But we still need to handle certificates and so on which are provided as file and stored as such (`store_file`) at the end. I thought handling it the way I proposed may be the easiest and fastest hacky solution for now so that functionality which we still need at the end is not getting lost in the void.

nichtsfrei: Maybe? Since this pr is based on another pr and I don't know the internal certificate handling I'm still not 100% sure what you have in mind. Probably it makes sense that you could add a missing comment?

src/openvas.c
```diff
@@ -382,6 +776,48 @@ stop_single_task_scan (void)
...
+ * @brief Creates the main kb and inserts the scan id
+ *
+ * @param scan_id
+ * @return int 0 on success, -1 on failure
+ */
+int
+create_main_kb (const char *scan_id)
+{
+  kb_t main_kb;
+  printf ("db_address: %s\n", prefs_get ("db_address"));
```
ArnoStiefvater, line 789: Left over debug message? We don't use printf here. Use g_debug if needed.

src/openvas.c
```diff
@@ -382,6 +776,48 @@ stop_single_task_scan (void)
...
+  if (kb_new (&main_kb, prefs_get ("db_address") ?: KB_PATH_DEFAULT))
```
ArnoStiefvater, line 790:
```suggestion
  if (kb_new (&main_kb, prefs_get ("db_address")))
```
The defaults were already set previously.

src/attack.c
```diff
@@ -1112,28 +1184,27 @@ attack_network (struct scan_globals *globals)
...
       return;
     }
-
   if (plugins_init_error > 0)
     {
-      sprintf (buf,
-               "%d errors were found during the plugin scheduling. "
-               "Some plugins have not been launched.",
-               plugins_init_error);
+      snprintf (buf, 96,
```
ArnoStiefvater, line 1189: Use g_strdup_printf instead.

src/attack.c
```diff
@@ -1144,22 +1215,20 @@ attack_network (struct scan_globals *globals)
...
   g_slist_free_full (unresolved, g_free);
-
   /* Apply Hosts preferences. */
   apply_hosts_preferences_ordering (hosts);
   apply_hosts_reverse_lookup_preferences (hosts);
 
   /* Send the hosts count to the client, after removing duplicated and
    * unresolved hosts.*/
-  sprintf (buf, "%d", gvm_hosts_count (hosts));
+  snprintf (buf, 96, "%d", gvm_hosts_count (hosts));
```
ArnoStiefvater, line 1231: Use g_strdup_printf

misc/plugutils.c
```diff
@@ -502,13 +488,15 @@ proto_post_wrapped (const char *oid, struct script_infos *desc, int port,
   data = g_convert (buffer, -1, "UTF-8", "ISO_8859-1", NULL, &length, NULL);
 
   /* Send result via MQTT. */
+  context = prefs_get ("mqtt_context");
+  snprintf (topic, 128, "%s/scan/info", context);
```
ArnoStiefvater, line 492: Use g_strdup_printf.

ArnoStiefvater (approved):

PR #852, opened 2021-08-31
Change: Instead of using g_memdup2 set deprecation for it as warning

nichtsfrei:
Since g_memdup2 will use a 32bit number instead of a 64bit there could
be the case that it will allocate less memory than anticipated could due
to lack of control end up in an overflow scenario.

To prevent that we will leave it as g_memdup but reduce the severity
down to a warning instead of an error.


misc/network.c
```diff
@@ -737,6 +737,9 @@ socket_negotiate_ssl (int fd, openvas_encaps_t transport,
  * @param[out]  cert    Memory pointer to fill cert pointer.
  * @param[out]  certlen Size of cert.
  */
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic warning "-Wdeprecated-declarations"
```
ArnoStiefvater, line 742: `#pragma GCC diagnostic pop` missing after the g_memdup call. 
And I would only enclose the `g_memdup` call and not the complete function with the pragma.
Same for all the other uses.

ArnoStiefvater (dismissed):

misc/network.c
nichtsfrei: After enclosing just the call it doesn't compile; therefore will leave it on function level.

nasl/nasl_packet_forgery.c
```diff
@@ -1898,8 +1897,11 @@ get_icmp_element (lex_ctxt *lexic)
           retc->size =
             get_var_size_by_name (lexic, "icmp") - (ip->ip_hl * 4) - 8;
           if (retc->size > 0)
+#pragma GCC diagnostic push
```
y0urself, line 1900: you will need to enclose this if statement with brackets I guess ...

y0urself (dismissed):

y0urself (dismissed):

y0urself (approved):

jjnicola (approved):

PR #853, opened 2021-08-31
Add nasl function sftp_enabled_check()

jjnicola:
**What**:
Add nasl function sftp_enabled_check() to check if sftp subsystem is enabled in the target

Jira: SC-376

**Why**:
To check if sftp subsystem is enabled in the target

**How**:
Create  the following nasl script sftp_test.nasl:
```
port = 22;
user = 'user';
pass = 'pass';

#Socket

soc1 = open_sock_tcp( port );
if( ! soc1 ) exit( 0 );
display ("socket opened:",soc1);

# New session
sess1 = ssh_connect( socket:soc1 );
display("session1: ", sess1);

# User auth.
ssh_userauth(sess1, login:user, password:pass);

#Check SFTP
ret = sftp_enabled_check (sess1);

display("SFTP: ", ret);

# Close session
a = ssh_disconnect(sess1);
```
Disable SFTP subsystem in the target
- Comment out sftp support in sshd_config and of course restart sshd:
`#Subsystem sftp /usr/lib/openssh/sftp-server`

Run the script:
``` openvas-nasl -X -B -d -i /path/to/var/lib/openvas/plugins -t <TARGET> sftp_test.nasl```

Re-enable SFTP and re run the test.

**Checklist**:

- [ ] Tests

### DIFF
--- a/nasl/nasl_ssh.c
+++ b/nasl/nasl_ssh.c
@@ -1903,7 +1903,7 @@ nasl_ssh_shell_close (lex_ctxt *lexic)
  */
 
 /**
- * @brief Check if SFTP is enabled.
+ * @brief Check if the SFTP subsystem is enabled on the remote SSH server.
  * @naslfn{sftp_enabled_check}
  *
  * @nasluparam
@@ -1911,7 +1911,8 @@ nasl_ssh_shell_close (lex_ctxt *lexic)
  * - An SSH session id.
  *
  * @naslret An integer: 0 on success, -1 (SSH_ERROR) on Channel request
- * subsystem failure. Greather than 0 means an error during SFTP init.
+ * subsystem failure. Greater than 0 means an error during SFTP init. NULL
+ * indicates a failure during session id verification.
  *
  * @param[in] lexic Lexical context of NASL interpreter.
  */
--- a/nasl/nasl_ssh.c
+++ b/nasl/nasl_ssh.c
@@ -504,7 +504,7 @@ do_nasl_ssh_disconnect (int tbl_slot)
  *
  * @nasluparam
  *
- * - An ssh session id.  A value of 0 is allowed and acts as a NOP.
+ * - An SSH session id.  A value of 0 is allowed and acts as a NOP.
  *
  * @naslret Nothing
  *
@@ -612,7 +612,7 @@ nasl_ssh_session_id_from_sock (lex_ctxt *lexic)
  *
  * @nasluparam
  *
- * - An ssh session id.
+ * - An SSH session id.
  *
  * @naslret An integer representing the socket or -1 on error.
  *
@@ -715,7 +715,7 @@ leave:
  *
  * @nasluparam
  *
- * - An ssh session id.
+ * - An SSH session id.
  *
  * @naslnparam
  *
@@ -801,7 +801,7 @@ nasl_ssh_set_login (lex_ctxt *lexic)
  *
  * @nasluparam
  *
- * - An ssh session id.
+ * - An SSH session id.
  *
  * @naslnparam
  *
@@ -1000,7 +1000,7 @@ leave:
  *
  * @nasluparam
  *
- * - An ssh session id.
+ * - An SSH session id.
  *
  * @naslnparam
  *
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: nasl/nasl_ssh.c [hunk 1]
```diff
@@ -1896,3 +1897,63 @@ nasl_ssh_shell_close (lex_ctxt *lexic)
...
+ * NASL SFTP
+ */
+
+/**
+ * @brief Check if SFTP is enabled.
+ * @naslfn{sftp_enabled_check}
+ *
+ * @nasluparam
+ *
+ * - An ssh session id.
```
cfi-gb, line 1911:
```suggestion
 * - An SSH session id.
```

nasl/nasl_ssh.c
```diff
@@ -1896,3 +1897,63 @@ nasl_ssh_shell_close (lex_ctxt *lexic)
...
+ *
+ * @naslret An integer: 0 on success, -1 (SSH_ERROR) on Channel request
+ * subsystem failure. Greather than 0 means an error during SFTP init.
```
ArnoStiefvater, line 1914:
```suggestion
 * subsystem failure. Greater than 0 means an error during SFTP init.
```

nasl/nasl_ssh.c
```diff
@@ -1896,3 +1897,63 @@ nasl_ssh_shell_close (lex_ctxt *lexic)
 
   return NULL;
 }
+
+/*
+ * NASL SFTP
+ */
+
+/**
+ * @brief Check if SFTP is enabled.
```
cfi-gb, line 1906:
```suggestion
 * @brief Check if the SFTP subsystem is enabled on the remote SSH server.
```

nasl/nasl_ssh.c
```diff
@@ -1896,3 +1897,63 @@ nasl_ssh_shell_close (lex_ctxt *lexic)
...
+{
+  int tbl_slot, session_id;
+  tree_cell *retc;
+  sftp_session sftp;
+  ssh_session session;
+  int rc;
+
+  session_id = get_int_var_by_num (lexic, 0, -1);
+  if (!verify_session_id (session_id, "sftp_enabled_check", &tbl_slot, lexic))
+    return NULL;
```
ArnoStiefvater, line 1929: Is there a special distinction between errors via retc and NULL? If not we can just return retc with a new negative value. Else return NULL should at least be documented in the doxygen.

nasl/nasl_ssh.c
cfi-gb, line 1914: Seems the function can also return NULL which is missing here.

cfi-gb (changes requested):

nasl/nasl_ssh.c
```diff
@@ -1896,3 +1897,63 @@ nasl_ssh_shell_close (lex_ctxt *lexic)
...
+ *
+ * - An ssh session id.
+ *
+ * @naslret An integer: 0 on success, -1 (SSH_ERROR) on Channel request
+ * subsystem failure. Greather than 0 means an error during SFTP init.
+ *
+ * @param[in] lexic Lexical context of NASL interpreter.
+ */
+tree_cell *
+nasl_sftp_enabled_check (lex_ctxt *lexic)
```
ArnoStiefvater, line 1919: No sure about the naming.
Alternative would be `nasl_sftp_is_enabled` and then return 1 on success. 0 on subsystem failure because that basically means that it is not enabled. And negative values on other errors.

jjnicola: I prefer the name as it is, and return different values for different errors and gives a hint to feed devs. Errors can happen because wrong session id (nasl error NULL), error creating the the sftp(ssh error, negative) and initializing the sftp connection (sftp error positive).
Changing the name, just a boolean would be expected.

nasl/nasl_ssh.c
jjnicola: Added docstring. I prefer to differentiate the errors. In this case is a nasl error (e.g. bad session id, not present in session table)

ArnoStiefvater (approved):

cfi-gb (approved):

y0urself (approved):

PR #875, opened 2021-09-14
Add `find_all` to eregmatch() nasl function

jjnicola:
**What**:
Add `find_all` to eregmatch() nasl function

Jira: SC-395

**Why**:
The current implementation only returns the first match in the string.
Now it will return all matches if the option `find_all: TRUE` is passed to the function.
For backward compatibility, if FALSE is passed or no option, it will return the first match as before.

**How**:
Run the following nasl script with openvas-nasl without the option, with the option set FALSE/TRUE
```
foo = "11 abc
22 def
221 def
222 def
223 def
224 def
225 def
3333 ghi";
bar = eregmatch(string:foo, pattern:"[0-9]+", find_all: TRUE);
#display(foo);
display(bar);

```

**Checklist**:

- [ ] Tests

### DIFF
--- a/nasl/nasl_text_utils.c
+++ b/nasl/nasl_text_utils.c
@@ -786,8 +786,21 @@ nasl_egrep (lex_ctxt *lexic)
 
 /**
  * @brief Does extended regular expression pattern matching.
+ * @naslfn{eregmatch}
  *
- * In NASL, this function returns an array.
+ * @nasluparam
+ *
+ * - @a pattern An regex pattern
+ * - @a string A string
+ * - @a icase Boolean, for case sensitve
+ * - @a find_all Boolean, to find all matches
+ *
+ * @naslret An array with the first match (find_all: False)
+ *          or an array with all matches (find_all: TRUE).
+ *          NULL or empty if no match was found.
+ *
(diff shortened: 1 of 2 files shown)


nasl/nasl_text_utils.c
```diff
@@ -814,24 +828,63 @@ nasl_eregmatch (lex_ctxt *lexic)
...
+        }
+
+      int i;
+      for (i = 0; i < NS; i++)
+        if (subs[i].rm_so != -1)
+          {
+            v.var_type = VAR2_DATA;
+            v.v.v_str.s_siz = subs[i].rm_eo - subs[i].rm_so;
+            v.v.v_str.s_val = (unsigned char *) string + subs[i].rm_so;
+            (void) add_var_to_list (a, i, &v);
```
y0urself, line 849: Is this even make sense now? Why put that information in the `i`th place of `a`, if we only match once?
If `find_all` is true, it makes sense to use the `index` that will be incremented after every match, but this `i` is iterating through the `subs` not through the array, right?

jjnicola: This is for backward compatibility. So the behavior is the same as before, when the option is FALSE, or it is not passed.

y0urself (approved):

PR #942, opened 2021-11-10
Fix: Check inc files for Undefined function too

ArnoStiefvater:
**What**:

Check inc files for Undefined function too.
Jira: SC-174

**Why**:

Files were sometimes ignored previously.

**How**:

1. Create foo.inc with the following content:

```
bar();
function foo(){} 
```

2. Create bar.inc with the following content:

`function bar() {} `

3. Create foobar.nasl with the following content:

```
include("foo.inc");
include("bar.inc");
```

4. Run the following to see that errors are showing up now

```
$ openvas-nasl-lint foobar.nasl foo.inc bar.inc`
lib  nasl-Message: 15:11:19.036: [3361](foo.inc:1) Undefined function 'bar'
```

Previously no error was shown.

**Checklist**:

- [ ] Tests

### DIFF
--- a/nasl/lint.c
+++ b/nasl/lint.c
@@ -132,8 +132,7 @@ reverse_search (GSList **def_func_tree, GSList *finfo)
   GSList *finfo_aux;
 
   // The file name is the original file to be tested. It is not an include.
-  if (!g_strcmp0 (fdata->caller_file, nasl_name)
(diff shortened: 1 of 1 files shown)


ArnoStiefvater:
Closing PR as it appears that linter warnings for `.inc` files are not wanted/needed.

PR #949, opened 2021-11-15
Fix: stable.Dockerfile use gvm-libs as root

nichtsfrei:
The stable.Dockerfile was using the feed image as a root image which is
incorrect; this changes that to use gvm-libs instead.

**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/.docker/stable.Dockerfile
+++ b/.docker/stable.Dockerfile
@@ -8,6 +8,7 @@ RUN DESTDIR=/install cmake --build /build -- install
 
 FROM greenbone/community-feed-vts AS feed
 
+FROM greenbone/gvm-libs:$VERSION
 RUN apt-get update && apt-get install --no-install-recommends --no-install-suggests -y \
     bison \
     libjson-glib-1.0-0 \


.docker/stable.Dockerfile
```diff
@@ -8,6 +8,7 @@ RUN DESTDIR=/install cmake --build /build -- install
 
 FROM greenbone/community-feed-vts AS feed
```
bjoernricks, line 10: I guess you forgot to remove
```suggestion
```

nichtsfrei: no, that's there on purpose due to the copy later on

bjoernricks: oh sorry. looked weird to me in the context of your PR description :D

nichtsfrei: I could change it to:
```
COPY --from=greenbone/community-feed-vts /opt/greenbone/feed/plugins /var/lib/openvas/plugins
```
to get rid of the feed declaration but it looks clunky to me.

y0urself (approved):

PR #950, opened 2021-11-15
[oldstable] [CI] [docker] Refactor CI as in main

y0urself:
**What**:

* I will use `prod.Dockerfile` as a name (for all branches) for the production Docker image from now on. @nichtsfrei 

**Why**:

**How**:

**Checklist**:

- [ ] Tests


.docker/build.Dockerfile
```diff
@@ -0,0 +1,21 @@
+ARG VERSION=unstable
```
y0urself, line 1:
```suggestion
ARG VERSION=oldstable
```

.docker/build.Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+FROM greenbone/eulabeia-c-lib:$VERSION
```
y0urself, line 3:
```suggestion
FROM greenbone/gvm-libs:$VERSION
```

ArnoStiefvater (approved):

PR #958, opened 2021-11-19
Remove an invalid dependency for openvas-scanner package

shanmukh-awn:


y0urself:
You can change the base branch to 21.4, that would remove the other commits ...

PR #974, opened 2021-11-30
Fix warning detect by ccc-analyzer

jjnicola:
Fix warning detect by ccc-analyzer
`warning: Address of stack memory associated with local variable 'ifr' returned to caller`

Jira: SC-382
**What**:

**Why**:
to fix the warning! 

**How**:
compile with a newer gcc or use ccc-analyzer

**Checklist**:

- [ ] Tests

### DIFF
--- a/nasl/nasl_frame_forgery.c
+++ b/nasl/nasl_frame_forgery.c
@@ -377,22 +377,23 @@ nasl_dump_frame (lex_ctxt *lexic)
  * @brief Get the MAC address of host
  *
  * @param[in] ip_address    Local IP address
+ * @param[out] mac          The MAC address
  *
- * @return The MAC address of the host. NULL otherwise
+ * @return 0 on success. MAC address is put into buffer. -1 on error.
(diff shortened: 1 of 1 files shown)


nichtsfrei (dismissed):

nasl/nasl_frame_forgery.c
```diff
@@ -377,22 +377,22 @@ nasl_dump_frame (lex_ctxt *lexic)
  * @brief Get the MAC address of host
  *
  * @param[in] ip_address    Local IP address
+ * @param[out] mac    The mac address
  *
  * @return The MAC address of the host. NULL otherwise
```
ArnoStiefvater, line 382: Needs to be updated to something like
```suggestion
 * @return 0 on success. MAC address is put into buffer. -1 on error.
```

ArnoStiefvater (approved):
Tested and works as expected.

Kraemii (approved):

PR #995, opened 2021-12-08
Change: extend cert_query() nasl function to get the public key algorithm

jjnicola:
**What**:
Change: extend cert_query() nasl function to get the public key algorithm. 
Also, the old option "algorithm-name" is replaced with "signature-algorithm-name"

Jira: SC-473

**Why**:
The new cert information public-key-algorithm-name is required. It was confused with the signature algorithm.
 

**How**:
  `openvas-nasl -X -B -d -t google.com -i $VTDIR gb_tls_version_get.nasl gb_ssl_tls_cert_details.nasl --kb="TCP/PORTS=443"`
    
    And you should see now in the output both, the signature algorithm and the new the public key algorithm.

**Checklist**:

- [ ] Tests

### DIFF
--- a/nasl/nasl_cert.c
+++ b/nasl/nasl_cert.c
@@ -761,15 +761,21 @@ get_name (const char *string)
  *
  * - @a image       Return the entire certificate as binary data.
  *
- * - @a algorithm-name: Return the algorithm name. Get the OID of
- *                      the digest algorithm and translated to a name
- *                      from a list from Wireshark.
- *                      See epan/dissectors/packet-pkcs1.c
+ * - @a algorithm-name  Same as signature-algorithm-name. TODO: Remove it and
+ *                      leave only signature-algorithm-name.
  *
- * - @a modulus:     Return the RSA public key's modulus found in the
+ * - @a signature-algorithm-name  Return the algorithm name used to sign the
+ *                                certificate. Get the OID of the digest
+ *                                algorithm and translated to a name from a
+ *                                list from Wireshark.
+ *                                See epan/dissectors/packet-pkcs1.c
+ *
+ * - @a public-key-algorithm-name  Return the algorithm name of the public key.
+ *
+ * - @a modulus      Return the RSA public key's modulus found in the
  *                   structure of the given cert.
  *
- * - @a exponent:   Return the RSA public key's exponent found in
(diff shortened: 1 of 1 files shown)


nasl/nasl_cert.c
```diff
@@ -761,10 +761,15 @@ get_name (const char *string)
  *
  * - @a image       Return the entire certificate as binary data.
  *
- * - @a algorithm-name: Return the algorithm name. Get the OID of
- *                      the digest algorithm and translated to a name
- *                      from a list from Wireshark.
- *                      See epan/dissectors/packet-pkcs1.c
+ * - @a algorithm-name: Same as signature-algorithm-name. TODO: Remove it and
```
cfi-gb, line 764: Other parameters like `image` are not using the `:`, this one is using it. IMHO the style should stay consistent throughout the whole function description.

nasl/nasl_cert.c
```diff
@@ -761,10 +761,15 @@ get_name (const char *string)
...
+ *                      and leave only signature-algorithm-name.
+ *
+ * - @a signature-algorithm-name: Return the algorithm name. Get the OID of
```
cfi-gb, line 767:
```suggestion
 * - @a signature-algorithm-name: Return the algorithm name used to sign the certificate. Get the OID of
```

Not absolutely sure about the text but this should make it a little bit more clear.

cfi-gb (changes requested):

nasl/nasl_cert.c
```diff
@@ -930,8 +936,36 @@ nasl_cert_query (lex_ctxt *lexic)
...
+          return NULL;
+        }
+      algo_name = gnutls_pk_algorithm_get_name (algo) ? g_strdup (gnutls_pk_algorithm_get_name (algo)) : g_strdup ("unknown");
+      retc = alloc_typed_cell (CONST_DATA);
+      retc->size = strlen (algo_name);
+      retc->x.str_val = algo_name;
+    }
   else if (!strcmp (command, "modulus"))
     {
+
```
cfi-gb, line 968: This also looks superfluous and could be dropped as none of the other `else if` parts are using a space after the `{`.

ArnoStiefvater (approved):

Kraemii (approved):

PR #996, opened 2021-12-08
Handle string encoding conversion fail

ArnoStiefvater:
**What**:

Handle string encoding converison fail

**Why**:

Previously a failed conversion was not cought
which meant Null was supplied to `kb_item_push_str`
which resulted in a segfault.

`kb_item_push_str` will also be made more resilient
on gvm-libs side with a seperate PR.

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/misc/plugutils.c
+++ b/misc/plugutils.c
@@ -397,6 +397,7 @@ proto_post_wrapped (const char *oid, struct script_infos *desc, int port,
   const char *hostname = "";
   char *buffer, *data, port_s[16] = "general";
   char ip_str[INET6_ADDRSTRLEN];
+  GError *err = NULL;
   GString *action_str;
   gsize length;
   kb_t kb;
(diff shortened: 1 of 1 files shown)


misc/plugutils.c
```diff
@@ -424,7 +425,15 @@ proto_post_wrapped (const char *oid, struct script_infos *desc, int port,
                             msg_type_to_str (msg_type), ip_str, hostname ?: " ",
                             port_s, proto, oid, action_str->str, uri ?: "");
   /* Convert to UTF-8 before sending to Manager. */
```
bjoernricks, line 427: I suppose this comment isn't valid anymore because openvas doesn't send anything to the manager anymore?

ArnoStiefvater: At the end the result still end up in gvmd.
So instead of `openvas-scanner->redis->gvmd` its now `openvas-scanner->redis->ospd-openvas->gvmd`.
I think I would leave the comment as is, as hints at problems which might occur on manager side when the strings are not converted. But I don't deeply understand the issues with encoding problems. I could e.g. not get the `g_convert()` to fail on purpose.

jjnicola (approved):

Kraemii (approved):

PR #1046, opened 2022-02-24
openvas nasl incoherency

nichtsfrei:
Changed the behaviour of openvas-nasl-lint:
1. to lint all errors and don't abort eagerly
2. to only add functions of an include as used when they are called from an different caller than themselves.

Ways to verify:
Go to the vts git dir

```
git checkout ca2aebc4136c2f2ca82cea3223219e810ad9d015
openvas-nasl-lint -d gb_nominum_vantio_dns_detect.nasl
```

### DIFF
--- a/nasl/lint.c
+++ b/nasl/lint.c
@@ -181,7 +181,7 @@ static void
 check_called_files (gpointer key, gpointer value, GSList **unusedfiles)
 {
   if (key != NULL)
-    if (!g_strcmp0 (value, "NO"))
+    if (g_strcmp0 (value, "YES") != 0)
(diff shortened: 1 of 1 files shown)


nasl/lint.c
```diff
@@ -416,6 +415,28 @@ validate_function (lex_ctxt *lexic, tree_cell *st)
   return FAKE_CELL;
 }
 
+/**
+ * @brief Returns 1 if the function is at least sued once by another caller than
```
cfi-gb, line 419:
```suggestion
 * @brief Returns 1 if the function is at least used once by another caller than
```

ArnoStiefvater (approved):

jjnicola (approved):

PR #1068, opened 2022-03-21
Do not block on plug_get_host_fqdn and plug_get_key

nichtsfrei:
# changes plugin_load to not be forked

Since `openvas -u` ends itself after the child process is finished it does not make sense to fork it. Instead just execute it and end normally.

Testing:
```
openvas -u
```

# changes openvas-nasl to not fork on main

Since openvas-nasl does wait until the child process is finished to count the error the first fork is not necessary.

Testing:
```
openvas-nasl -X $path_to_a_scripts
```

# Changes `plug_get_host_fqdn` and `plug_get_key` to not wait directly after a fork for a spawned child process

Instead of immediately waiting for a child process to finish it is waiting before it ends the parent process. This allows parallel execution of scripts per vhost and get_kb.

Testing:

configure /etc/hosts to contain multiple hostnames /domains

```
127.0.1.1       pop-os.localdomain      pop-os
```

use nasl script
```
host = get_host_name();
sleep(2);
log_message(data:host);
```

and verify that it takes roughly two seconds:
```
time openvas-nasl -X -t pop-os test.nasl
```

### DIFF
--- a/nasl/nasl_func.c
+++ b/nasl/nasl_func.c
@@ -205,7 +205,7 @@ nasl_func_call (lex_ctxt *lexic, const nasl_func *f, tree_cell *arg_list)
       retc = pf2 (lexic2);
 
       // ignore when parent and plugin_util got forked
-      if (pu_is_parent ())
+      if (plug_is_parent ())
         {
           g_debug ("%s: setting return value to FAKE_CELL because it is parent "
                    "of forked child processes\n",
--- a/nasl/nasl_func.c
+++ b/nasl/nasl_func.c
@@ -18,6 +18,7 @@
 
 #include "nasl_func.h"
 
+#include "../misc/plugutils.h"
 #include "exec.h"
 #include "nasl_debug.h"
 #include "nasl_global_ctxt.h"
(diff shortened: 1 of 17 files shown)


nasl/nasl_func.c
```diff
@@ -202,6 +203,16 @@ nasl_func_call (lex_ctxt *lexic, const nasl_func *f, tree_cell *arg_list)
       tree_cell *(*pf2) (lex_ctxt *) = f->block;
 #pragma GCC diagnostic pop
       retc = pf2 (lexic2);
+
+      // ignore when parent and plugin_util got forked
+      if (pu_is_parent ())
+        {
+          g_debug ("%s: setting return value to FAKE_CELL because it is parent "
```
jjnicola, line 210: during tests, this is being log in the stdout. The log domain is missed in the file

nasl/nasl_packet_forgery.c
```diff
@@ -58,7 +58,9 @@
  * Checksum routine for Internet Protocol family headers (C Version)
  * From ping examples in W.Richard Stevens "UNIX NETWORK PROGRAMMING" book.
  */
-static int np_in_cksum (p, n) u_short *p;
+static int
```
jjnicola: This is the opportunity to write this in a modern way

nasl/nasl_packet_forgery_v6.c
```diff
@@ -81,7 +81,9 @@
  * Checksum routine for Internet Protocol family headers (C Version)
  * From ping examples in W.Richard Stevens "UNIX NETWORK PROGRAMMING" book.
  */
-static int np_in_cksum (p, n) u_short *p;
+static int
```
jjnicola: same as above

jjnicola (changes requested):

nichtsfrei:
Closing due to too unforeseeable in a running environment

PR #1070, opened 2022-03-24
reduce forks

nichtsfrei:
- Refactor: remove fork on plugin_load
- Refactor: remove fork within nasl#main

### DIFF
--- a/src/pluginload.c
+++ b/src/pluginload.c
@@ -251,8 +251,8 @@ cleanup_leftovers (int num_files)
   g_slist_free_full (oids, g_free);
 }
 
-static void
-plugins_reload_from_dir (void *folder)
+static int
+plugins_reload_from_dir (const char *folder)
 {
   GSList *files = NULL, *f;
   int loaded_files = 0, num_files = 0;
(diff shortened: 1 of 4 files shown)


src/pluginload.c
```diff
@@ -304,7 +304,7 @@ plugins_reload_from_dir (void *folder)
         g_message ("Loading %s", name);
       if (g_str_has_suffix (name, ".nasl"))
         {
-          if (nasl_plugin_add (folder, name))
+          if (nasl_plugin_add ((char *) folder, name))
```
jjnicola, line 307: Why is the const qualifier being removed? wouldn't be better to make const also the parameter in the nasl_plugin_add() (in src/nasl_plugins.c) ?

jjnicola (changes requested):

src/pluginload.c
nichtsfrei: yes.

jjnicola (approved):

ArnoStiefvater (approved):

PR #1075, opened 2022-03-31
Fix: duplicate results

jjnicola:
**What**:
Compares the scan id in globals, set at the beginning
of the scan, with the one found in the main kb.
It helps to detect that the kb was not taken by another
task/scan, and that the current plugins does not stores
results in a wrong kb.

Jira: SC-378

**Why**:
A scan can left result in a kb, which is later taken for a new scan. This produces that results from an interrupted scan are shown in the report of a new scan. With this patch, the scanner checks the kb before leaving a new result.

**How**:
Run a scan. The race condition is hard to reproduce.

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/attack.c
+++ b/src/attack.c
@@ -53,7 +53,7 @@
 #include <gvm/util/mqtt.h>
 #include <gvm/util/nvticache.h> /* for nvticache_t */
 #include <pthread.h>
-#include <stdlib.h>   /* for exit() */
+#include <signal.h>
 #include <string.h>   /* for strlen() */
 #include <sys/wait.h> /* for waitpid() */
 #include <unistd.h>   /* for close() */
@@ -426,6 +426,7 @@ run_table_driven_lsc (const char *scan_id, kb_t kb, const char *ip_str,
       return -1;
     }
   /* Get the OS release. TODO: have a list with supported OS. */
+
   os_release = kb_item_get_str (kb, "ssh/login/release_notus");
   /* Get the package list. Currently only rpm support */
   package_list = kb_item_get_str (kb, "ssh/login/package_list_notus");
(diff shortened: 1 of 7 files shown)


nichtsfrei (dismissed):

ArnoStiefvater (changes requested):
Blocking accidentally merge for now. Should be merged after new major release is out.

src/attack.c
```diff
@@ -655,8 +664,22 @@ attack_host (struct scan_globals *globals, struct in6_addr *ip, GSList *vhosts,
...
+        {
+          // As long as we don't have a proper communication channel
+          // to our ancestors we just kill our parent and ourselves
+          // (but let our grandparents live).
+          // To prevent duplicate results we don't let ACT_END run.
+          me = getpid ();
+          kill (parent, SIGQUIT);
+          // just in case
+          kill (me, SIGQUIT);
+          return;
```
Kraemii, line 677: A few things about this part
1. The signal SIGQUIT will be ignored by the main openvas-process, which should be the parent here. Maybe a SIGUSR1 can be send instead, this would signal that the scan should be stopped.
2. Is there a reason why using SIGQUIT instead of SIGTERM?
3. To send a signal to the own process the function `raise` can be used

Kraemii:
I found several `kb_item_push_str` where the check is missing:
- https://github.com/jjnicola/openvas-scanner/blob/kb-inconsitency/misc/network.c#L1994
- https://github.com/jjnicola/openvas-scanner/blob/kb-inconsitency/nasl/nasl_host.c#L138
- https://github.com/jjnicola/openvas-scanner/blob/kb-inconsitency/nasl/nasl_host.c#L140
- https://github.com/jjnicola/openvas-scanner/blob/kb-inconsitency/src/attack.c#L185
- https://github.com/jjnicola/openvas-scanner/blob/kb-inconsitency/src/attack.c#L219
- https://github.com/jjnicola/openvas-scanner/blob/kb-inconsitency/src/attack.c#L706
- https://github.com/jjnicola/openvas-scanner/blob/kb-inconsitency/src/attack.c#L771
- https://github.com/jjnicola/openvas-scanner/blob/kb-inconsitency/src/openvas.c#L400
Not sure if it is necessary in all cases, but wouldn't it be better if the consistency is automatically checked when trying to 
push things into it? So we do not have to check this on our own every time?

nichtsfrei:
replaced by:https://github.com/greenbone/openvas-scanner/pull/1159

PR #1080, opened 2022-04-11
Add: possibility to fallback to LEGACY:%COMPAT:%UNSAFE_RENEGOTIATION

nichtsfrei:
When a TLS connection with `NORMAL:+ARCFOUR-128:%COMPAT` as well as with
a INSECURE_DH_PRIME_BITS fails then find-service will try a LEGACY
connection before falling back to an unencrypted connection.

### DIFF
--- a/nasl/nasl_builtin_find_service.c
+++ b/nasl/nasl_builtin_find_service.c
@@ -50,6 +50,10 @@
 
 #define NUM_CHILDREN "Number of connections done in parallel : "
 
+// we cannot use the GNU ones due to number mismatch
+#define TLS_PRIME_UNACCEPTABLE -2
+#define TLS_FATAL_ALERT -3
+
 #undef G_LOG_DOMAIN
 /**
  * @brief GLib logging domain.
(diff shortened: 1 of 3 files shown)


nasl/nasl_builtin_find_service.c
```diff
@@ -1488,6 +1488,56 @@ may_be_time (time_t *rtime)
...
+  // failed to create a TLS connection
+  if (cnx == -2 && test_ssl)
+    {
+      // retry with insecure flag
+      cnx = open_stream_connection_ext (desc, port, trp, timeout, p, inf);
+      if (cnx == -2)
+        {
+          g_debug ("%s: %s failed, retrying with %s", __func__, p, lp);
+          // try with legacy connection string
+          cnx = open_stream_connection_ext (desc, port, trp, timeout, lp, inf);
```
cfi-gb, line 1517: AFAICT this code is never reached if the server is not responding with `GNUTLS_E_DH_PRIME_UNACCEPTABLE` (`-2` status code) which might happen when the server is e.g. not supporting any Diffie-Hellman cipher suites.

Looks like we need a different nested if/else code like e.g.:

```
cnx = open_stream_connection (desc, port, trp, timeout);

# nb: Catching GNUTLS_E_DH_PRIME_UNACCEPTABLE (-2) and GNUTLS_E_FATAL_ALERT_RECEIVED (-3)
if (cnx <= -2 && test_ssl)
  if (cnx == -2)
    // retry with insecure flag
    if (cnx >= 0)
      // Set INSECURE_DH_PRIME_BITS as default
  if (cnx == -3)
    // try with legacy connection string
    if (cnx >= 0)
      // Set LEGACY prio string as default
```

cfi-gb (changes requested):

nasl/nasl_builtin_find_service.c
nichtsfrei: Right, thanks; I adapted it.

cfi-gb: Thanks a lot for the follow-up changes. :+1: 

LGTM logic wise now but can't judge on the code itself so just marking this as solved and awaiting a review from the scanner team.

cfi-gb: Mhhh, i don't have any permissions to resolve this conversation so please feel free to resolve this on your own.

Kraemii (approved):

PR #1091, opened 2022-04-28
Refactor: replace gvm-libs proctitle-set with libbsd

nichtsfrei:
Instead of implementing our own version of proctitle manipulation we
should use existing libraries providing this functionality.

This commit switches to libbsd setproctitle.


Kraemii:
The dependency to libbsd should be added to the [INSTALL.md](https://github.com/greenbone/openvas-scanner/blob/main/INSTALL.md)

nichtsfrei:
> The dependency to libbsd should be added to the [INSTALL.md](https://github.com/greenbone/openvas-scanner/blob/main/INSTALL.md)

Right, done :)

Kraemii (approved):

PR #1124, opened 2022-06-07
Remove: script_bugtraq_id completely

Kraemii:
**What**:
As it is not needed anymore and all VTs are adjusted, we are now able to drop the support for this function completely
SC-195

**Why**:

**How**:

**Checklist**:

- [ ] Tests


nasl/nasl_scanner_glue.c
```diff
@@ -172,17 +146,11 @@ script_bugtraq_id_dummy ()
  * list of comma-separated values.
  *
  * In fact, if name is "cve" or "bid", it is equivalent
```
cfi-gb, line 148: Seems this comment is still referring to a `bid`.

cfi-gb (changes requested):

nichtsfrei (approved):

PR #1125, opened 2022-06-09
Add: smoketest for openvas-nasl-lint

Kraemii:
**What**:
A smoke-test to verify the functionality of openvas-nasl-lint

**Why**:

**How**:
In the current state the smoketest can only be triggered manually, but a github-actions implementation is already ongoing.
To start the smoketest just go into the `smoketest_lint` folder and run `make all`

**Checklist**:

- [ ] Tests

### DIFF
--- a/smoketest_lint/cmd/main.go
+++ b/smoketest_lint/cmd/main.go
@@ -1,30 +1,37 @@
 package main
 
 import (
+	"flag"
 	"fmt"
+	"os"
 	"path/filepath"
 
 	"github.com/greenbone/openvas-scanner/smoketest_lint/test"
 )
 
-var path = "data"
-
 func main() {
 
+	openvasExe := flag.String("e", "openvas-nasl-lint", "openvas-nasl-lint executable, must be either in $PATH, an absolute path or relative to the data folder")
+	path := flag.String("d", "data", "data folder for the nasl test data")
+	flag.Parse()
+
 	tfs := test.TestFiles()
-	err := filepath.Walk(path, tfs.Parse)
+	fmt.Println("Parsing Nasl Test Files")
+	err := filepath.Walk(*path, tfs.Parse)
 	if err != nil {
-		fmt.Printf("Unable to parse files in %s: %s\n", path, err)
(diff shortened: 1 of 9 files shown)


smoketest_lint/cmd/main.go
```diff
@@ -0,0 +1,36 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/greenbone/openvas-scanner/smoketest_lint/test"
+)
+
+func main() {
+
+	openvasDir := flag.String("openvasDir", "", "location of the openvas-nasl-lint executable, has to be absolute or relative to test files directory")
```
nichtsfrei, line 14: Instead of just adding the root dir of a executable I would rather use executable or e and set the default to openvas-nasl-lint?

nichtsfrei: Also a little bit of output when running would be nice.

nichtsfrei (changes requested):

nichtsfrei (changes requested):

smoketest_lint/test/testfile.go
```diff
@@ -0,0 +1,145 @@
...
+// as a list of errors
+func (t testFile) Test(openvasExe string) []error {
+	errs := make([]error, 0)
+	// run openvas-nasl-lint and collect output
+	cmd := exec.Command(openvasExe, t.Name)
+	cmd.Dir = t.dir
+	// Ignoring error as openvas-nasl-lint does not exit with 0 when errors
+	// were found
+	out, err := cmd.CombinedOutput()
+	if !strings.Contains(err.Error(), "exit status") {
```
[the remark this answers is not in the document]
Kraemii: The comment is a bit irritating as I forgot to adjust it. I wantet to filter for Errors which are not ExitErrors, as these will occur when checking files which contains openvas-nasl-lint errors. Now it should be much more clear, also used `errors.As`, as `errors.Is` checks for equality of two errors.

nichtsfrei (approved):

PR #1134, opened 2022-06-30
Revert "Testing if arm is responsible for grey synk symbol on docker hub"

y0urself:
Reverts greenbone/openvas-scanner#1131


y0urself:
This was not the problem.

PR #1137, opened 2022-07-08
Fix: result len of aes{128,256}_gcm_{encrypt,decrypt}_auth

nichtsfrei:
The result len when using
- aes_128_gcm_encrypt_auth
- aes_128_gcm_decrypt_auth
was incorrect. It was computing it by:
`((datalen / 16) + 1) * 16`
although it should be datalen.

To verify it there a is a new test in:
- nasl/tests/test_crypt_data_aes.nasl

And to verify it the execution Makefile is simplified and added to
github workflows.

To manually verify it go to `nasl/tests` and execute `make check`.


.github/workflows/nasl.yml
```diff
@@ -0,0 +1,28 @@
...
+      - uses: actions/checkout@v3
+      - name: build
+        run: |
+          cmake -Bbuild -DCMAKE_BUILD_TYPE=Release
+          cmake --build build
+      - name: redis
+        run: |
+          apt-get update && apt-get install --no-install-recommends --no-install-suggests -y redis
+          mkdir /run/redis
+          redis-server .github/redis.conf || exit 1
```
ArnoStiefvater, line 24: We already have a redis config in the config folder.

config/redis-openvas.conf
```diff
@@ -106,7 +106,7 @@ tcp-backlog 511
 # incoming connections. There is no default, so Redis will not listen
 # on a unix socket when not specified.
 #
-unixsocket /run/redis-openvas/redis.sock
+unixsocket /run/redis/redis.sock
```
ArnoStiefvater, line 109: When we change the socket location the install documentation does not match anymore: https://greenbone.github.io/docs/latest/21.04/source-build/index.html#setting-up-the-redis-data-store

nichtsfrei: That's 21.04 there it is still accurate?

nichtsfrei: Anyway changed it back and adapted workflow

ArnoStiefvater (approved):

PR #1139, opened 2022-07-14
Fix: openvas-nasl should not exit when forked

nichtsfrei:
When executing openvas-nasl and calling a function that forks the
process then it will happen that the plugin execution is stopped
and other plugins won't be executed afterwards.

To verify it create two nasl plugins:

```
host = get_host_name();
display(strcat("starting ", host));
exit(0);
```
hostname.nasl

```
display("you should see me");
exit( 0);
```
log.nasl

and execute them in the order hostname.nasl log.nasl with openvas-nasl.

`openvas-nasl -X hostname.nasl log.nasl`

jira: SC-617

### DIFF
--- a/nasl/exec.c
+++ b/nasl/exec.c
@@ -1624,7 +1624,7 @@ exec_nasl_script (struct script_infos *script_infos, int mode)
 {
   naslctxt ctx;
   nasl_func *pf;
-  int err = 0, to, process_id;
+  int err = 0, to;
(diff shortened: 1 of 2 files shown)


ArnoStiefvater (dismissed):
Blocking accidental merge for now. Should be merged after new major release is out.

Kraemii (approved):

PR #1142, opened 2022-07-26
Fix process handling

Kraemii:
**What**:
This PR introduces a new abstraction for process and signal handling.

Process handling:
Adds a process management layer to keep track of child processes. Also automates some initialization handling for the spawned childs and setup default signal handling.

Signal handling:
Adds the possibility to easily overwrites default signal handling and adds some additional features for it. Signal handlers can be chained in case more than one thing has to be handled in case of a signal. Also adds the possibility to control the flow of the signal handling chain. In addition data can be provided to the handler.

The reason for this PR is the almost opaque and in many cases not working process and signal handling. These problems resulted in dangling processes, which never got cleaned and signal which got ignored or not handled correctly.

SC-626

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/nasl/nasl_builtin_find_service.c
+++ b/nasl/nasl_builtin_find_service.c
@@ -2415,32 +2415,6 @@ plugin_do_run (struct script_infos *desc, GSList *h, int test_ssl)
 
 static pid_t sons[MAX_SONS];
 
-static void
-sigterm (int s)
-{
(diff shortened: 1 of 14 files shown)


Kraemii:
Closed in favor of IPC via pipes

PR #1143, opened 2022-08-03
disable dump and memory writes to disk

Dexus:
**What**:

same fix for DEVOPS-316

**Why**:

**How**:

**Checklist**:

- [ ] Tests


bjoernricks (approved):

ArnoStiefvater (commented):
I don't see the use of this PR.

1. `appendonly no` is already set at line 699
2. We don't have any save points in the config. So no need to delete them with `save ""`. The save config is explained in line 198.

Same for https://github.com/greenbone/greenbone-docker/pull/25

PR #1152, opened 2022-08-15
Refactor: tidy up exit calls

Kraemii:
**What**:
In case a child process exits, it should exit with _exit. Calling exit instead it will execute atexit and on_exit calls. This can lead to errors in other processes as it will e.g. remove files and streams set in stdio.h.
SC-564

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/nasl/nasl.c
+++ b/nasl/nasl.c
@@ -61,12 +61,6 @@ extern char *
 nasl_version (void);
 
 static void
-sighandler ()
-{
-  exit (0);
-}
-
-static void
 my_gnutls_log_func (int level, const char *text)
 {
   fprintf (stderr, "[%d] (%d) %s", getpid (), level, text);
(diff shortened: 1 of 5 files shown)


nichtsfrei (approved):

nasl/nasl.c
```diff
@@ -289,8 +283,6 @@ main (int argc, char **argv)
       fprintf (stderr, "** WARNING : packet forgery will not work\n");
       fprintf (stderr, "** as NASL is not running as root\n");
     }
-  signal (SIGINT, sighandler);
-  signal (SIGTERM, sighandler);
```
jjnicola, line 293: I think that the intention here was to finish the program in a normal way (with exit(0)) even when the user sends a kill or interrupts with `ctrl + c`. 
openvas-nasl sometimes forks. I an not sure if it would be the same for child processes to kill or interrupt the parent process than an exit(). For sure we are changing the exit code from 0 to the respectively 2 and 15.
Do we want to change this behavior? TBH I don't know what to do here . Feel free to change it if you want.

jjnicola (changes requested):

jjnicola (approved):

PR #1153, opened 2022-08-15
Change: Update nasl lib and misc lib version from 21.4.1 to 21.4.5

jjnicola:
**What**:
Change: Update nasl lib and misc lib version from 21.4.1 to 21.4.5

Jira: SC-654

**Why**:
It was not updated before and after releasing.

**How**:
Update, install and check new installed versions

**Checklist**:

- [ ] Tests


jjnicola:
No will be done for old stable

PR #1156, opened 2022-08-16
Remove: unnecessary code in libs CMakeLists.txt files

jjnicola:
**What**:
Remove the project string setup code for the nasl and misc CMakeLists.txt files
Jira: SC-654

**Why**:
The main project version string is already used.

**How**:
Run cmake and compile (with and without release flag, with and without dev version var set). Check the installed binary's names

**Checklist**:

- [ ] Tests


nasl/CMakeLists.txt
```diff
@@ -21,26 +21,6 @@
 ## Project version
 # The following three variables should be set through the project command once
 # we require CMake >= 3.0
```
Kraemii, line 23:
```suggestion
```

ArnoStiefvater (approved):

PR #1159, opened 2022-08-18
kb inconsitency

nichtsfrei:
Add: functions to check for kb inconsistences

Compares the scan id in globals, set at the beginning
of the scan, with the one found in the main kb.
It helps to detect that the kb was not taken by another
task/scan, and that the current plugins does not stores
results in a wrong kb.

Additionally it will kill the running process as well as the parent, when
the inconsistency is detected at starting a host scan.

To avoid using static char * within openvas.c and make it more
convenient to use scan_id handling is handled within scan_id.c by the
methods:

- set_scan_id - to set the scan_id
- get_scan_id - to get it

If the scan_id is not yet set then get_scan_id returns NULL.

To make it easier to verify the kb while trying to write there the
functions:

- kb_check_push_str (kb_t, const char *, const char *, size_t);
- kb_check_set_str (kb_t, const char *, const char *, size_t);
- kb_check_add_str_unique (kb_t, const char *, const char *, size_t, int);
- kb_check_set_int (kb_t, const char *, int);
- kb_check_add_int (kb_t, const char *, int);
- kb_check_add_int_unique (kb_t, const char *, int);

The call `check_kb_inconsistency` and if there is a global scan_id and
the db contains a scan_id then it verifies if they're the same.

If not those methods will not call the underlying `kb_item_` methods.

If there is no no scan_id in the kb it will print a debug message that
this call is maybe misplaced.
But it will store it within the given kb in those cases. This is done
intentionally to may figure cases in that `kb_item_push_str` is more
suitable.

Co-authored-by: Juan Jose Nicola <juan.nicola@greenbone.net>

### DIFF
--- a/src/attack.c
+++ b/src/attack.c
@@ -53,7 +53,7 @@
 #include <gvm/util/mqtt.h>
 #include <gvm/util/nvticache.h> /* for nvticache_t */
 #include <pthread.h>
-#include <stdlib.h>   /* for exit() */
+#include <signal.h>
 #include <string.h>   /* for strlen() */
 #include <sys/wait.h> /* for waitpid() */
 #include <unistd.h>   /* for close() */
@@ -426,6 +426,7 @@ run_table_driven_lsc (const char *scan_id, kb_t kb, const char *ip_str,
       return -1;
     }
   /* Get the OS release. TODO: have a list with supported OS. */
+
   os_release = kb_item_get_str (kb, "ssh/login/release_notus");
   /* Get the package list. Currently only rpm support */
   package_list = kb_item_get_str (kb, "ssh/login/package_list_notus");
@@ -637,6 +638,8 @@ attack_host (struct scan_globals *globals, struct in6_addr *ip, GSList *vhosts,
   /* Used for the status */
   int num_plugs, forks_retry = 0, all_plugs_launched = 0;
   char ip_str[INET6_ADDRSTRLEN];
+  struct scheduler_plugin *plugin;
+  pid_t parent, me;
 
   addr6_to_str (ip, ip_str);
   openvas_signal (SIGUSR2, set_check_new_vhosts_flag);
@@ -653,9 +656,6 @@ attack_host (struct scan_globals *globals, struct in6_addr *ip, GSList *vhosts,
   num_plugs = plugins_scheduler_count_active (sched);
   for (;;)
     {
-      struct scheduler_plugin *plugin;
-      pid_t parent;
-
       /* Check that our father is still alive */
       parent = getppid ();
       if (parent <= 1 || process_alive (parent) == 0)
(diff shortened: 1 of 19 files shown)


src/attack.c
```diff
@@ -655,8 +662,21 @@ attack_host (struct scan_globals *globals, struct in6_addr *ip, GSList *vhosts,
...
+      if (check_kb_inconsistency (main_kb) != 0)
+        {
+          // As long as we don't have a proper communication channel
+          // to our ancestors we just kill our parent and ourselves
+          // (but let our grandparents live).
+          // To prevent duplicate results we don't let ACT_END run.
+          kill (parent, SIGTERM);
+          // just in case
+          raise (SIGTERM);
+          return;
```
Kraemii, line 674:
```suggestion
          kill (parent, SIGUSR1);
          raise (SIGUSR1);
```
Killing the parent will be just ignored and killing itself disables the clean up. Currently running scripts won't get stopped. Instead sending a `SIGUSR1` to the parent and itself will work as a stop scan signal, which should work in this case, as the scheduler will call `scan_is_stopped()` directly after this block.

Kraemii:
I also get a strange behavior, if I change the scan ID directly in redis before openvas even started, as the ID is set by ospd-openvas. In this case the openvas process will stop execution pretty fast, but will not be registered as stopped within ospd-openvas. Instead the scan process in ospd-openvas gets kinda stucked and it is not possible to stop it normally. Not sure how the impact of this case is, or if this even can happen.

jjnicola (approved):

src/attack.c
jjnicola: An alternative to start the stop scan chain, would be to use the parent pid as group pid. Like in openvas.c, you can call directly 
```suggestion
          killpg (parent, SIGUSR1);
```

jjnicola:
For testing:

- Start a F&F scan against a single host.
- kill -9 the ospd-openvas task process (the parent of the openvas scan ). Openvas still runs but its parent is the init process (pid 1)
- fake the release of kbs in redis. If you started with a clean db, it should be db 2 and 3
 ```
redis /run/redis-openvas/redis.sock> info keyspace
# Keyspace
db0:keys=1,expires=0,avg_ttl=0
db1:keys=195852,expires=0,avg_ttl=0
db2:keys=5,expires=0,avg_ttl=0
db3:keys=7702,expires=0,avg_ttl=0
redis /run/redis-openvas/redis.sock> HGETALL GVM.__GlobalDBIndex
1) "1"
2) "1"
3) "2"
4) "1"
5) "3"
6) "1"
redis /run/redis-openvas/redis.sock> HDEL GVM.__GlobalDBIndex 2 3
(integer) 2
```
- Restart the scan. Do **not** resume, to ensure a new scan id

jjnicola:
with the last commit (killpg (parent, SIGUSR1)), the host process SIGSEGV because the kb was flush'ed and the host process can't find the key with openvas pid.

jjnicola (approved):
With the last changes, my the tests were successful.

PR #1165, opened 2022-08-19
Add: fork observing mechanism

Kraemii:
**What**:
This PR extends the process module and allows to manage forked childs as well sets a limit to the number of forks possible. It also cleans up child processes.

**Why**:
In order to use IPC some sort of fork observing mechanism is necessary

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/processes.c
+++ b/src/processes.c
@@ -46,39 +46,165 @@
  */
 #define G_LOG_DOMAIN "sd   main"
 
+typedef struct proc_entry
+{
+  pid_t pid;
+  int terminated;
+} proc_entry;
+
+static proc_entry *procs = NULL;
+static int initialized = 0;
+static int max_procs;
+static int num_procs;
+
 /**
- * @brief Send SIGTERM to the pid process. Try to wait the
- * the process. In case of the process has still not change
-   the state, it sends SIGKILL to the process and must be waited
-   later to avoid leaving a zombie process
+ * @brief Function for handling SIGCHLD to clear procs
+ *
+ */
+static void
+clear_child ()
+{
+  if (!initialized)
+    return;
 
-   @param[in] pid Process id to terminate.
+  pid_t pid;
+  // In case we receive multiple SIGCHLD at once
+  while ((pid = waitpid (-1, NULL, WNOHANG)) > 0)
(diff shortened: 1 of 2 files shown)


src/processes.c
```diff
@@ -46,76 +46,224 @@
...
 
+typedef struct proc_entry
+{
+  pid_t pid;
+  int terminated;
+} proc_entry;
+
+static proc_entry *procs = NULL;
+static int initialized = 0;
+static int last;
```
nichtsfrei, line 57: Considering that it is a relatively small array it may be better to find the last index when we need it instead of storing it in a global variable?

src/processes.h
```diff
@@ -28,9 +28,25 @@
...
 
+#define FORKFAILED -1
+#define NOINIT -2
+#define PROCSFULL -3
+#define NOCHILD -4
+
 typedef void (*process_func_t) (void *);
+
+void
+init_procs (int max);
```
nichtsfrei, line 39: maybe it makes sense to apply a prefix to the methods defined here? Like procs_terminate_child, procs_init, ... to make it easier to find the implementation of those methods when looking for them.

nichtsfrei (changes requested):

src/processes.h
Kraemii: I did it for `terminate_childs` and `procs_init` as they currently are not used in the rest of openvas. Feel free to change their names too, when applying IPC

src/processes.c
Kraemii: Changed it

nichtsfrei (approved):

src/processes.c
```diff
@@ -46,76 +46,220 @@
...
+ * @return pid of spawned process on success or one of the following errors:
+ * FORKFAILED, NOINIT, PROCSFULL
  */
 pid_t
-create_process (process_func_t function, void *argument)
+create_process (process_func_t func, void *args)
 {
-  int pid;
-
-  gvm_log_lock ();
```
jjnicola, line 104: Why did you remove this? This code was introduced some time ago to avoid deadlock. See PR #491

PR #1176, opened 2022-08-24
Fix: determine SSL/TLS support on services

jjnicola:
**What**:
Scanner failed to determine SSL/TLS support on services requiring a client certificate.
Jira: SC-659

**Why**:
Now the variable trp is passed to the function to be set.

**How**:
`sudo openvas-nasl -X -B -d -i ./ -t zero.webappsecurity.com find_service.nasl --kb="Ports/tcp/443=1"`

without the patch, I have seen like the transport key stores a random integer (trp not set/initiliazed)
`lib  misc-Message: 13:59:08.804: set key Transports/TCP/443 -> 109136`

with the patch, the transport key is set with the right value:
`lib  misc-Message: 14:02:51.355: set key Transports/TCP/443 -> 9`

**Checklist**:

- [ ] Tests

### DIFF
--- a/nasl/nasl_builtin_find_service.c
+++ b/nasl/nasl_builtin_find_service.c
@@ -1494,18 +1494,18 @@ may_be_time (time_t *rtime)
 
 static int
 retry_stream_connection (int test_ssl, struct script_infos *desc, int port,
-                         int timeout)
+                         int timeout, int *trp)
 {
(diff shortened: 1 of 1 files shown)


Kraemii (approved):

cfi-gb:
> support on services requiring a client certificate.

Note that this seems to also affect other kind of SSL/TLS services, not only the ones requiring a client certificate

PR #1180, opened 2022-08-25
Refactor: uniform include guards

### LINKED ISSUE #446: reserved identifier violation
I would like to point out that an identifier like “[`_OPENVAS_H`](https://github.com/greenbone/openvas/blob/d12adfde604fbfcb5cafacd31cb7a1d79069f857/src/openvas.h#L25 "Update candidate")” [does not fit](https://www.securecoding.cert.org/confluence/display/c/DCL37-C.+Do+not+declare+or+define+a+reserved+identifier#DCL37-C.Donotdeclareordefineareservedidentifier-NoncompliantCodeExample%28IncludeGuard%29 "Do not use identifiers which are reserved for the compiler implementation.") to the expected naming convention of the C language standard.
Would you like to adjust your selection for unique names?

Kraemii:
There were many violation of compliant include guards across the project. This PR fixes that. Additionally a uniform include guard format is introduced with the following scheme: DIR_FILE_H
There are two exceptions:
1. If the name of the file is the same as the folder the naming scheme is FILE_H
2. For the src folder the prefix OPENVAS is used instead of SRC

fixes #446 
SC-658

If you have any opinions on the naming scheme, please feel free to comment

### DIFF
--- a/misc/bpf_share.h
+++ b/misc/bpf_share.h
@@ -23,8 +23,8 @@
  * @brief Header file for module bpf_share.
  */
 
-#ifndef OPENVAS_BPF_SHARE_H
-#define OPENVAS_BPF_SHARE_H
+#ifndef MISC_BPF_SHARE_H
+#define MISC_BPF_SHARE_H
 
 #include <sys/types.h>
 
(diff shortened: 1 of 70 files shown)


nichtsfrei (approved):

elfring:
Thanks for your adjustment of affected include guards.

PR #1192, opened 2022-09-12
[main] Handle User-Agent sent to the parent process via IPC SC-655

jjnicola:
**What**:
Via IPC, the first time that the User-Agent is built, can be sent to the parent process and stored. Later, the plugin processes inherits it and uses it.

Jira: SC-655

**Why**:
Currently, the user agent must be taken from a script preference, and in case this is not set, it will get from the vendor. As fall back it uses the nasl library version. This will be done thousand times during a F&F scan, since it is done each time a nasl script calls an nasl_http_* function.

**How**:
Run a F&F scan. Set vendor version in the openvas.conf or a User-Agent in the plugin preferences of global_settings.nasl.

**Checklist**:

- [ ] Tests

### DIFF
--- a/misc/CMakeLists.txt
+++ b/misc/CMakeLists.txt
@@ -76,6 +76,7 @@ set (CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -Werror")
 
 
 ## Pass-throughs
+add_definitions (-DOPENVAS_MISC_VERSION="${PROJECT_VERSION_STRING}")
 
 include_directories (${GLIB_INCLUDE_DIRS} ${GLIB_JSON_INCLUDE_DIRS}
                      ${LIBGVM_BASE_INCLUDE_DIRS}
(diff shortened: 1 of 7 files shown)


nichtsfrei (approved):

cfi-gb:
@jjnicola @nichtsfrei It seems the commit message and the PR title is referencing SC-655 while it should reference SC-665 so this PR probably needs some updates for the PR title and the description (the commit itself can't be changed as it was already merged) to avoid confusion in the future.

As the #1202 backport contains the same typo i would suggest to reword the commit message before merging so that the correct issue number is included in the stable branch.

ArnoStiefvater:
> @jjnicola @nichtsfrei It seems the commit message and the PR title is referencing SC-655 while it should reference SC-665 so this PR probably needs some updates for the PR title and the description (the commit itself can't be changed as it was already merged) to avoid confusion in the future.
> 
> As the #1202 backport contains the same typo i would suggest to reword the commit message before merging so that the correct issue number is included in the stable branch.

Thanks! fixed it.

PR #1203, opened 2022-09-27
Improvement: send the lsc data when it is gathered.

jjnicola:
**What**:
Currently, the scanner sends the package list (if exists) via MQTT to Notus, at the end of the plugin launch.
Notus will get the package list once the plugin scheduler finished. Even if the package list was already stored earlier.  

This adds a new nasl function called _**update_table_driven_lsc_data()**_ . It must be called from the nasl script, once the OS release and the package list keys are stored in the host kb. So, the nasl script gather-package-list.nasl must be modified.

Jira: SC-681

**Why**:
Now with IPC, the plugin can communicate the parent process that the package list was already gathered, and the parent process sends it to the notus-scanner.

**How**:
Run a scan including LSC. Compare scan progress and the amount of results. With the patch, you should see the notus results earlier in scan progress, not at the end of the scan.

**Checklist**:

- [ ] Tests


jjnicola:
This PR must be merged once the nasl script **_gather-package-list.nasl_** is updated with something like  
```
# Notify the scanner there is data for running a LSC
   if( defined_func( "update_table_driven_lsc_data" ) && pkg_list_notus && oskey_notus)
     update_table_driven_lsc_data( pkg_list: pkg_list_notus, os_release: oskey_notus );
```

nasl/nasl_scanner_glue.c
```diff
@@ -1077,6 +1078,36 @@ nasl_vendor_version (lex_ctxt *lexic)
...
+ *        for use in the host kb
+ *
+ * @naslfn{update_table_driven_lsc_data}
+ *
+ * @param[in] lexic  Lexical context of the NASL interpreter.
+ *
+ * @return NULL
+ */
+tree_cell *
+nasl_update_table_driven_lsc_data (lex_ctxt *lexic)
```
nichtsfrei, line 1092: In the example you provide:
```
update_table_driven_lsc_data( pkg_list: pkg_list_notus, os_release: oskey_notus );
```

but in this function you don't gather this data nor do you handle it; meaning it does still require a

```
set_kb_item( name: "ssh/login/package_list_notus", value: value);
set_kb_item( name: "ssh/login/release_notus", value: value);
```
on the nasl side making it unnecessarily complex to use; I think it would be much easier to maintain and use later on when this specific key handling is done within this method instead of enforcing the user to know internals.

jjnicola: I fixed the example. The function `update_table_driven_lsc_data()` doesn't receive any parameter.

src/attack.c
```diff
@@ -319,183 +319,108 @@ append_vhost (const char *vhost, const char *source)
...
+
+                          if (!ipc_get_lsc_data_ready_flag (idata))
+                            return;
+
+                          hostkb = args->host_kb;
+                          /* Get the OS release. TODO: have a list with
+                           * supported OS. */
+                          os_release =
+                            kb_item_get_str (hostkb, "ssh/login/release_notus");
+                          /* Get the package list. Currently only rpm support */
```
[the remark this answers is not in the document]
jjnicola: solved with https://github.com/greenbone/openvas-scanner/pull/1203/commits/38201f2b36495fd31376be73d1ff3d5b6ec994d4

nasl/nasl_scanner_glue.c
nichtsfrei: But my whole critique that I think it is bad design is still valid? 

So instead of two calls you need with that change calls to do the same thing. 

I still think it is bad design and that it would be better to handle notus specific restrictions within that `update_table_driven_lsc_data` method.

nichtsfrei: So for me a call with the data:
```
update_table_driven_lsc_data( pkg_list: pkg_list_notus, os_release: oskey_notus );
```

would be much easier to remember than having to call:
```
set_kb_item( name: "ssh/login/package_list_notus", value: oskey_notus);
set_kb_item( name: "ssh/login/release_notus", value: pkg_list_notus);
update_table_driven_lsc_data( );
```

misc/user_agent.c
```diff
@@ -111,7 +111,6 @@ user_agent_set (const gchar *ua)
 const gchar *
 user_agent_get (struct ipc_context *ipc_context)
 {
-  g_usleep (20 * 1000 * 1000);
```
ArnoStiefvater, line 114: Why is this removed or why was it there?

ArnoStiefvater (changes requested):
Empty change request to prevent accidental merges.

nasl/nasl_scanner_glue.c
jjnicola: Changes introduced with ec079f7 .
However, the changes in the plugin are not so trivial as setting the key or not, since other function are used, eg. in case of rpms, the function register_rpms is used for extra error and error return.

ArnoStiefvater (approved):

nichtsfrei (approved):

PR #1214, opened 2022-10-19
Fix: use main_kb to verify scan_id for inconsistencies

nichtsfrei:
This commit fixes the issue that the inconsistencies were not
necessarily checked against the shared database between ospd and
openvas.

Since ospd sets the scanner id into redis before calling openvas it is
necessary to use the main_kb to verify the locally set scanid with the
one in the database.

This is done by globally storing the main_kb via plugutils `set_main_kb`
and getting it via `get_main_kb`.

Fixes: SC-647

### DIFF
--- a/misc/plugutils.c
+++ b/misc/plugutils.c
@@ -62,7 +62,7 @@ plug_current_vhost (void)
   return current_vhost->value;
 }
 
-static int plug_fork_child (kb_t);
+static int plug_fork_child (kb_t, kb_t);
 
 void
 plug_set_dep (struct script_infos *args, const char *depname)
@@ -265,7 +265,7 @@ plug_get_host_fqdn (struct script_infos *args)
     return g_strdup (current_vhost->value);
   while (vhosts)
     {
-      int ret = plug_fork_child (args->key);
+      int ret = plug_fork_child (args->results, args->key);
 
       if (ret == 0)
         {
@@ -419,6 +419,39 @@ check_kb_inconsistency (kb_t main_kb)
   return -3;
 }
 
+// shared database between openvas and ospd.
+kb_t main_kb = NULL;
+
+/**
+ * @brief sets the shared database between ospd and openvas as a main_kb for
+ * further usage.
+ * @description this sets the given kb as a main_kb global variable. It is NOT
+ * threadsafe and must be called after each reconnect or fork.
+ *
+ * @param main_kb Current main kb.
+ *
+ */
+void
+set_main_kb (kb_t kb)
+{
+  main_kb = kb;
+}
+
+/**
+ * @brief gets the main_kb.
+ * @description returns the previously set main_kb; when asserts are enabled it
+ * will abort when main_kb is not set. However each usage must check if the
+ * return is NULL or not.
+ *
+ * @return the set main_kb
+ */
+kb_t
+get_main_kb (void)
+{
+  assert (main_kb);
+  return main_kb;
+}
+
 /**
  * @brief calls check_kb_inconsistency and logs as debug when local scan_id is
  missing.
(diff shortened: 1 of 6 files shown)


Kraemii (dismissed):

misc/plugutils.c
```diff
@@ -1095,6 +1136,8 @@ plug_fork_child (kb_t kb)
       sig_n (SIGTERM, _exit);
       mqtt_reset ();
       kb_lnk_reset (kb);
+      kb_lnk_reset (main);
+      set_main_kb (main);
```
jjnicola, line 1140: why does it need to be stored here again ? wouldn't it be enough just once at the beginning of the scan?

src/processes.c
```diff
@@ -174,12 +175,14 @@ init_child_signal_handlers (void)
...
 static void
-pre_fork_fun_call (struct ipc_context *ctx, void *args)
+pre_fn_call (struct ipc_context *ctx, void *args)
 {
   (void) ctx;
   (void) args;
+  kb_t main_kb = (kb_t) args;
   // in a chuld we clean up every preexisting context
   ipc_destroy_contexts (ipcc);
+  set_main_kb (main_kb);
```
jjnicola, line 185: Same here: why does it need to be stored here again ? wouldn't it be enough just once at the beginning of the scan?

src/openvas.c
```diff
@@ -220,9 +220,13 @@ overwrite_openvas_prefs_with_prefs_from_client (struct scan_globals *globals)
     return -1;
 
   snprintf (key, sizeof (key), "internal/%s/scanprefs", globals->scan_id);
+
   kb = kb_find (prefs_get ("db_address"), key);
   if (!kb)
     return -1;
+  // 2022-10-19: currently internal/%s/scanprefs are set by ospd which is the
+  // main_kb in our context
+  set_main_kb (kb);
```
jjnicola, line 229: IIUC, only here would be necessary to be stored.

src/processes.c
```diff
@@ -230,7 +234,7 @@ reuse_or_add_context (struct ipc_context *ctx)
  * FORKFAILED
  */
 pid_t
-create_ipc_process (ipc_process_func func, void *args)
+create_ipc_process (kb_t main_kb, ipc_process_func func, void *args)
```
jjnicola, line 237: If main kb is set via `set_main_kb()` in the main process (in openvas.c) at the beginning of the scan and stored it in global variable, it shouldn't be necessary to pass it as argument, since it is accessible from everywhere and from every forked process.

misc/plugutils.c
nichtsfrei: that's after the fork; probably just kb_lnk_reset main is sufficient; let me check.

jjnicola: I think yes, a link reset should be sufficient.

misc/plugutils.c
```diff
@@ -435,9 +468,11 @@ check_kb_inconsistency (kb_t main_kb)
  * @return 0 on success, -1 on inconsistency.
  */
 static int
-check_kb_inconsistency_log (kb_t kb, const char *name)
+check_kb_inconsistency_log (const char *name)
 {
   char *current_scan_id;
+  // we use the local kb entry only when there is no main_kb set.
```
jjnicola, line 474: This comment doesn't apply anymore.

ArnoStiefvater (changes requested):
Empty change request to not merge "accidentally"(waiting for release).

src/processes.c
```diff
@@ -230,7 +232,7 @@ reuse_or_add_context (struct ipc_context *ctx)
...
```
jjnicola, line 235: if only the global set variable is used for checking kb inconsistencies, what is the passed main kb used for?

jjnicola (changes requested):
see comments

jjnicola (dismissed):

ArnoStiefvater (approved):

PR #1215, opened 2022-10-19
Change: abort on db inconsistencies

nichtsfrei:
When finding a db inconsistency the running plugin should be abort
instead of just not writing to redis since the results will be lost
anyway.

This will shorten the runtime of plugins when a redis flush happened due
to memory issues or other causes.

This depends on: https://github.com/greenbone/openvas-scanner/pull/1214

Relates to: SC-647

### DIFF
--- a/misc/plugutils.c
+++ b/misc/plugutils.c
@@ -468,7 +468,7 @@ get_main_kb (void)
  * @return 0 on success, -1 on inconsistency.
  */
 static int
-check_kb_inconsistency_log (const char *name)
+check_kb_inconsistency_log (void)
 {
   char *current_scan_id;
   kb_t kb = get_main_kb ();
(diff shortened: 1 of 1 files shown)


misc/plugutils.c
```diff
@@ -484,19 +484,21 @@ check_kb_inconsistency_log (kb_t lkb, const char *name)
     {
     case -3:
       current_scan_id = kb_item_get_str (kb, ("internal/scanid"));
-      g_warning (
-        "%s: scan_id (%s) does not match global scan_id (%s); will not "
-        "store %s in kb.",
-        __func__, current_scan_id, get_scan_id (), name);
+      g_error ("%s: scan_id (%s) does not match global scan_id (%s); abort to "
```
jjnicola, line 487: g_error should not be use for runtime errors (see https://docs.gtk.org/glib/error-reporting.html). 
I found also, that in earlier glib version it raised SIGABORT. From version 2.39 it raises SIGTRAP.
Probably we should just use g_warning here.

jjnicola (changes requested):

jjnicola (approved):

PR #1229, opened 2022-11-18
Fix: missing NASL dependency netstat

nichtsfrei:
A direct call of netstat from NASL is done in:
- ./pre2008/netstat_portscan.nasl
therefore it is required in the Docker image if a user wants to execute
this script.


cfi-gb:
Note that you can see which external tools are required via `toolcheck.nasl` and cross-check that in the docker container via:

```
openvas-nasl -X toolcheck.nasl
```

bjoernricks (approved):

Kraemii (approved):

PR #1238, opened 2022-11-24
rs use peek

nichtsfrei:
Refactor: instead of using unhandled token peek the token instead.

To make the parsing a little bit easier to handle and a bit faster
compared to storing an unhandled token the lexer uses peek functionality
instead of advancing and storing unhandled token.

Depends on https://github.com/greenbone/openvas-scanner/pull/1236 https://github.com/greenbone/openvas-scanner/pull/1237

### DIFF
--- a/rust/nasl-syntax/src/infix_extension.rs
+++ b/rust/nasl-syntax/src/infix_extension.rs
@@ -371,7 +371,7 @@ mod test {
     #[test]
     fn repeat_call() {
         assert_eq!(
-            result("lol() x 2;"),
+            result("x() x 2;"),
             Operator(
                 X,
                 vec![
--- a/rust/nasl-syntax/src/infix_extension.rs
+++ b/rust/nasl-syntax/src/infix_extension.rs
@@ -2,11 +2,10 @@
 
 use crate::{
     error::SyntaxError,
-    {AssignOrder, Statement},
     lexer::{End, Lexer},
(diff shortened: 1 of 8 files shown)


rust/nasl-syntax/src/infix_extension.rs
```diff
@@ -362,4 +367,27 @@ mod test {
             )
         );
     }
+
+    #[test]
+    fn repeat_call() {
+        assert_eq!(
+            result("lol() x 2;"),
```
nichtsfrei, line 374: be more serious

jjnicola (dismissed):

rust/nasl-syntax/src/lexer.rs
```diff
@@ -132,16 +126,21 @@ impl<'a> Lexer<'a> {
...
                 left = nl;
                 if let End::Done(cat) = end {
                     end_statement = End::Done(cat);
                     break;
+                } else {
+                    // jump to the next without handling it as an error
+                    continue;
                 }
             }
+            // Due to peeking it can end up in and endlessloop
```
Kraemii, line 142:
```suggestion
            // Due to peeking it can end up in an endless loop
```

ArnoStiefvater (approved):

ArnoStiefvater (approved):

Kraemii (approved):

PR #1257, opened 2022-12-12
Fix: set the main kb during init openvas-nasl.

jjnicola:
**What**:
Fix: set the main kb during init openvas-nasl.
Jira: SC-753

**Why**:
This solves a segmentation fault when the script finishes. Before, it was trying to get the uninitialized main kb ctx

**How**:
Compiled with debug flag, without the patch it asserts on get_main_kb(). 
```
openvas-nasl -X -B -d -i /var/lib/openvas/plugins/ -t 192.168.0.1 --kb="Services/ssh=22" --kb="Secret/SSH/login=user" --kb="Secret/SSH/password=pass" gather-package-list.nasl
```

**Checklist**:

- [ ] Tests


n-thumann (approved):
Tested as described in the linked Jira ticket. It no longer segfaults and now works as expected.
Thanks for the fast fix!

ArnoStiefvater (approved):

PR #1286, opened 2023-01-12
rs includes

nichtsfrei:
Add includes

Enables includes by using a Loader trait, using a reference of a mutable
register instead of a register ownership within the interpreter.

This trait needs to be implemented in order to load the file content on
an include. For testing purposes there is the NoOpLoader.

### DIFF
--- a/rust/nasl-interpreter/src/loader.rs
+++ b/rust/nasl-interpreter/src/loader.rs
@@ -1,0 +1,78 @@
+//! This crate is used to load NASL code based on a name.
+
+use std::{path::Path, fs};
+
+use crate::error::InterpretError;
+
+/// Defines abstract Loader error cases
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum LoadError {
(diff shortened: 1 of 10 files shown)


ArnoStiefvater (approved):

rust/nasl-interpreter/src/loader.rs
```diff
@@ -0,0 +1,78 @@
...
+struct FSPluginLoader<'a> {
+    root: &'a Path,
+}
+
+impl<'a> Loader for FSPluginLoader<'a> {
+    fn load(&self, key: &str) -> Result<String, LoadError> {
+        let path = self.root.join(key);
+        if !path.is_file() {
+            return Err(LoadError::NotFound(format!(
+                "{} does not exist or is not accessable.",
```
Kraemii, line 58:
```suggestion
                "{} does not exist or is not accessible.",
```

Kraemii (approved):

ArnoStiefvater (approved):

PR #1291, opened 2023-01-18
rs use reference statement

nichtsfrei:
Change use reference in interpreter::resolve

Since the statement within resolve of interpreter does not mutate the
statement but reads it to generate Values it is changed to a reference
instead.

This makes it clear that Interpreter does not own the Statement and that
it is reading it making it easier to handle and clearer in the future.

### DIFF
--- a/rust/nasl-interpreter/src/error.rs
+++ b/rust/nasl-interpreter/src/error.rs
@@ -32,7 +32,7 @@ impl InterpretError {
 
     pub fn from_statement(stmt: &Statement, reason: String) -> Self {
         let (line, col) = stmt.as_token().map(|x| x.position).unwrap_or_default();
-        return InterpretError { reason, line, col };
+        InterpretError { reason, line, col }
     }
 
     pub fn internal_error(stmt: &Statement, dspl: &dyn Display) -> Self {
--- a/rust/nasl-interpreter/src/error.rs
+++ b/rust/nasl-interpreter/src/error.rs
@@ -2,7 +2,6 @@ use std::fmt::Display;
 
 use nasl_syntax::{Statement, SyntaxError};
 
-// TODO refactor error handling
 #[derive(Debug)]
 pub struct FunctionError {
     pub reason: String,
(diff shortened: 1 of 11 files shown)


ArnoStiefvater (approved):

rust/nasl-interpreter/src/error.rs
```diff
@@ -15,13 +14,26 @@ impl FunctionError {
 }
 
 #[derive(Debug, PartialEq, Eq)]
+/// Is used to represent an error while interpreting
 pub struct InterpretError {
+    /// Represents a human readable reason
     pub reason: String,
+    /// The line number within the script when that error occured
```
Kraemii, line 21:
```suggestion
    /// The line number within the script when that error occurred
```

rust/nasl-interpreter/src/error.rs
```diff
@@ -15,13 +14,26 @@ impl FunctionError {
...
+    ///
+    /// It starts 1.
     pub line: usize,
+    /// The colum number within the script when that error occured
```
Kraemii, line 25:
```suggestion
    /// The colum number within the script when that error occurred
```

rust/nasl-interpreter/src/error.rs
```diff
@@ -15,13 +14,26 @@ impl FunctionError {
...
+    ///
+    /// It starts 1.
     pub col: usize,
 }
 
 impl InterpretError {
+    /// Creates a new Error with line and col set to 0
+    ///
+    /// Use this only when there is no statement available.
+    /// If thre line as well as col is null Interpreter::resolve will replace it
```
Kraemii, line 35:
```suggestion
    /// If the line as well as col is null Interpreter::resolve will replace it 
```

rust/nasl-interpreter/src/lib.rs
```diff
@@ -1,81 +1,34 @@
+//! Is a crate to use Stataments from nasl-syntax and execute them.
```
Kraemii, line 1:
```suggestion
//! Is a crate to use Statements from nasl-syntax and execute them.
```

rust/nasl-interpreter/src/lib.rs
```diff
@@ -1,81 +1,34 @@
...
+pub use error::InterpretError;
 pub use interpreter::{Interpreter, NaslValue};
 pub use loader::*;
-use loader::NoOpLoader;
-use nasl_syntax::parse;
 use sink::{Sink, SinkError};
 
-pub type NaslFunction = fn(&str, &dyn Sink, &Register) -> Result<NaslValue, FunctionError>;
-pub fn lookup(function_name: &str) -> Option<NaslFunction> {
+// Is a type definition for inbuild functions
```
Kraemii, line 24:
```suggestion
// Is a type definition for built-in functions
```

Kraemii (approved):

ArnoStiefvater (approved):

PR #1292, opened 2023-01-20
rs get hostnames

nichtsfrei:
Implements get_host_name and get_host_names as NASL function.

Unfortunately there is no wrapper std wrapper for libc
- getnameinfo
- gethostname

therefore this Implements wrapper methods for that. Unfortunately those
are unix only, for windows it would require to use the windows specific
API instead.

### DIFF
--- a/rust/nasl-interpreter/src/built_in_functions/hostname.rs
+++ b/rust/nasl-interpreter/src/built_in_functions/hostname.rs
@@ -1,0 +1,275 @@
+use std::{
+    ffi::{c_char, CStr},
+    mem::{self, MaybeUninit},
+    net::{IpAddr, SocketAddr, SocketAddrV4, SocketAddrV6},
+};
+
+use libc::{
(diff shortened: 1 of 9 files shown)


rust/nasl-interpreter/src/built_in_functions/hostname.rs
```diff
@@ -0,0 +1,87 @@
...
+    _: &dyn Sink,
+    register: &Register,
+) -> Result<NaslValue, FunctionError> {
+    resolve_hostname(register).map(|x| NaslValue::Array(vec![NaslValue::String(x)]))
+}
+
+/// NASL function to get the current hostname
+///
+/// As of now (2023-01-20) there is no vhost handling.
+/// Therefore this function does load the registered TARGET and if it is an IP Addres resolves it via DNS instead.
```
Kraemii, line 47:
```suggestion
/// Therefore this function does load the registered TARGET and if it is an IP Address resolves it via DNS instead.
```

Kraemii (approved):

ArnoStiefvater (approved):

PR #1293, opened 2023-01-20
rs rand

nichtsfrei:
Implements the NASL rand function. Currently it is only available in
unix and reads 8 bytes of `/dev/urandom` to gather a random i64 number.

### DIFF
--- a/rust/nasl-interpreter/src/built_in_functions/rand.rs
+++ b/rust/nasl-interpreter/src/built_in_functions/rand.rs
@@ -1,0 +1,68 @@
+//! Defines NASL functions that deal with random and helpers
+
+use std::{
+    fs::File,
+    io::{self, Read},
+};
+
+use sink::Sink;
+
+use crate::{error::FunctionError, NaslFunction, NaslValue, Register};
+
+impl From<io::Error> for FunctionError {
+    fn from(e: io::Error) -> Self {
+        Self {
+            reason: format!("Internal error on rand {}", e),
+        }
+    }
+}
+
+#[inline]
+#[cfg(unix)]
+/// Reads 8 bytes from /dev/urandom and parses it to an i64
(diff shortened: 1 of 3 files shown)


Kraemii (approved):

rust/nasl-interpreter/src/built_in_functions/rand.rs
```diff
@@ -0,0 +1,68 @@
...
+            reason: format!("Internal error on rand {}", e),
+        }
+    }
+}
+
+#[inline]
+#[cfg(unix)]
+/// Reads 8 bytes from /dev/urandom and parses it to an i64
+fn random_impl() -> Result<i64, FunctionError> {
+    let mut rng = File::open("/dev/urandom")?;
```
ArnoStiefvater, line 24: Is this good enough?
In NASL we do lrand48()

https://github.com/greenbone/openvas-scanner/blob/ca12c694e1fd162ac0386e6ec47b9a5925c8e01a/nasl/nasl_misc_funcs.c#L57

nichtsfrei: But I think it is actually better than relying on linear congruential algorithm and 48-bit integer arithmetic entropy wise?

As far as I understand the man page  urandom uses a entropy pool which should be there after the boot process is finished. We could also use random which could block when the entropy pool isn't there yet but I think it is unnecessary in our case considering that we used lrand48.

If we want to have more variety we could use the rand crate and choose a generator: https://rust-random.github.io/book/guide-rngs.html

ArnoStiefvater: I am just a bit taken off by the need to open a file for random number generation.
We can leave it for now though. Better than before is good enough for now.

ArnoStiefvater (approved):

PR #1294, opened 2023-01-23
rs string functions

nichtsfrei:
Implement substr
    Implement string
    Implement strlen
    Implement tolower and toupper
    Implement raw_string
    Implement hexstr

### DIFF
--- a/rust/nasl-interpreter/src/built_in_functions/string.rs
+++ b/rust/nasl-interpreter/src/built_in_functions/string.rs
@@ -23,7 +23,17 @@ impl From<fmt::Error> for FunctionError {
     }
 }
 
-fn write_nasl_value(s: &mut String, value: &NaslValue) -> Result<(), FunctionError> {
+/// NASL function to parse numeric values into characters and combine with additional values
+fn raw_string(_: &str, _: &dyn Sink, register: &Register) -> Result<NaslValue, FunctionError> {
+    let positional = resolve_positional_arguments(register);
+    let mut s = String::with_capacity(2 * positional.len());
+    for p in positional {
+        write_raw_nasl_string(&mut s, &p)?;
+    }
+    Ok(s.into())
+}
+
+fn write_raw_nasl_string(s: &mut String, value: &NaslValue) -> Result<(), FunctionError> {
     match value {
         NaslValue::String(x) => write!(s, "{}", x),
         NaslValue::Number(x) => {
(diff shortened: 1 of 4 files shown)


rust/nasl-interpreter/src/built_in_functions/string.rs
```diff
@@ -0,0 +1,319 @@
...
+                (&s[lidx as usize..ridx as usize]).into()
+            }
+            _ => (&s[lidx as usize..]).into(),
+        })
+    }
+}
+
+/// NASL function to return  a hex presentation of given string as a positional argument.
+///
+/// If the positional arguments are empty it reutnrns NaslValue::Null.
```
Kraemii, line 161:
```suggestion
/// If the positional arguments are empty it returns NaslValue::Null.
```

rust/nasl-interpreter/src/built_in_functions/string.rs
```diff
@@ -0,0 +1,319 @@
...
+            Some(nv) => {
+                let ridx: i64 = nv.into();
+                (&s[lidx as usize..ridx as usize]).into()
+            }
+            _ => (&s[lidx as usize..]).into(),
+        })
+    }
+}
+
+/// NASL function to return  a hex presentation of given string as a positional argument.
```
Kraemii, line 159:
```suggestion
/// NASL function to return a hex representation of a given positional string argument.
```

rust/nasl-interpreter/src/built_in_functions/string.rs
```diff
@@ -0,0 +1,319 @@
...
+///
+/// If the positional arguments are empty it reutnrns NaslValue::Null.
+/// It only uses the first positional argument and when it is not a NaslValue:String than it returns NaslValue::Null.
+fn hexstr(_: &str, _: &dyn Sink, register: &Register) -> Result<NaslValue, FunctionError> {
+    let positional = resolve_positional_arguments(register);
+    // we already checked that positional has at least one argument
```
Kraemii, line 165: I am pretty sure this is not true, as resolve_positional_argument should return an empty vec, if no positional arguments were given. In the end we will return `NaslValue::NULL` in this case, as it is covered within the followed match block.

nichtsfrei: yes, that comment is a leftover of a previous solution. Should be deleted

Kraemii (approved):

ArnoStiefvater (approved):

PR #1314, opened 2023-02-03
Add: scan ID to log reference

Kraemii:
**What**:
To be able to detect corresponding log messages, the scan ID is now logged for every message for start and stop scan.
This PR depends on https://github.com/greenbone/gvm-libs/pull/735
SC-768

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/openvas.c
+++ b/src/openvas.c
@@ -560,6 +560,12 @@ openvas (int argc, char *argv[], char *env[])
     }
   tzset ();
 
+#ifdef LOG_REFERENCES_AVAILABLE
+  if (scan_id)
+    set_log_reference (scan_id);
+  if (stop_scan_id)
+    set_log_reference (stop_scan_id);
+#endif // LOG_REFERENCES_AVAILABLE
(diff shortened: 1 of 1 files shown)


src/openvas.c
```diff
@@ -560,6 +560,10 @@ openvas (int argc, char *argv[], char *env[])
     }
   tzset ();
 
+  if (scan_id)
```
nichtsfrei, line 563: To make deployment of GOS less error prone and since we currently don't have semantic version as of yet maybe it makes sense to add a
```
#define LOG_REFERENCES_AVAILABLE
```

in gvm libs to verify here if the functions are indeed available.

Otherwise we would require that gvm-libs are updated first before updating openvas. Since we currently don't have semantic versioning this could be messy otherwise.

jjnicola (dismissed):

jjnicola (approved):

nichtsfrei (approved):

PR #1318, opened 2023-02-07
Add: gzip() and gunzip() nasl functions

jjnicola:
**What**:
Add: gzip() and gunzip() nasl functions

**Why**:

**How**:

**Checklist**:

- [x] Tests


rust/nasl-interpreter/src/built_in_functions/misc.rs
```diff
@@ -40,12 +41,67 @@ pub fn dec2str(_: &str, _: &dyn Sink, register: &Register) -> Result<NaslValue,
...
+/// Compress given data with gzip, when headformat is set to 'gzip' it uses gzipheader.
+pub fn gzip(_: &str, _: &dyn Sink, register: &Register) -> Result<NaslValue, FunctionError> {
+    let data = match register.named("data") {
+        Some(ContextType::Value(NaslValue::String(x))) => x,
+        Some(ContextType::Value(NaslValue::Null)) => return Ok(NaslValue::Null),
+        _ => return Err(FunctionError::new("gzip", ("data", "data").into())),
+    };
+    let headformat = match register.named("headformat") {
+        Some(ContextType::Value(NaslValue::String(x))) => x,
+        Some(ContextType::Value(NaslValue::Null)) => "noheaderformat",
```
nichtsfrei, line 53: This line can be removed because the catchall `_` case is the same.

rust/nasl-interpreter/src/built_in_functions/misc.rs
```diff
@@ -40,12 +41,67 @@ pub fn dec2str(_: &str, _: &dyn Sink, register: &Register) -> Result<NaslValue,
     }
 }
 
+/// Compress given data with gzip, when headformat is set to 'gzip' it uses gzipheader.
+pub fn gzip(_: &str, _: &dyn Sink, register: &Register) -> Result<NaslValue, FunctionError> {
+    let data = match register.named("data") {
+        Some(ContextType::Value(NaslValue::String(x))) => x,
+        Some(ContextType::Value(NaslValue::Null)) => return Ok(NaslValue::Null),
+        _ => return Err(FunctionError::new("gzip", ("data", "data").into())),
```
nichtsfrei, line 49: in those cases you can also do

```
        x => return Err(FunctionError::new("gzip", ("data", "data", x).into())),
```

This reads like:
`("key", "type", x)` where x is the actual type. 

It is actually handled within the `From<...>` implementations within naslvalue.rs.

While explaining it, it may be better to replace it with another construct for readability sake.

rust/nasl-interpreter/src/error.rs
```diff
@@ -26,6 +26,13 @@ impl From<(&str, &str, &str)> for FunctionErrorKind {
     }
 }
 
+impl From<&str> for FunctionErrorKind {
+    fn from(value: &str) -> Self {
+        FunctionErrorKind::MissingArguments(vec![format!("Missing Argument {value}")])
```
nichtsfrei, line 31:
```
vec![value.to_owned()]
```

That it is a missing argument is clear from the enum type.

rust/nasl-interpreter/src/built_in_functions/misc.rs
```diff
@@ -62,6 +64,66 @@ pub fn unixtime(_: &str, _: &dyn Sink, _: &Register) -> Result<NaslValue, Functi
...
+        Some(ContextType::Value(NaslValue::Null)) => return Ok(NaslValue::Null),
+        Some(ContextType::Value(x)) => Vec::<u8>::from(x),
+        _ => return Err(FunctionError::new("gzip", ("data").into())),
+    };
+
+    let mut uncompress = ZlibDecoder::new(&data[..]);
+    let mut uncompressed = String::new();
+    match uncompress.read_to_string(&mut uncompressed) {
+        Ok(_) => return Ok(NaslValue::String(uncompressed)),
+        Err(_) => (),
```
nichtsfrei, line 115: I think I would put the second try in here so that it is clearer that it is a first try and the second is the aftermath.

Otherwise you need to keep track of the return statement to understand it.

rust/nasl-interpreter/src/built_in_functions/misc.rs
```diff
@@ -62,6 +64,66 @@ pub fn unixtime(_: &str, _: &dyn Sink, _: &Register) -> Result<NaslValue, Functi
...
+    };
+
+    let mut uncompress = GzDecoder::new(&data[..]);
+    let mut uncompressed = String::new();
+    match uncompress.read_to_string(&mut uncompressed) {
+        Ok(_) => return Ok(NaslValue::String(uncompressed)),
+        Err(_) => (),
+    };
+    return Ok(NaslValue::Null);
```
nichtsfrei, line 124: Is it correct that if both attempts fail it silently returns null instead of an error?

rust/nasl-interpreter/src/built_in_functions/misc.rs
```diff
@@ -62,6 +64,66 @@ pub fn unixtime(_: &str, _: &dyn Sink, _: &Register) -> Result<NaslValue, Functi
...
+    };
+    let headformat = match register.named("headformat") {
+        Some(ContextType::Value(NaslValue::String(x))) => x,
+        _ => "noheaderformat",
+    };
+
+    match headformat.to_string().eq_ignore_ascii_case("gzip") {
+        true => {
+            let mut e = GzEncoder::new(Vec::new(), Compression::default());
+            match e.write_all(&data) {
```
nichtsfrei, line 82: mhm I don't like the repetition everything after new call is identical. Maybe there is a better way.

rust/nasl-interpreter/src/built_in_functions/misc.rs
```diff
@@ -62,6 +64,66 @@ pub fn unixtime(_: &str, _: &dyn Sink, _: &Register) -> Result<NaslValue, Functi
...
+    let data = match register.named("data") {
+        Some(ContextType::Value(NaslValue::Null)) => return Ok(NaslValue::Null),
+        Some(ContextType::Value(x)) => Vec::<u8>::from(x),
+        _ => return Err(FunctionError::new("gzip", ("data").into())),
+    };
+
+    let mut uncompress = ZlibDecoder::new(&data[..]);
+    let mut uncompressed = String::new();
+    match uncompress.read_to_string(&mut uncompressed) {
+        Ok(_) => return Ok(NaslValue::String(uncompressed)),
```
nichtsfrei, line 114: Does the Ok statement contain the string? If so I think it is better to write:

```
Ok(result) => return Ok(result.into()),
```

jjnicola: Result in this case is an usize containing the amount bytes decompressed. The string is stored in the passed variable (uncomressed here).

nichtsfrei (approved):

rust/nasl-interpreter/src/built_in_functions/misc.rs
Kraemii: I looked into it and thought, that `write_all` is a function of the `Write` trait. Maybe there is a way of utilizing this?

Kraemii (approved):

PR #1319, opened 2023-02-08
Change mandatory_keys handling, adjust cves, simplify redis-sink

nichtsfrei:
Adjust CVES to store all CVEs instead of just the first, changing
mandatory keys to not contain double fields that are matched by the re:
field.

Add a Redis Namespace selector

Since the tag handling is already handled we remove the functional split
methods of nvt inside redis-sink.

**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/nasl-interpreter/src/built_in_functions/description.rs
+++ b/rust/nasl-interpreter/src/built_in_functions/description.rs
@@ -132,7 +132,9 @@ fn get_named_parameter<'a>(
     }
 }
 
-fn as_timeout_field(arguments: &[&NaslValue]) -> Result<NVTField, FunctionError> {
+type Transform = Result<NVTField, FunctionError>;
+
+fn as_timeout_field(arguments: &[&NaslValue]) -> Transform {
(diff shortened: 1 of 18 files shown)


jjnicola:
I am unsure about hardcoding the usage of the DB1 for the nvtcache. Currently it is Notus, which is updated via ospd-openvas which has the same mechanism for taking a free namespace. Are we sure that they don't compete and notus wont take the DB1 first? What happens if we decide to add a "new scanner" (or any other stuff) which is initialized in parallel? Also, it makes the code less reusable.

nichtsfrei:
Yes hardcoding isn't the right choice and I actually dislike that it is selecting a DB on init to begin with.

So you're and I'd rather change it completely and don't try to be smart about it by letting the client select a DB after initialization.

Regarding the parallelization of a new scanner I think it will be handled differently than currently but that's a completely different discussion.

jjnicola (dismissed):

Kraemii (approved):

jjnicola (approved):

PR #1322, opened 2023-02-08
Add specification for the upcoming http api

Kraemii:
**What**:
The first API specification for the upcoming HTTP API. It is derived from the OSP specification, transformed into JSON and adjusted.

The following commands are taken or left out from the OSP specification:
- ~~help: Get the help text.~~
- ~~get_performance: Return performance information from an external program.~~
- get_scans: Get a stored scan in buffer.
- delete_scan: Delete a finished scan.
- get_version: Return various versions.
- ~~check_feed: Perform sync feed self test and check lockfile status.~~
- ~~get_scanner_details: Return scanner description and parameters.~~
- get_vts: Return information about vulnerability tests, if offered by scanner.
- start_scan: Start a new scan.
- stop_scan: Stop a currently running scan.
- ~~get_memory_usage: Return memory usage information of the osp daemon.~~

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -9,6 +9,41 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
+name = "aead"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5c192eb8f11fc081b0fe4259ba5af04217d4e0faddd02417310a927911abd7c8"
+dependencies = [
+ "crypto-common",
+ "generic-array",
+]
+
+[[package]]
+name = "aes"
+version = "0.8.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "433cfd6710c9986c576a25ca913c39d66a6474107b406f34f91d4a8923395241"
+dependencies = [
+ "cfg-if",
+ "cipher",
+ "cpufeatures",
+]
+
+[[package]]
+name = "aes-gcm"
+version = "0.10.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "82e1366e0c69c9f927b1fa5ce2c7bf9eafc8f9268c0b9800729e8b267612447c"
+dependencies = [
+ "aead",
+ "aes",
+ "cipher",
+ "ctr",
+ "ghash",
+ "subtle",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


rust/api.md
```diff
@@ -0,0 +1,532 @@
...
+        }],
+        <a href="#vt_group">"vt_group"</a>: [string]
+      }
+    }
+  </code>
+</pre>
+- Status codes: 
+  - 200 Ok
+  - 400 Bad Request
+- Response: None
```
nichtsfrei, line 59: Since it is important to use a unique id I think it would be better to generate and return the ID in the backend rather than let the client choose one.

rust/api.md
```diff
@@ -0,0 +1,532 @@
...
+
+
+### get scan
+
+The get scan command returns information about a given scan. By default only basic information about the scan are requested. With the fields details the command also returns results.
+
+- Path: `/scan/get`
```
nichtsfrei, line 66: I would never use a action in a path since this is already expressed by the http call GET and than I wouldn't require the client to use GET with sending a json but use path parameter instead.

e.g.
```
GET /scan/{id}?pop_resuls=true&host_info=...
```

rust/api.md
```diff
@@ -0,0 +1,532 @@
...
+    }
+  </code>
+</pre>
+
+
+### delete scan
+
+The delete scan request deletes a scan from the scanner including all data corresponds to it. Unless the scan is not finished, this request will fail.
+
+- Path: `/scan/delete`
```
nichtsfrei, line 115: Here I would also use a path parameter and no action duplication instead of json:
```
DELETE /scan/{id}
```

rust/api.md
```diff
@@ -0,0 +1,532 @@
...
+  - 200 Ok
+  - 400 Bad Request
+  - 404 Scan not found
+- Response: None
+
+### stop scan
+
+The stop scan request stops a running scan. Unless the scan is not running, this request will fail.
+
+- Path: `/scan/stop`
```
nichtsfrei, line 134: Here I would suggest to use POST and put the action into a json:

```
POST /scan/{id}
-----
{
   "action": "stop"
}
```

nichtsfrei: maybe even patch instead of POST but I'm not sure here.

rust/api.md
Kraemii: For debug purposes I'd like to be able to set the scan ID myself, so it is possible to have scripts that get scan information and delete scans without having to adjust them for each scan.
But yes we could make the scan ID optional and response with one if none was given

nichtsfrei: But you can easily write script by parsing the returned json?

rust/api.md
Kraemii: Not sure how we should handle the GET command with the pop_results. Normally the GET method should only be used to retrieve data, but I guess it is more a guideline than an obligation. But nonetheless it seems a bit dirty.

rust/api.md
Kraemii: I guess PATCH would be better, as it sound more like modifying the resources

rust/api.md
nichtsfrei: true, maybe the whole pop is a thing of a past and we should require a proper delete call?

Kraemii: That could be rather dangerous, if exactly between a get scan and delete results call, new results are generated, we would miss them

nichtsfrei: if the client tries to delete why the scan in running and not stopped we should just not allow it?

nichtsfrei: Currently the pop results is just there due to memory limitation if we store on io or db either way I don't think deleting results while the scan is still running is as necessary as in the current setup.

nichtsfrei: Considering that we use PATCH for stopping we could also use PATCH to pop results and return the popped results.

With that we would have standard GET and if somebody for some reasons think he wants to pop results while a scan is running he could use PATCH with something like

```
{
"pop": true
}
```

on `/scan/{id}`.

Kraemii: Yes that sounds good

Kraemii: I am just thinking about moving the management of scans to the scanner, as we wanted to do it with eulabeia. That would allow to differentiate things a bit more and does not combine distinct commands like create and start scan. What do you think about it?

nichtsfrei: As long as we keep in mind that it has to be easy to integrate in existing systems, that may already have plugin selection, scheduling etc, I think it is a good idea.

rust/api.md
jjnicola: Just for the record. In the current GVM stack, the scan id is generated and sent by client and corresponds to the report ID. The same Report ID is sent in case of a task resume.

Kraemii: Yeah, I am aware of that. This approach however seems to be very error prone and the danger of loosing data is high. I think it is better to be able to implement the start/stop/resume mechanism directly into the scanner and let the scanner directly handle all data that is necessary. I think this would overall simplify this functionality and the client does not have to bother about saving data that is only necessary for this mechanic. Additionally it would make the scanner more reliable when saving such data to a storage, so it is able to recover from crashes without having the client to react with some actions.

Kraemii:
Closed in favor of https://github.com/greenbone/openvas-scanner/pull/1341

PR #1325, opened 2023-02-09
Change: Use a single workflow file for releases

bjoernricks:
**What**:

Merge manual and automatic releases (via PR label) into a single CI workflow.

Note: strategy calendar must be changed to patch for the stable branch.

**Why**:

Easier release and better maintenance.

DEVOPS-547


.github/workflows/release-pontos.yml
```diff
@@ -0,0 +1,34 @@
...
+        with:
+          python-version: "3.10"
+          conventional-commits: true
+          github-user: ${{ secrets.GREENBONE_BOT }}
+          github-user-mail: ${{ secrets.GREENBONE_BOT_MAIL }}
+          github-user-token: ${{ secrets.GREENBONE_BOT_TOKEN }}
+          gpg-key: ${{ secrets.GPG_KEY }}
+          gpg-fingerprint: ${{ secrets.GPG_FINGERPRINT }}
+          gpg-passphrase: ${{ secrets.GPG_PASSPHRASE }}
+          strategy: calendar
```
nichtsfrei, line 33: that seems incorrect? although we were calendar based we still stick with the 22.4 notation until we make the switch to proper semantic and even than we will probably either stick with 22.x.x or make the switch to 23.x than.

y0urself: Yeah. We always use the `stable` branch for releases. Therefore we will change it in the backport and leave it `calendar` in the default branch.

nichtsfrei: ah, ok.

nichtsfrei (approved):

PR #1329, opened 2023-02-10
Introduce Data type

nichtsfrei:
In NASL `typeof('hello')` returns data therefore we need to add a new
Data TokenCategory to represent that.

This commit allows differentiation between
- "test" -> String("test")
- 'test' -> Data([116,101,115,116])

However in NASL scripts strings and data are used fluently that requires
all string based methods to handle data as an ordinary string.


jjnicola (approved):
Cool! That was fast!

Kraemii (approved):

PR #1335, opened 2023-02-14
rs ci verify feed update

nichtsfrei:
Adds a tests that verify `openvas -u` and `nasl-cli feed update` behave the same by executing `feed-verifier`.

Since we support `debian:stable` it additionally checks that `nasl-cli` is able to compile with libc 2.31 by utilizing ubuntu-20.4 instead of the latest ubuntu version.

Depends on https://github.com/greenbone/openvas-scanner/pull/1340

### DIFF
--- a/.github/workflows/rustification.yaml
+++ b/.github/workflows/rustification.yaml
@@ -70,69 +70,106 @@ jobs:
           - nasl-cli
     steps:
       - uses: actions/checkout@v3
-      - run: rustfmt --check ${{ matrix.crates }}/**/*.rs
-  build-releases:
-    runs-on: ubuntu-latest
+  releases:
+    runs-on:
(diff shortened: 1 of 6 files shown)


.github/workflows/rustification.yaml
```diff
@@ -0,0 +1,173 @@
...
+    # otherwise we would need to use target on build to not have a direct dependency to libc
+    # we need to build within a container based on the earliest glibc version we want to support
+    # the reason for that is that libc does use symbol versioning. Otherwise it can happen that 
+    # the built release is requiring a newer libc version
+    runs-on:
+     - self-hosted
+     - self-hosted-generic
+    container:
+      # we use rust latest; that seems to be built debian:latest
+      image: rust:latest
```
y0urself, line 84:
```suggestion
    container:
      # we use rust latest; that seems to be built debian:latest
      image: rust:latest
      options: --init
```

PR #1344, opened 2023-02-20
Add: time date nasl functions

jjnicola:
**What**:
Add: localtime()
Add: mktime()
Add: sleep()
Add: usleep()

Also, solves the cargo audit CI test. It was failing because a vulnerability in a default feature in the chrono crate. 
Disable the chrono's default features and enable only the required.

**Why**:

**How**:

**Checklist**:

- [x] Tests


rust/nasl-interpreter/src/built_in_functions/misc.rs
```diff
@@ -154,6 +182,135 @@ pub fn gunzip(_: &str, _: &dyn Sink, register: &Register) -> Result<NaslValue, F
...
+/// Takes seven named arguments sec, min, hour, mday, mon, year, isdst and returns the Unix time.
+pub fn mktime(_: &str, _: &dyn Sink, register: &Register) -> Result<NaslValue, FunctionError> {
+
+    let sec;
+    match register.named("sec") {
+        Some(ContextType::Value(NaslValue::Number(x))) => {sec = *x as u32},
+        _ => {sec = 0;},
+    };
+    let min;
+    match register.named("min") {
```
nichtsfrei, line 194: I it is easier to read when you write it like this instead:
```
let min =  match register.named("min") {
  Some(ContextType::Value(NaslValue::Number(x))) => x as u32,
  _ => 0
};
```

rust/nasl-interpreter/src/built_in_functions/misc.rs
```diff
@@ -154,6 +182,135 @@ pub fn gunzip(_: &str, _: &dyn Sink, register: &Register) -> Result<NaslValue, F
...
+                LocalResult::Single(x) => {tictac_local = x;},
+                _ => (),
+            };
+        }
+        strfmt = tictac_local.format("%S %M %H %d %m %Y %w %j").to_string();
+        tictac = strfmt.split(" ").collect();
+
+    }
+
+    let mut date: HashMap::<String, NaslValue> = HashMap::new();
```
nichtsfrei, line 298: You can directly initialize a HashMap when you do:
```
let date = HashMap::from([
   ("sec".into(), tictac[0].parse::<i64>().unwrap_or_default().into()),
... 
  ]);
```

this way it is immediately allocated and it doesn't require to allocate on insert.

rust/nasl-interpreter/src/built_in_functions/misc.rs
```diff
@@ -154,6 +182,128 @@ pub fn gunzip(_: &str, _: &dyn Sink, register: &Register) -> Result<NaslValue, F
...
+                if *x > 0 {
+                    secs = *x as i64;
+                }
+                else {
+                    return Err(FunctionError::new("localtime", ("0", "numeric").into()));
+                };
+            },
+            NaslValue::String(x) => {
+                let secstr: Vec<&str> = x.split(".").collect();
+                let r_secs = secstr[0].to_string().parse::<i64>();
```
nichtsfrei, line 257: This can be simplified to
```
match x.split('.').map(|x|x.parse::<i64>()).next() {
    Some(Ok(x)) => x,
   _ => return Err(...),
};
```
with that you don't need to collect and only process once.

rust/nasl-interpreter/src/built_in_functions/misc.rs
```diff
@@ -154,6 +182,128 @@ pub fn gunzip(_: &str, _: &dyn Sink, register: &Register) -> Result<NaslValue, F
...
+    match named {
+        Some(ContextType::Value(NaslValue::Number(x))) => {utc_flag = *x != 0i64;},
+        Some(ContextType::Value(NaslValue::Boolean(x))) => {utc_flag = *x;},
+        _ => {utc_flag = false;},
+    };
+        
+    let mut tictac_local: chrono::DateTime<Local> = Local::now();
+    let mut tictac_utc: chrono::DateTime<Utc> = Utc::now();
+    let tictac: Vec<&str>;
+    let mut secs: i64 = 0;
```
nichtsfrei, line 242: if you simplify the if else constructs you don't need mutable secs.

rust/nasl-interpreter/src/built_in_functions/misc.rs
```diff
@@ -154,6 +182,128 @@ pub fn gunzip(_: &str, _: &dyn Sink, register: &Register) -> Result<NaslValue, F
...
+    match r_dt {
+        LocalResult::Single(x) => Ok(NaslValue::Number(x.naive_local().timestamp() - offset as i64)),
+        _ => Ok(NaslValue::Null),
+    }
+}
+
+
+
+/// Returns an dict(mday, mon, min, wday, sec, yday, isdst, year, hour) based on optional given time in seconds and optional flag if utc or not.
+pub fn localtime(_: &str, _: &dyn Sink, register: &Register) -> Result<NaslValue, FunctionError> {
```
nichtsfrei, line 230: This whole function seems to be unnecessary hard to read and I get a lot of clippy warnings.

rust/nasl-interpreter/src/built_in_functions/misc.rs
```diff
@@ -154,6 +182,128 @@ pub fn gunzip(_: &str, _: &dyn Sink, register: &Register) -> Result<NaslValue, F
...
+    let named = register.named("utc");
+    let utc_flag;
```
nichtsfrei, line 232: unnecessary define and later assignment, use match to return and assign immediately.

rust/nasl-interpreter/src/built_in_functions/misc.rs
```diff
@@ -154,6 +182,128 @@ pub fn gunzip(_: &str, _: &dyn Sink, register: &Register) -> Result<NaslValue, F
...
+
+    let strfmt;
+    if utc_flag {
+        if secs != 0 {
+            match Utc.timestamp_opt(secs, 0) {
+                LocalResult::Single(x) => {tictac_utc = x;},
+                _ => (),
+            };
+        }
+        strfmt = tictac_utc.format("%S %M %H %d %m %Y %w %j").to_string();
```
nichtsfrei, line 275: is there no simpler way to just get the fields of a datetime construct than creating a format string, than create a vector tbased on that to get the index instead of just calling the corresponding methods of the date struct? This seems a bit complicated.

rust/nasl-interpreter/src/built_in_functions/misc.rs
nichtsfrei: After looking into it, I would rather rewrite it to:

```
fn create_localtime_map<T>(date: chrono::DateTime<T>) -> HashMap<String, NaslValue>
where
    T: chrono::TimeZone,
{
    HashMap::from([
        ("sec".to_string(), NaslValue::from(date.second() as i64)),
        ("min".to_string(), NaslValue::from(date.minute() as i64)),
        ("hour".to_string(), NaslValue::from(date.hour() as i64)),
        ("mday".to_string(), NaslValue::from(date.day() as i64)),
        ("mon".to_string(), NaslValue::from(date.month() as i64)),
        ("year".to_string(), NaslValue::from(date.year() as i64)),
        ("wday".to_string(), NaslValue::from(date.weekday() as i64 + 1)),
        ("yday".to_string(), NaslValue::from(date.ordinal() as i64)),
        // TODO: fix isdst
        ("isdst".to_string(), NaslValue::from(0)),
    ])
}

/// Returns an dict(mday, mon, min, wday, sec, yday, isdst, year, hour) based on optional given time in seconds and optional flag if utc or not.
pub fn localtime(_: &str, _: &dyn Sink, register: &Register) -> Result<NaslValue, FunctionError> {
    let utc_flag = match register.named("utc") {
        Some(ContextType::Value(NaslValue::Number(x))) => *x != 0,
        Some(ContextType::Value(NaslValue::Boolean(x))) => *x,
        _ => false,
    };

    let secs = match register.positional() {
        [] => 0,
        [x0, ..] => i64::from(x0),
    };
    let date = match (utc_flag, secs) {
        (true, 0) => create_localtime_map(Utc::now()),
        (true, secs) => match Utc.timestamp_opt(secs, 0) {
                LocalResult::Single(x) => create_localtime_map(x),
            _ => create_localtime_map(Utc::now()),
        }
        (false, 0) => create_localtime_map(Local::now()),
        (false, secs) => match Local.timestamp_opt(secs, 0) {
            LocalResult::Single(x) => create_localtime_map(x),
            _ => create_localtime_map(Local::now())
        },
    };

    Ok(NaslValue::Dict(date))
}
```

nichtsfrei (changes requested):

rust/nasl-interpreter/src/built_in_functions/misc.rs
jjnicola: Feel free add a new commit with your solution.

nichtsfrei (dismissed):

nichtsfrei (approved):

ArnoStiefvater (approved):

PR #1345, opened 2023-02-22
Change: Improve stop_denial. Test the host if still alive with boreas

jjnicola:
**What**:
Improve stop_denial. Test the host if still alive with boreas
Jira: SC-772

**Why**:
Some plugins produce the host scan to finish earlier because false positive.

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/misc/heartbeat.c
+++ b/misc/heartbeat.c
@@ -1,12 +1,11 @@
-/* Portions Copyright (C) 2009-2022 Greenbone Networks GmbH
- * Portions Copyright (C) 2006 Software in the Public Interest, Inc.
- * Based on work Copyright (C) 1998 - 2006 Tenable Network Security, Inc.
+/* Copyright (C) 2023 Greenbone Networks GmbH
  *
- * SPDX-License-Identifier: GPL-2.0-only
+ * SPDX-License-Identifier: GPL-2.0-or-later
  *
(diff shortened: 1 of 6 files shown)


nasl/nasl_misc_funcs.c
```diff
@@ -287,7 +289,12 @@ nasl_end_denial (lex_ctxt *lexic)
         }
     }
 
-  retc->x.i_val = 0;
+  // Services seem to not respond.
+  // Last test with boreas
+  if (check_host_still_alive (kb, plug_current_vhost ()))
```
nichtsfrei, line 294: mhm, if boreas is either disabled or has an error isn't it better to follow the assumption that it is dead?

nichtsfrei (dismissed):

misc/heartbeat.c
```diff
@@ -0,0 +1,93 @@
+/* Portions Copyright (C) 2009-2022 Greenbone Networks GmbH
```
ArnoStiefvater, line 1: License header is wrong, same for the .h file.

ArnoStiefvater (changes requested):

ArnoStiefvater (approved):

Kraemii (approved):

PR #1360, opened 2023-03-13
Change Company name to Greenbone AG

Tacire:
## What

Change Company name to Greenbone AG

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (C) 2011-2022 Greenbone Networks GmbH
+# Copyright (C) 2011-2022 Greenbone AG
 #
 # SPDX-License-Identifier: GPL-2.0-or-later
 #
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ issue to the Greenbone Support Portal.
 
 ## Maintainer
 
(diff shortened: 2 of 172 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2011-2022 Greenbone Networks GmbH
+# Copyright (C) 2011-2022 Greenbone AG
```
nichtsfrei, line 1: I think in those cases it must be either adjusted to 2023 or be left at GmbH since 2022 it was GmbH while the AG is the legal successor so I'd guess it would be fine in the cases that weren't changed?

nichtsfrei (changes requested):

CMakeLists.txt
nichtsfrei: will be ignored

nichtsfrei (approved):

ArnoStiefvater (changes requested):
There seems to be some internal discussion still going on about this changes. Therefore blocking this PR for now.

Tacire:
An update of the copyrightyear can at least be easily done using pontos https://greenbone.github.io/pontos/tools.html#pontos-update-header

ArnoStiefvater:
Closing this PR. We will make a new PR with correct headers.

PR #1367, opened 2023-03-21
Add additional API documentation for authentication and response header

Kraemii:
**What**:
This PR contains additional documentation regarding the new Scanner API. It includes documentation for
- authentication
- response header (meta information)

SC-780

**Why**:

**How**:

**Checklist**:

- [ ] Tests


rust/doc/api.md
```diff
@@ -0,0 +1,55 @@
...
+## API Response Header
+
+The API Response Header contains additional information, which can be used by the client to verify the used versions of the scanner components, as well as the version of the API, so it can do the appropriate calls.
+The header will contain the following fields:
+- `api-version` The version of the API
+- `scanner-version` The version of the Scan manager
+- `openvas-version` The version of the OpenVAS scanner
+- `notus-version` The version of the Notus scanner
+- `feed-version` The version of the feed used for the VTs
+- `authentication` The authentication method used
```
nichtsfrei, line 51: I would rather change it as `supported authentication methods` than what is used.

rust/doc/api.md
```diff
@@ -0,0 +1,55 @@
...
+More details about this authentication method follows with its implementation.
+
+## API Response Header
+
+The API Response Header contains additional information, which can be used by the client to verify the used versions of the scanner components, as well as the version of the API, so it can do the appropriate calls.
+The header will contain the following fields:
+- `api-version` The version of the API
+- `scanner-version` The version of the Scan manager
+- `openvas-version` The version of the OpenVAS scanner
+- `notus-version` The version of the Notus scanner
```
nichtsfrei, line 49: do we want to expose openvas-version, notus-version and scanner-version?

rust/doc/api.md
```diff
@@ -0,0 +1,55 @@
...
+In contrast to the API keys this can be used as a bidirectional authentication.
+This means the client can authenticate the server first, then the server authenticate the client.
+
+More details about this authentication method follows with its implementation.
+
+## API Response Header
+
+The API Response Header contains additional information, which can be used by the client to verify the used versions of the scanner components, as well as the version of the API, so it can do the appropriate calls.
+The header will contain the following fields:
+- `api-version` The version of the API
```
nichtsfrei, line 46: I think supported versions (e.g. `v1`, `v23-25`,...) may be better than a single version?

nichtsfrei (changes requested):

rust/doc/api.md
Kraemii: My thought on this was to get information about what the scanner is actual capable of. E.g. if the notus scanner is not available for this scanner, this can be took into account.
But also maybe such things should be handled within the scanner internally, so the client does not need such information. What I mean of this, in the end the client does not care if either OpenVAS or Notus handles VTs.

nichtsfrei (dismissed):

ArnoStiefvater (approved):

nichtsfrei (approved):

ArnoStiefvater (approved):

PR #1385, opened 2023-04-13
Rs webserver

Kraemii:
**What**:
Implementation of the specified API
Based on https://github.com/greenbone/openvas-scanner/pull/1322

This PR contains
- A model module, containing all JSON models used for the API
- A webserver serving as the new API implementation
- Default implementations for the Scan- and VT-Management for testing purposes
- Error handling and JSON responses for it

The authentication via API Key and Certificates will be handled in another PR

SC-807

**Why**:

**How**:

**Checklist**:

- [ ] Tests


ArnoStiefvater:
> ScannerAPI specification does not define parameters for command "/vts".

It was only present in a early draft. In its current state we only return the list of supported OIDs as defined in the API spec.

nichtsfrei (approved):

jjnicola (approved):

PR #1386, opened 2023-04-14
CI: remove dependency on build image, add release process for multiple projects

nichtsfrei:
CI: Implement a new release process for multi-project repository

Updates our CI release process to accommodate the switch from a
single project repository to a multi-project one. The new release process
includes the following changes:

1. We now base releases on git tags instead of individual projects.
1. We use major, minor, and patch releases instead of just patch releases.
1. Releases are made from the main branch instead of stable or unstable branches.
1. For major releases, we create a version branch to handle hotfixes.
1. Each project within the repository has the same version.
1. We've added nasl-cli as a release artifact for linux-x86_64 and aarch64.

To initiate the release process, you can either add one of the following labels
to a pull request:
- `major_release`,
- `minor_release`,
- `patch_release`.

Once the PR is merged, the release process will be triggered.

Alternatively, you can initiate the release via the GitHub action page.

Note that we are currently working on creating a docker image for releases,
which will require new docker tag layout and compose changes.

Switches from build-image to gvm-libs and install dependencies.

Enhances the robustness of the CI process by using GVM-libs and
installing the required dependencies, instead of relying on a pre-built
image that must be pushed before running CI tests.
This approach eliminates the need to push dependencies beforehand and
wait for a docker image to be pushed before utilizing it.

Furthermore, we have moved the smoketest.yml step to nasl.yml since it
is also NASL-based.

The build and tests now verify GVM-libs unstable as well as stable to
ensure both downwards and upwards compatibility when on stable.

Our intention is to be notified of any breaking changes.


.docker/build.Dockerfile
```diff
@@ -1,30 +1,31 @@
 ARG VERSION=unstable
 
 FROM greenbone/gvm-libs:$VERSION
+LABEL deprecayed="This image is deprecated and may be removed soon."
```
jjnicola, line 4:
```suggestion
LABEL deprecated="This image is deprecated and may be removed soon."
```

jjnicola (approved):

jjnicola (changes requested):
other than the typo, it looks good to me.

.github/workflows/release.yml
```diff
@@ -0,0 +1,182 @@
...
+      - name: "NEW_VERSION"
+        run: |
+          echo "NEW_VERSION=$(sh .github/enhance_version.sh ${{ env.LATEST_VERSION }} ${{ env.RELEASE_KIND }})" >> $GITHUB_ENV
+      - name: NEW_VERSION != NULL
+        run: ([ -n "${{ env.NEW_VERSION }}" ])
+      - name: set git credentials
+        run: |
+             git config --global user.email "${{ secrets.GREENBONE_BOT_MAIL }}"
+             git config --global user.name "${{ secrets.GREENBONE_BOT }}"
+      - name: "create working branch for previous major relese (${{ env.LATEST_VERSION }})"
```
Kraemii, line 110:
```suggestion
      - name: "create working branch for previous major release (${{ env.LATEST_VERSION }})"
```

Kraemii (approved):

jjnicola (approved):

PR #1400, opened 2023-04-26
Add: nasl-cli execute --target option

jjnicola:
**What**:
Add: nasl-cli execute --target option
Jira: SC-808

**Why**:
This allows to test nasl scripts/ nasl functions not only against localhost, but also different targets

**How**:
`nasl-cli execute --target <remote-host> <nasl-scritp>`

**Checklist**:

- [ ] Tests


rust/nasl-interpreter/src/context.rs
```diff
@@ -243,6 +243,8 @@ impl NaslContext {
 pub struct Context<'a, K> {
     /// key for this context. A name or an OID
     key: &'a K,
+    /// target to run a scan against
+    target: &'a String,
```
nichtsfrei, line 247: should be &str

nichtsfrei (approved):

nichtsfrei (approved):

ArnoStiefvater (approved):

PR #1401, opened 2023-04-26
Change: set the old script greenbone-nvt-sync as deprecated.

jjnicola:
**What**:
Change: set the old script greenbone-nvt-sync as deprecated.
However its installation is set ON as default.
Jira: SC-773

**Why**:

**How**:

**Checklist**:

- [ ] Tests


bjoernricks (commented):
Maybe we should add a deprecation message that's shown on every run to the `greenbone-nvt-sync` script too?

bjoernricks (approved):

bjoernricks:
Thanks a lot!

nichtsfrei (approved):

PR #1410, opened 2023-05-02
Fix: prod.Dockerfile multi-arch support and workflow

nichtsfrei:
When running the 'docker build push' action in a matrix, it overrides the previously existing platform tag instead of adding it.

To solve this issue, differentiate defined platforms within the 'prod.Dockerfile' and adjust the settings accordingly.


bjoernricks (approved):
Just remembered this nasty issue :-(

PR #1416, opened 2023-05-04
Add: Implement packet forgery nasl functions in rust.

jjnicola:
**What**:
Add: Implement packet forgery nasl functions in rust.
Jira: SC-821

**Why**:

**How**:

**Checklist**:

- [ ] Tests


rust/nasl-interpreter/src/built_in_functions/packet_forgery.rs
```diff
@@ -0,0 +1,2872 @@
...
+///  
+/// Returns the modified IP datagram
+fn set_ip_elements<K>(
+    register: &Register,
+    _configs: &Context<K>,
+) -> Result<NaslValue, FunctionErrorKind> {
+    let mut buf = match register.named("ip") {
+        Some(ContextType::Value(NaslValue::Data(d))) => d.clone(),
+        _ => {
+            return Err(FunctionErrorKind::Diagnostic(
```
nichtsfrei, line 265: is that really a diagnostic error? to me it seems more like a usage error.

rust/nasl-interpreter/src/built_in_functions/packet_forgery.rs
```diff
@@ -0,0 +1,2872 @@
...
+                    .to_string(),
+                Some(NaslValue::Null),
+            ));
+        }
+    };
+    let length = match register.named("length") {
+        Some(ContextType::Value(NaslValue::Number(x))) => *x as usize,
+        _ => {
+            return Err(FunctionErrorKind::Diagnostic(
+                "Usage : insert_ip_options(ip:<ip>, code:<code>, length:<len>, value:<value"
```
nichtsfrei, line 482: It seems wrong on an usage error to not abort but continue with a default? 

I think aborting seems to be the better choice? 

Even when not I think we should introduce a special case rather than using diagnostic everywhere.

rust/nasl-interpreter/src/built_in_functions/packet_forgery.rs
```diff
@@ -0,0 +1,2872 @@
...
+                    .to_string(),
+                Some(NaslValue::Null),
+            ));
+        }
+    };
+    let value = match register.named("value") {
+        Some(ContextType::Value(NaslValue::String(x))) => x.as_bytes(),
+        Some(ContextType::Value(NaslValue::Data(x))) => x,
+        _ => {
+            return Err(FunctionErrorKind::Diagnostic(
```
nichtsfrei, line 492: usage vs diagnostic error.

rust/nasl-interpreter/src/built_in_functions/packet_forgery.rs
```diff
@@ -0,0 +1,2872 @@
...
+/// Returns an TCP element from a IP datagram.
+fn get_tcp_element<K>(
+    register: &Register,
+    _configs: &Context<K>,
+) -> Result<NaslValue, FunctionErrorKind> {
+    let buf = match register.named("tcp") {
+        Some(ContextType::Value(NaslValue::Data(d))) => d.clone(),
+        _ => {
+            return Err(FunctionErrorKind::Diagnostic(
+                "get_tcp_element: missing <tcp> field".to_string(),
```
nichtsfrei, line 717: usage error

rust/nasl-interpreter/src/built_in_functions/packet_forgery.rs
```diff
@@ -0,0 +1,2872 @@
...
+            "th_off" => Ok(NaslValue::Number(tcp.get_data_offset() as i64)),
+            "th_flags" => Ok(NaslValue::Number(tcp.get_flags() as i64)),
+            "th_win" => Ok(NaslValue::Number(tcp.get_window() as i64)),
+            "th_sum" => Ok(NaslValue::Number(tcp.get_checksum() as i64)),
+            "th_urp" => Ok(NaslValue::Number(tcp.get_urgent_ptr() as i64)),
+            "th_data" => Ok(NaslValue::Data(tcp.payload().to_vec())),
+            _ => Ok(NaslValue::Null),
+        },
+        _ => Err(FunctionErrorKind::Diagnostic(
+            "get_tcp_element: missing valid tcp element".to_string(),
```
nichtsfrei, line 752: usage error?

rust/nasl-interpreter/src/built_in_functions/packet_forgery.rs
```diff
@@ -0,0 +1,2872 @@
...
+            3 => Ok(NaslValue::Number(window)),
+            4 => Ok(NaslValue::Number(sack_permitted)),
+            8 => Ok(NaslValue::Array(timestamps)),
+            _ => Err(FunctionErrorKind::Diagnostic(
+                "get_tcp_option: invalid option".to_string(),
+                Some(NaslValue::Null),
+            )),
+        },
+        _ => Err(FunctionErrorKind::Diagnostic(
+            "get_tcp_option: missing option argument".to_string(),
```
nichtsfrei, line 842: usage error?

rust/nasl-interpreter/src/built_in_functions/packet_forgery.rs
```diff
@@ -0,0 +1,2872 @@
...
+    }
+
+    match register.named("option") {
+        Some(ContextType::Value(NaslValue::Number(el))) => match el {
+            2 => Ok(NaslValue::Number(max_seg)),
+            3 => Ok(NaslValue::Number(window)),
+            4 => Ok(NaslValue::Number(sack_permitted)),
+            8 => Ok(NaslValue::Array(timestamps)),
+            _ => Err(FunctionErrorKind::Diagnostic(
+                "get_tcp_option: invalid option".to_string(),
```
nichtsfrei, line 837: usage error?

rust/nasl-interpreter/src/built_in_functions/packet_forgery.rs
```diff
@@ -0,0 +1,2872 @@
...
+/// - update_ip_len: is a flag (TRUE by default). If set, NASL will recompute the size field of the IP datagram.
+fn set_tcp_elements<K>(
+    register: &Register,
+    _configs: &Context<K>,
+) -> Result<NaslValue, FunctionErrorKind> {
+    let buf = match register.named("tcp") {
+        Some(ContextType::Value(NaslValue::Data(d))) => d.clone(),
+        _ => {
+            return Err(FunctionErrorKind::Diagnostic(
+                "get_tcp_element: missing <tcp> field".to_string(),
```
nichtsfrei, line 871: usage error

rust/nasl-interpreter/src/built_in_functions/packet_forgery.rs
```diff
@@ -0,0 +1,2872 @@
...
+
+    // pnet will panic if the total length set in the ip datagram field does not much with the total length.
+    // Therefore, the total length is set to the right one before setting the payload.
+    // By default it was always updated, but if desired, the original length is set again after setting the payload.
+    let original_ip_len = pkt.get_total_length();
+    pkt.set_total_length(l as u16);
+    pkt.set_payload(&fin_tcp_buf);
+    match register.named("update_ip_len") {
+        Some(ContextType::Value(NaslValue::Boolean(l))) if !(*l) => {
+            println!("set the ori again");
```
nichtsfrei, line 1009: don't use println!

rust/nasl-interpreter/src/built_in_functions/packet_forgery.rs
```diff
@@ -0,0 +1,2872 @@
...
+    }
+
+    for tcp_seg in positional.iter() {
+        match tcp_seg {
+            NaslValue::Data(data) => {
+                let ip = match packet::ipv4::Ipv4Packet::new(data) {
+                    Some(ip) => ip,
+                    None => {
+                        return Err(FunctionErrorKind::Diagnostic(
+                        "Usage : insert_ip_options(ip:<ip>, code:<code>, length:<len>, value:<value"
```
nichtsfrei, line 1254: usage error

rust/nasl-interpreter/src/built_in_functions/packet_forgery.rs
```diff
@@ -0,0 +1,2872 @@
...
+                            .to_string(),
+                        Some(NaslValue::Null),
+                    ));
+                    }
+                };
+
+                match packet::tcp::TcpPacket::new(ip.payload()) {
+                    Some(pkt) => {
+                        println!("------\n");
```
nichtsfrei, line 1263: don't use println!

rust/nasl-interpreter/src/built_in_functions/packet_forgery.rs
```diff
@@ -0,0 +1,2872 @@
...
+                                f.push('|')
+                            };
+                            f.push_str("TH_URG");
+                        }
+                        if f.is_empty() {
+                            f.push_str(" 0");
+                        } else {
+                            f.push_str(&format!(" {:?}", flags));
+                        };
+                        println!("{}", f);
```
nichtsfrei, line 1324: don't use println!

rust/nasl-interpreter/src/built_in_functions/packet_forgery.rs
```diff
@@ -0,0 +1,2872 @@
...
+/// Returns the modified IP datagram or NULL on error.
+fn forge_udp_packet<K>(
+    register: &Register,
+    _configs: &Context<K>,
+) -> Result<NaslValue, FunctionErrorKind> {
+    let mut ip_buf = match register.named("ip") {
+        Some(ContextType::Value(NaslValue::Data(d))) => d.clone(),
+        _ => {
+            return Err(FunctionErrorKind::Diagnostic(
+                "set_ip_element: missing <ip> field".to_string(),
```
nichtsfrei, line 1385: usage error/

rust/nasl-interpreter/src/built_in_functions/packet_forgery.rs
```diff
@@ -0,0 +1,2872 @@
...
+    let mut pkt = packet::ipv4::MutableIpv4Packet::new(&mut ip_buf).ok_or_else(|| {
+        FunctionErrorKind::Diagnostic(
+            "No possible to create a packet from buffer".to_string(),
+            Some(NaslValue::Null),
+        )
+    })?;
+    pkt.set_total_length(l as u16);
+    match register.named("update_ip_len") {
+        Some(ContextType::Value(NaslValue::Boolean(l))) if !(*l) => {
+            println!("set the ori again");
```
nichtsfrei, line 1465: don't use println!

rust/nasl-interpreter/src/built_in_functions/packet_forgery.rs
```diff
@@ -0,0 +1,2872 @@
...
+    }
+
+    for udp_datagram in positional.iter() {
+        match udp_datagram {
+            NaslValue::Data(data) => {
+                let ip = match packet::ipv4::Ipv4Packet::new(data) {
+                    Some(ip) => ip,
+                    None => {
+                        return Err(FunctionErrorKind::Diagnostic(
+                            "Usage : dump_udp_packet_options(udp)".to_string(),
```
nichtsfrei, line 1631: usage error

nichtsfrei (dismissed):

nichtsfrei:
Generally you seem to use the wrong error codes. Don't use diagnostic for each case we have different cases:
```
#[derive(Debug, Clone, PartialEq, Eq)]
/// Describtive kind of error that can occur while calling a function
pub enum FunctionErrorKind {
    /// Function called with insufficient arguments
    MissingPositionalArguments {
        /// Expected amount of arguments
        expected: usize,
        /// Actual amount of arguments
        got: usize,
    },
    /// Function called without required named arguments
    MissingArguments(Vec<String>),
    /// Wraps formatting error
    FMTError(std::fmt::Error),
    /// Wraps Infallible
    Infallible(Infallible),
    /// Wraps io::ErrorKind
    IOError(io::ErrorKind),
    /// Function was called with wrong arguments
    WrongArgument(String),
    /// Diagnostic string is informational and the second arg is the return value for the user
    Diagnostic(String, Option<NaslValue>),
    /// Generic error
    GeneralError(GeneralErrorType),
}
```

So for missing arguments use `MissingArguments` for wrong arguments use `WrongArgument` for missing positional arguments you can use `MissingPositionalArguments`. That way you don't have to generate each string manually and it is clearer how to react from an interpreter point of view.

Additionally when you want to print out debug information use the logger trait with the debug function. That way we have more control over it.

jjnicola:
> for missing arguments use `MissingArguments` for wrong arguments use `WrongArgument` for missing positional arguments you can use `MissingPositionalArguments`. That way you don't have to generate each string manually and it is clearer how to react from an interpreter point of view.

The problem with these error types is that they abort the whole script execution, not only the function. Therefore, we added the `Diagnostic` type, which shows the warning, but also gives the possibility to return a `NaslValue` (commonly a Null), then the script continues with the next calls.
The options I find are: 
- Modify the current errors to also return a `NaslValue`, like `Diagnostic`
- Add new errors like `MissingArgumentsNicely` , `WrongArgumentNice`(or something like that).
 

> Additionally when you want to print out debug information use the logger trait with the debug function. That way we have more control over it.
Done.

nichtsfrei:
I think it should be up to the interpreter if it should abort the complete script or not. 

I don't think it should be up to the function call to decide that. This leads to chaos as there are already multiple view points on that. I for an example think we should abort the whole script on an usage error while you disagree.

In any case it makes sense to normalize this behavior. 

The diagnostic option was more meant for corrupted or incorrect data that is not necessarily controlled by the user. 

Even if we decide that we don't want to abort on such cases it make sense to have normalized behaviour (e.g. they all return Null) that is not up to the developer of a function to decide but rather on the interpreting part.

rust/nasl-interpreter/src/built_in_functions/packet_forgery.rs
Kraemii: There is a `MissingArguments` error for such cases

jjnicola:
Depends on #1438

Kraemii (approved):

jjnicola:
> I think it should be up to the interpreter if it should abort the complete script or not.
> 
> I don't think it should be up to the function call to decide that. This leads to chaos as there are already multiple view points on that. I for an example think we should abort the whole script on an usage error while you disagree.
> 
> In any case it makes sense to normalize this behavior.
> 
> The diagnostic option was more meant for corrupted or incorrect data that is not necessarily controlled by the user. For that you also have a couple of things that I also think it should be diagnostic for an example when the buffer is set but incorrect.
> 
> Even if we decide that we don't want to abort on such cases it make sense to have normalized behaviour (e.g. they all return Null) that should not up to the developer of a function to decide but rather on the interpreting part so that it is easier to control and less random so to speak.
> 
> So in short I think it makes more sense to have different errors for different things:
> 
> * a given data buffer is set but incorrect should be diagnostic, as it may not be the fault of the script developer but maybe the returned buffer of a different function is wrong or corrupt. this shouldn't abort the whole script as it may be wanted to be checked later on.
> * a missing argument is a fault of the script developer and should abort the whole script to indicate usage error. The reason for that is that this will fail everytime regardless of an outcome of a different function.

It was decide that functions miss-usage abort the script, while other cases will return Null. Each case must be evaluated separately even for already merged builtin functions . Therefore required changes related to error handling will be done in a another PR.

ArnoStiefvater (approved):

PR #1420, opened 2023-05-11
Fix: allow/deny hosts.

jjnicola:
**What**:
Fix: allow/deny hosts.
Jira SC-837

**Why**:
This PR fix an issue which produces the scan to be interrupted, because the denied hosts are not removed from the hosts list, and the final host count doesn't match with the amount of alive hosts. This ends with an interrupted scan produced by a wrong scan progress calculation.

These fix depends on PR greenbone/gvm-libs#771. However, it is not a breaking change, since a pre-processor directive check for a flag. Without an updated gvm-libs behaves as before and uses the feature (the fix) if enabled.

**How**:
Set a allow/deny hosts for a user and run a scan with a target which includes the allow/deny hosts. 
Also, check with the scanner only settings (openvas.conf) `sys_hosts_deny` and `sys_hosts_allow`

**Checklist**:

- [ ] Tests


src/attack.c
```diff
@@ -1019,6 +1023,55 @@ apply_hosts_excluded (gvm_hosts_t *hosts)
...
+{
+  kb_t kb = NULL;
+  int *sw = systemwide;
+  connect_main_kb (&kb);
+  if (*sw == 0)
+    message_to_client ((kb_t) kb, "Host access denied.", (gchar *) data, NULL,
+                       "ERRMSG");
+  else if (*sw == 1)
+    message_to_client ((kb_t) kb,
+                       "Host access denied (systm-wide restriction).",
```
ArnoStiefvater, line 1038:
```suggestion
                       "Host access denied (system-wide restriction).",
```

src/attack.c
```diff
@@ -1019,6 +1023,55 @@ apply_hosts_excluded (gvm_hosts_t *hosts)
...
+  const char *sys_deny_hosts = prefs_get ("sys_hosts_deny");
+  if (allow_hosts || deny_hosts)
+    {
+      systemwide = 1;
+      removed = gvm_hosts_allowed_only (hosts, sys_deny_hosts, sys_allow_hosts);
+      g_slist_foreach (removed, print_host_access_denied,
+                       (gpointer) &systemwide);
+      g_slist_free_full (removed, g_free);
+    }
+}
```
Kraemii, line 1072:
```suggestion
  const char *sys_allow_hosts = prefs_get ("sys_hosts_allow");
  const char *sys_deny_hosts = prefs_get ("sys_hosts_deny");
  if (sys_allow_hosts || sys_deny_hosts)
    {
      systemwide = 1;
      removed = gvm_hosts_allowed_only (hosts, sys_deny_hosts, sys_allow_hosts);
      g_slist_foreach (removed, print_host_access_denied,
                       (gpointer) &systemwide);
      g_slist_free_full (removed, g_free);
    }
}
```

Kraemii:
With the small change, everything works, as expected

ArnoStiefvater (approved):

Kraemii (approved):

PR #1431, opened 2023-05-30
rs separate builtin

nichtsfrei:
Splits the nasl-builtin-functions from nasl-interpreter into various crates.

This enables us to to 
- create own caching mechanism for specific builtin-functions (e.g. ssh sessions),
- toggle them on compile time, 
- have localized dependencies

with the cost of higher complexity.

The principle of dealing with that is to have a special `nasl-builtin-std` crate in which all builtin functions are defined.

This allows us to just include std and the ContextBuilder to gain access to all defined builtin functions.

[SC-863](https://jira.greenbone.net/browse/SC-863)
[SC-809](https://jira.greenbone.net/browse/SC-809)


rust/nasl-builtin-cryptographic/README.md
```diff
@@ -0,0 +1,107 @@
...
+
+Implements cryptographic functions within NASL.
+
+It is part of the std lib which is proven by the tests.
+
+To use this module you have to initiate Cryptographic and look for the function:
+
+```
+let functions = nasl_builtin_utils::NaslfunctionRegisterBuilder::new()
+    .add(nasl_builtin_cryptographic::Cryptographic)
```
Kraemii, line 11:
```suggestion
    .push_register(nasl_builtin_cryptographic::Cryptographic)
```
Must be changed for all occurrences

rust/nasl-builtin-description/README.md
```diff
@@ -0,0 +1,31 @@
+# Extends nasl-function that don't have network or file capabilities
+
+It is part of the std which is proven in the tests.
+
+To use this module you have to initiate `Description` and look for the function:
+
+```
+let functions = nasl_builtin_utils::NaslfunctionRegisterBuilder::new()
+    .add(nasl_builtin_description::Description)
```
Kraemii, line 9:
```suggestion
    .push_register(nasl_builtin_description::Description)
```

rust/nasl-builtin-std/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    /// Sets the functions available to the nasl script.
+    pub fn functions(mut self, functions: NaslFunctionRegister<K>) -> Self {
+        self.functions = functions;
+        self
+    }
+}
+
+impl<K: AsRef<str>> ContextBuilder<K, KeyDispatcherSet<K>> {
```
jjnicola, line 190: This implementation is already done for `new()` above in the line 151

nichtsfrei: Can you elaborate it a bit? 

The new function in line 151 creates a new ContextBuilder while the build function marked here creates a Context. So those are not the same and do completely different things unless I overlooked something.

jjnicola: yes, build() and new() do different things, but both are implementation for `ContextBuilder` What I mean is that they can be inside the same `impl {}` block. Just for easier reading.

nichtsfrei: ah, got it. You're right thanks.

rust/nasl-builtin-cryptographic/tests/helper/mod.rs
```diff
@@ -0,0 +1,7 @@
+use std::num::ParseIntError;
+pub(crate) fn decode_hex(s: &str) -> Result<Vec<u8>, ParseIntError> {
+    (0..s.len())
+        .step_by(2)
+        .map(|i| u8::from_str_radix(&s[i..i + 2], 16))
+        .collect()
+}
```
Kraemii, line 7: This code is [duplicated](https://github.com/greenbone/openvas-scanner/pull/1431/files#diff-0cb1d2de0b3359b2737a9099ecd00899a89342c6a55b484626f5cd3c0679b390R16) and never used

rust/nasl-builtin-utils/src/error.rs
```diff
@@ -0,0 +1,141 @@
...
+use nasl_syntax::NaslValue;
+use storage::StorageError;
+
+use crate::context::ContextType;
+
+/// Reuses the StorageError definitions as they should fit most cases.
+pub type GeneralErrorType = StorageError;
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+/// Describtive kind of error that can occur while calling a function
```
Kraemii, line 13:
```suggestion
/// Descriptive kind of error that can occur while calling a function
```

rust/nasl-builtin-utils/src/lib.rs
```diff
@@ -0,0 +1,216 @@
...
+pub use error::FunctionErrorKind;
+
+/// The result of a function call.
+pub type NaslResult = Result<nasl_syntax::NaslValue, FunctionErrorKind>;
+
+/// Is a type definition for built-in functions
+///
+/// It is mostly used internally when building a NaslFunctionExecuter.
+/// The register as well as the context are given by the interpreter that wants either a result or
+//Lw
```
Kraemii, line 16: Is this line intended?

rust/nasl-builtin-cryptographic/tests/helper/mod.rs
nichtsfrei: lt's used, regarding the duplication: it's on purpose. As it's only used within tests, I think it's less confusing to duplicate this short functionality than to have a dependency between crypto and string

jjnicola (approved):

Kraemii (approved):

PR #1445, opened 2023-06-28
Add: support for host_details within the osp crate

Kraemii:
The host_details information are part of the osp response for get_scans. In order to get this details, the request must contain `progress='1'`. The information needed are then all part of the `progress` element.
Within the response message note that finished and alive is currently the same value.
SC-850

To test it
1. Set everything up (openvasd, ospd-openvas, notus-scanner)
2. Create a new Scan with the XML below POST `http://127.0.0.1:3000/scans`. Remember to change the credentials to your setup. This config contains the smallest possible scan, that will trigger Notus as well. This is necessary, as the scan will be too fast to get a good test response. I setup a loop back address as `10.0.0.1` to ensure to have a better response (more running hosts) with `sudo ip addr add 10.0.0.1/32 dev lo`.
```xml
{
  "scan_id": "6c591f83-8f7b-452a-8c78-ba35779e682f",
  "target": {
    "hosts": [
      "127.0.0.1",
      "::1",
      "10.0.0.1"
    ],
    "ports": [
      {
        "protocol": "udp",
        "range": [
          {
            "start": 1,
            "end": 3000
          }
        ]
      },
      {
        "protocol": "tcp",
        "range": [
          {
            "start": 1,
            "end": 3000
          }
        ]
      }
    ],
    "credentials": [
      {
        "service": "ssh",
        "port": 22,
        "up": {
          "username": "user",
          "password": "pw"
        }
      }
    ]
  },
  "vts": [
    {
      "oid": "1.3.6.1.4.1.25623.1.0.50282"
    }
  ]
}
```
3. Start the scan with POST `http://127.0.0.1:3000/scans/6c591f83-8f7b-452a-8c78-ba35779e682f`:
```xml
{
  "action": "start"
}
```
4. Send requests until you are satisfied GET `http://127.0.0.1:3000/scans/6c591f83-8f7b-452a-8c78-ba35779e682f/status`. In the end I got:
```xml
{
    "start_time": 1687959628,
    "end_time": 0,
    "status": "running",
    "host_info": {
        "all": 2,
        "excluded": 0,
        "dead": 0,
        "alive": 1,
        "queued": 0,
        "finished": 1,
        "scanning": ["::1", "10.0.0.1"]
    }
}
```

### DIFF
--- a/rust/osp/src/response.rs
+++ b/rust/osp/src/response.rs
@@ -177,9 +177,9 @@ pub enum Response {
     /// <get_scans_response status_text="OK" status="200">
     ///  <scan id="9750f1f8-07aa-49cc-9c31-2f9e469c8f65"
     ///      target="192.168.1.1"
-    ///      end_time="1432000000"
-    ///      progress="100"
-    ///      status="finished"
+    ///      end_time="0"
+    ///      progress="78"
+    ///      status="running"
     ///      start_time="1432000000">
     ///      <results>
     ///         <result host="192.168.1.1"
@@ -192,6 +192,15 @@ pub enum Response {
     ///         An SSH service was detected on the remote host.
     ///         </result>
     ///     </results>
+    ///     <progress>
+    ///         <host name="127.0.0.1">45</host>
+    ///         <host name="10.0.0.160">98</host>
+    ///         <overall>78</overall>
+    ///         <count_alive>2</count_alive>
+    ///         <count_dead>10</count_dead>
+    ///         <count_excluded>3</count_excluded>
+    ///         <count_total>511</count_total>
+    ///      </progress>
     ///  </scan>
     ///</get_scans_response>
     ///```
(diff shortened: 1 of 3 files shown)


rust/osp/src/response.rs
```diff
@@ -512,6 +521,47 @@ pub struct Scan {
...
+/// Information about the scan progress
+#[derive(Clone, Debug, Deserialize, PartialEq)]
+pub struct HostInfo {
+    #[serde(default)]
+    /// Currently scanned hosts
+    pub host: Vec<Host>,
+    /// Overall progress
+    pub overall: ElementU32,
+    /// Number of alive hosts finished
+    /// TODO: Consider divide into alive and finished
```
nichtsfrei, line 538: Nit picking: I think the TODO should not be part of the documentation but a normal comment?

nichtsfrei (approved):

rust/osp/src/response.rs
Kraemii: fixed

jjnicola:
Could you add the example in the PR description as example in `rust/example/simple_scan.json` ?  I think it is a useful example and should be accessible for anyone who wants to try this.

Kraemii:
We have examples in the API specification, not sure if I should also add them into the examples directory

nichtsfrei:
I think the examples in the API are sufficient

nichtsfrei (approved):

jjnicola (approved):

PR #1446, opened 2023-07-04
Rs minor improvements

Kraemii:
**What**:
- In the rust implementation of openvas were many artifacts of sensord left in the documentation, examples and the code, which also leads to some inconsistency between those parts, e.g. the documentation contains a different config path, than the config module. SC-868
- Some of the configs did not work for openvasd (API backend): SC-869
  - openvas log level could only be adjusted via environment variable, now there exists a configuration, also including it in the help message
  - ospd socket was only read from the environment variable
- Some fixes regarding the OSPD socket handling: SC-867
  - A warning is now printed, in case the OSPD socket is not available
  - The path of the socket is included in all warnings and error messages regarding the socket
  - A error message, if openvasd tries to reach OSPD without a valid socket
  - A scan is only set to requested, when the OSPD command worked


rust/examples/openvasd/config.example.toml
```diff
@@ -10,16 +10,19 @@ enable_get_scans = true
...
 
 [ospd.result_check_interval]
 secs = 1
 nanos = 0
 
 [listener]
 address = "127.0.0.1:3000"
+
+[log]
+level = "DEBUG"
```
Kraemii, line 28: Currently the log level cannot be parsed with the config file. I spent some time investigating, with no clue, why it does not work

rust/openvasd/README.md
```diff
@@ -84,7 +78,7 @@ files.
 
 
 If you want to enforce the requirement for clients to provide an x-api-key
-header, make sure to set the key configuration option under [endpoints] to the
+header, make sure to set the key configuration option under `endpoints` to the
```
nichtsfrei, line 81: that `[]` was on purpose to enable links in the rust doc

nichtsfrei (approved):

jjnicola (approved):

PR #1460, opened 2023-07-26
Fix: override an existing scan

jjnicola:
**What**:
Fix: override an existing scan only if it is on Stored status
Jira: SC-893

**Why**:
This patch solves an issue in which a running scan could be overriden if a new scan is post/stored with the same scan ID, producing a reset of the running scan status and results lost
Instead, the API changed and now it is not possible to provide a scan id when creating a scan.
The scan id is automatically generated by openvasd and sent back to the client in the response.

**How**:
Create a scan with an scan_id. I will get an error. Create a scan without an scan_id, the scan is created and you get the ID in the response.
Without the patch, you are able to create a scan passing a scan ID. If you repetitively create the scan with the same ID, the already existing scan is override, independent of the status, losing results and any other information. 

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/models/src/status.rs
+++ b/rust/models/src/status.rs
@@ -43,6 +43,10 @@ impl Status {
     pub fn is_done(&self) -> bool {
         !self.is_running()
     }
+
+    pub fn is_stored(&self) -> bool {
+        self.status.is_stored()
+    }
 }
 
 /// Enum of the possible phases of a scan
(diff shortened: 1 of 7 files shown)


Kraemii:
IMO a scan should never be able to be overwritten, but always deleted first. In case a stored scan can be overwritten, there is always the chance of deleting data, that is still needed and can also lead to clashes in a multi user scenario, as the user has the chance to set an own scan ID. When we enforce the deletion first it is also more explicit.

rust/models/src/status.rs
```diff
@@ -43,6 +43,10 @@ impl Status {
     pub fn is_done(&self) -> bool {
         !self.is_running()
     }
+
+    pub fn is_stored(&self) -> bool {
```
nichtsfrei, line 47: why introducing this method? The name of the enum is already Stored.

nichtsfrei (changes requested):

Kraemii (approved):

nichtsfrei (approved):

PR #1461, opened 2023-07-26
Change: improve handling of sensitive data

jjnicola:
**What**:
Change: improve handling of sensitive data
Jira: SC-877

**Why**:
With this patch, sentitive data, like credential, are hidden when GET scan_id, and delete from the scan config once the scan was started.

**How**:
Run a scan with, the scan must contain credentials, of course:
`curl --verbose  --insecure --cert $CLIENT_CERT --cert-type PEM --key $CLIENT_KEY --request POST  https://127.0.0.1:3000/scans -H "X-API-KEY: changeme" -d  '@/tmp/simple_scan.json' | jq`

And get the scan before and after staring it. Credentials shouldn't be shown after the path is applied.
`curl --verbose  --insecure --cert $CLIENT_CERT --cert-type PEM --key $CLIENT_KEY --request GET https://127.0.0.1:3000/scans/<THE SCAN ID> -H "X-API-KEY: changeme" | jq .`

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/models/src/credential.rs
+++ b/rust/models/src/credential.rs
@@ -61,35 +61,46 @@ pub enum CredentialType {
     /// User/password credentials.
     UP {
         /// The username for authentication.
+        #[cfg_attr(feature = "serde_support", serde(serialize_with = "crate::censor"))]
         username: String,
         /// The password for authentication.
+        #[cfg_attr(feature = "serde_support", serde(serialize_with = "crate::censor"))]
         password: String,
     },
     #[cfg_attr(feature = "serde_support", serde(rename = "usk"))]
     /// User/ssh-key credentials.
     USK {
         /// The username for authentication.
+        #[cfg_attr(feature = "serde_support", serde(serialize_with = "crate::censor"))]
         username: String,
         /// The password for authentication.
+        #[cfg_attr(feature = "serde_support", serde(serialize_with = "crate::censor"))]
         password: String,
         #[cfg_attr(feature = "serde_support", serde(rename = "private"))]
         /// The private key for authentication.
+        #[cfg_attr(feature = "serde_support", serde(serialize_with = "crate::censor"))]
         private_key: String,
     },
     #[cfg_attr(feature = "serde_support", serde(rename = "snmp"))]
     /// SNMP credentials.
     SNMP {
         /// The SNMP username.
+        #[cfg_attr(feature = "serde_support", serde(serialize_with = "crate::censor"))]
         username: String,
         /// The SNMP password.
+        #[cfg_attr(feature = "serde_support", serde(serialize_with = "crate::censor"))]
         password: String,
         /// The SNMP community string.
+        #[cfg_attr(feature = "serde_support", serde(serialize_with = "crate::censor"))]
         community: String,
         /// The SNMP authentication algorithm.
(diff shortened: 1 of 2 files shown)


rust/openvasd/src/controller/entry.rs
```diff
@@ -187,6 +187,7 @@ where
                                                 .as_ref()
                                                 .unwrap_or(&"".to_string())
                                         );
+                                        progress.clean_sensitive_data();
```
nichtsfrei, line 190: I'm not sure if removing credentials immediately after starting is sustainable, as the scanner may need the data later on. Although I am aware that currently the data is duplicated by ospd the moment that changes this call can be easily overlooked and debugging missing credentials on a large scan is very cumbersome.

Instead I would promote encryption using a stream cipher like cha20 for credentials so that it is encrypted in memory and decrypting on a per need basis. Additionally we should remove credentials when a scan is finished or if we want to allow reusage of scans later on delete.

rust/openvasd/src/controller/entry.rs
```diff
@@ -250,7 +251,13 @@ where
             response_blocking(move || {
                 let scans = ctx.scans.read()?;
                 match scans.get(&id) {
-                    Some(prgrs) => Ok(ctx.response.ok(&prgrs.scan)),
+                    Some(prgrs) => {
+                        // Clone the scan to remove sensitive data before
+                        // sending it back into the response
+                        let mut prgrs_aux = prgrs.clone();
+                        prgrs_aux.clean_sensitive_data();
```
nichtsfrei, line 258: Do we consider everything in a credential confidential? With that a customer will never be aware if a scan even contains credentials which will be very confusing if a customer uses this endpoint to verify the creation of a scan. I think a replacement with `*` on passwords etc will be less confusing than just deletion.

rust/openvasd/src/controller/entry.rs
jjnicola: I removed this line. So the credentials are still in memory.

rust/openvasd/src/controller/entry.rs
Kraemii: Maybe it would be even better to allow the user to set a custom ID for each credential. In this case we could just send an ID back and the user can lookup in his local setup, which credentials he actually used. To prevent sending back the credentials, we then can just disable the serialization of them with a feature flag with serde directly.

Kraemii (commented):
I had an alternative idea, by implementing this directly into serde. This can be done, by adding a line like:
```rust
#[serde(serialize_with = "crate::censor")]
```
To the object, that should be hidden and a function like:
```rust
fn censor<S, T>(_: &T,serializer: S) -> std::result::Result<S::Ok, S::Error> where S: Serializer {
  serializer.serialize_str("***")
}
```
In this case we would be able to censor any data, by just adding the serde flag to it

rust/models/src/credential.rs
```diff
@@ -50,6 +52,10 @@ impl AsRef<str> for Service {
     }
 }
 
+fn censor<S, T>(_: &T,serializer: S) -> std::result::Result<S::Ok, S::Error> where S: Serializer {
+  serializer.serialize_str("***")
+}
+
```
Kraemii, line 58: I would move it to lib.rs, so it can be used in the whole crate

Kraemii (approved):

PR #1463, opened 2023-07-28
Fix automagic dependency on netsnmp that is not required.

MocioF:
When building openvas-scanner Cmake searches for netsnmp and links to it but the dependency is not defined as required: https://github.com/greenbone/openvas-scanner/blob/d1b7268a05eb81b5deab0febe7ac472a730815c0/nasl/CMakeLists.txt#L73-L80)

In source-based distributions (like gentoo) this can create an issue on system update. Reading here you can have a more detailed description of the issue: https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies#Why_automagic_dependencies_are_wrong

Please add a BUILD_WITH_NETSNMP variable in openvas-scanner/CMakeLists.txt and openvas-scanner/nasl/CMakeLists.txt as suggested here: https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies#CMake

**What**:

**Why**:

Avoid automagic dependency

**How**:

**Checklist**:

- [ ] Tests N/A
- [X] PR merge commit message adjusted


nasl/CMakeLists.txt
```diff
@@ -70,14 +70,20 @@ if (NOT GPGME_FOUND)
     OUTPUT_STRIP_TRAILING_WHITESPACE)
 endif (NOT GPGME_FOUND)
 
-message (STATUS "Looking for netsnmp...")
-find_library (SNMP netsnmp)
-message (STATUS "Looking for netsnmp... ${SNMP}")
-if (SNMP)
+if (DEFINED BUILD_WITH_NETSNMP)
```
Flowdalic, line 73: Isn't `BUILD_WITH_NETSNMP`, due to the added block in `CmakeLists.txt`  always defined? If so, then I think you can get rid of one layer of `if` conditions :)

nichtsfrei (approved):

PR #1464, opened 2023-08-02
Poc rust c nasl

Kraemii:
This PR introduces a concept for implementing c based nasl libraries within rust.
This works by a build script, which uses the rust cc library to compile the c code and link it to the created rust library.
    
This POC also contains a solution integrating the GMAC algorithm, as there is currently no support for it in the RustCrypto project.

SC-839
SC-883


rust/nasl-c-lib/src/nasl_aes_mac_gcm.rs
```diff
@@ -0,0 +1,35 @@
...
+        out: *mut u8,
+    ) -> c_int;
+    fn get_aes_mac_gcm_len() -> c_uint;
+}
+
+pub fn aes_mac_gcm(data: &[u8], key: &[u8], iv: &[u8]) -> Result<Vec<u8>, i32> {
+    unsafe {
+        let mut ret = data.to_vec();
+        // let mut ret: Vec<u8> = Vec::with_capacity(get_aes_mac_gcm_len() as usize);
+        let out = ret.as_mut_ptr();
```
Kraemii, line 20: For some reason the mut pointer of ret is NULL (within the c code), when it does not contain any values

rust/nasl-c-lib/build.rs
```diff
@@ -0,0 +1,10 @@
+fn main() {
+    // Tell Cargo that if the given file changes, to rerun this build script.
+
+    cc::Build::new()
+        .file("c/nasl_aes_mac_gcm/nasl_aes_mac_gcm.c")
+        .compile("crypt");
+
+    println!("cargo:rerun-if-changed=src/nasl_aes_mac_gcm.c");
+    println!("cargo:rustc-link-lib=gcrypt");
```
nichtsfrei, line 9: maybe it would make sense to statically link gcrypt? so that we don't have to install gcrypt on a client.

PR #1467, opened 2023-08-04
Add: openvasd: encryption of credentials and results

nichtsfrei:
To allow easier handling of sensitive data the direct usage of a hash
map is replaced with a storage interface and field encryption is added.

This is done by introducing interfaces:
- to get the progress of a scan
- store scans
- append results

instead of using a hashmap of a progress struct. This allows us to
exchange the storage from inmemory to something like a relational
database.

Additionally a crypt interface with a chacha20 implementation is
introduces to enable encryption of sensitive fields like passwords or
results.

[SC-900](https://jira.greenbone.net/browse/SC-900)

### DIFF
--- a/rust/openvasd/src/controller/entry.rs
+++ b/rust/openvasd/src/controller/entry.rs
@@ -88,12 +88,16 @@ where
 {
     use KnownPaths::*;
     // on head requests we just return an empty response without checking the api key
-
+    tracing::trace!(
+        "{} {}:{:?}",
+        req.method(),
+        req.uri().path(),
+        req.uri().query()
+    );
     if req.method() == Method::HEAD {
         return Ok(ctx.response.empty(hyper::StatusCode::OK));
     }
     let kp = KnownPaths::from_path(req.uri().path());
-    tracing::debug!("{} {}:{:?}", req.method(), kp, req.uri().query());
     if let Some(key) = ctx.api_key.as_ref() {
         match req.headers().get("x-api-key") {
             Some(v) if v == key => {}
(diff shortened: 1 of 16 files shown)


rust/openvasd/src/controller/entry.rs
```diff
@@ -134,165 +107,128 @@ where
...
     }
+    tracing::debug!("{} {}", req.method(), kp);
 
     match (req.method(), kp) {
-        (&Method::HEAD, _) => Ok(ctx.response.empty(hyper::StatusCode::OK)),
         (&Method::POST, Scans(None)) => {
             match crate::request::json_request::<models::Scan>(&ctx.response, req).await {
                 Ok(mut scan) => {
-                    response_blocking(move || {
+                    if scan.scan_id.is_none() {
```
jjnicola, line 116: Why is this `if` statement introduced again? In a previous PR we removed and even changed the API to not accept an scan_id for creating a scan

rust/openvasd/src/controller/entry.rs
```diff
@@ -68,59 +68,32 @@ impl Display for KnownPaths {
...
 {
     use KnownPaths::*;
+    // on head requests we just return an empty response without checking the api key
+
+    if req.method() == Method::HEAD {
+        return Ok(ctx.response.empty(hyper::StatusCode::OK));
+    }
     let kp = KnownPaths::from_path(req.uri().path());
-    tracing::debug!("{} {}", req.method(), kp);
+    tracing::debug!("{} {}:{:?}", req.method(), kp, req.uri().query());
```
Kraemii, line 96: I would suggest to put the logging before the handling of the HEAD request

rust/openvasd/src/controller/entry.rs
```diff
@@ -134,165 +107,128 @@ where
             }
         }
     }
+    tracing::debug!("{} {}", req.method(), kp);
```
Kraemii, line 110: What is the second log for? These information are already contained in the log before

rust/openvasd/src/controller/entry.rs
Kraemii, line 116: The scan ID should always be generated

jjnicola (changes requested):

rust/openvasd/src/controller/entry.rs
```diff
@@ -134,165 +107,128 @@ where
...
+                        use models::Phase::*;
+                        let expected = &[Stored, Stopped, Failed, Succeeded];
+                        Ok(ctx.response.not_accepted(&status.status, expected))
+                    } else {
+                        match ctx.scanner.start_scan(scan).await {
+                            Ok(_) => {
+                                status.status = models::Phase::Requested;
+                                match ctx.db.update_status(&id, status).await {
+                                    Ok(_) => Ok(ctx.response.no_content()),
                                     Err(e) => Ok(ctx.response.internal_server_error(&e)),
```
Kraemii, line 145: I think we also have to stop the scan at this point, as
```rust
ctx.scanner.start_scan(scan).await
```
was successfully executed

rust/openvasd/src/controller/entry.rs
nichtsfrei: rebase error, going to remove it.

jjnicola (approved):

PR #1474, opened 2023-08-23
Add: read timeout setting to the ospd

jjnicola:
**What**:
Add: read timeout setting to the ospd
Jira: SC-906

**Why**:
If for some reason the ospd socket does not answer, now there is the option to set a timeout. This is the maximum time in seconds that openvasd waits for a response from ospd-openvas.

**How**:

**Checklist**:

- [ ] Tests


rust/openvasd/src/scan.rs
```diff
@@ -56,7 +58,14 @@ impl<T> From<PoisonError<T>> for Error {
...
     /// Creates a new instance of OSPDWrapper
     pub fn new(socket: PathBuf) -> Self {
-        Self { socket }
+        Self {
+            socket,
+            r_timeout: None,
+        }
+    }
+
+    pub fn set_read_timeout(&mut self, secs: Duration) {
```
nichtsfrei, line 67: why not just add it in `new`? As far as I can see the only way to change the timeout is in the configuration, environment variable and via argument all of those are just read within start period.

From that understanding I don't understand why doing it via a setter rather than on constructing.

rust/openvasd/src/main.rs
```diff
@@ -24,7 +24,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     if !config.ospd.socket.exists() {
         tracing::warn!("OSPD socket {} does not exist. Some commands will not work until the socket is created!", config.ospd.socket.display());
     }
-    let scanner = scan::OSPDWrapper::new(config.ospd.socket.clone());
+    let mut scanner = scan::OSPDWrapper::new(config.ospd.socket.clone());
+    if let Some(rtimeout) = config.ospd.read_timeout {
```
nichtsfrei, line 28: I think it would be simpler to read when this is done within the new method by just passing the option into new.

rust/osp/src/connection.rs
```diff
@@ -15,9 +15,15 @@ use crate::{
 };
 
 /// Sends a command to the unix socket and returns the response
-pub fn send_command<T: AsRef<Path>>(address: T, cmd: ScanCommand) -> Result<Response, Error> {
+pub fn send_command<T: AsRef<Path>>(address: T, r_timeout: Option<Duration>, cmd: ScanCommand) -> Result<Response, Error> {
     let mut socket = UnixStream::connect(address)?;
     let cmd = cmd.try_to_xml()?;
+    // Timeout can not be zero. If not set, it waits for ever
```
nichtsfrei, line 21: That comment confuses me. If the timeout can not be zero but it is just set when it is not zero than it could be zero as it is not overridden?

nichtsfrei (approved):

PR #1480, opened 2023-08-30
Add: Smoketest for the scanner API

jjnicola:
**What**:
Add smoketest.
Currently, runs two test:
- test 1: connection with the server (request HEAD)
- test 2: run a scan (optionally with mtls activated)

**Why**:

**How**:

**Checklist**:

- [ ] Tests


rust/models/src/host_info.rs
```diff
@@ -24,8 +24,8 @@ pub struct HostInfo {
     pub finished: u32,
     #[cfg_attr(
         feature = "serde_support",
-        serde(skip_serializing_if = "Vec::is_empty")
+        serde(skip_serializing_if = "Option::is_none")
     )]
     /// IPs of hosts, that are currently scanned.
-    pub scanning: Vec<String>,
+    pub scanning: Option<Vec<String>>,
```
nichtsfrei, line 30: why do you need `Option<Vec<...>>`? Over just leave it empty?

rust/smoketest/src/tests/test1.rs
```diff
@@ -0,0 +1,18 @@
+/// Check the server sending the HEAD request
+pub async fn check_server(cli: &reqwest::Client, server: &String) -> bool {
```
nichtsfrei, line 2: is there a reason why you return true or false over let it panic and use `assert_*` macros?

rust/smoketest/src/main.rs
```diff
@@ -0,0 +1,32 @@
...
+
+use std::process::ExitCode;
+
+use crate::tests::*;
+use clap::Parser;
+use config::Args;
+use smoketest::new_client;
+
+#[tokio::main]
+async fn main() -> ExitCode{
```
nichtsfrei, line 12: Generally I'm wondering why you wrote it as an program instead of reusing `rusts test ecosystem'? When using the test runner we don't have to setup the reporting by 'foot' via log messages and so on.

rust/models/src/host_info.rs
jjnicola: this as given some errors before the in file storage was merged, because serde deserializer.

nichtsfrei (approved):

rust/models/src/host_info.rs
nichtsfrei: is that still the case?

nichtsfrei (changes requested):
Unittests are failing

rust/models/src/host_info.rs
jjnicola: yes it is. Smoketest fails if when getting the status and there is no scanning member

ArnoStiefvater (approved):

PR #1509, opened 2023-10-10
Fix: fix an issue which produces an interrupted scan, originated on a wrong count of excluded hosts.

jjnicola:
**What**:
Fix: fix an issue which produces an interrupted scan, originated on a wrong count of excluded hosts.
Although there is no breaking changes, this fix depends on greenbone/gvm-libs#797

Jira: SC-929

**Why**:
Since excluded hosts are handled in OSPD, we need to keep the total host count which includes the hosts which will be excluded later

**How**:
* Create a target with many alive hosts, someone with dns name
* Exclude a host which doesn't have a name
* Allow simultaneous scanning via multiple IPs: Yes
* Port List All IANA assigned TCP
* Alive Test Scan Config Default
* No Credentials
* Reverse Lookup Only : *Yes*
* Reverse Lookup Unify : *no*
* Create a task that is using the target and the scan config Host Discovery
* Starting this task should interrupt the task at some percentage.

without the patch, the scan is interrupted. With the patch the scan finishes successfully. 
Try other targets and configurations.

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/attack.c
+++ b/src/attack.c
@@ -1083,9 +1083,44 @@ apply_hosts_preferences_ordering (gvm_hosts_t *hosts)
     g_debug ("hosts_ordering: Sequential.");
 }
 
-static void
+static int
 apply_hosts_reverse_lookup_preferences (gvm_hosts_t *hosts)
 {
+#ifdef FEATURE_REVERSE_LOOKUP_EXCLUDED
+  const char *exclude_hosts = prefs_get ("exclude_hosts");
+  int hosts_excluded = 0;
+
+  if (prefs_get_bool ("reverse_lookup_unify"))
+    {
+      gvm_hosts_t *excluded;
+
+      excluded = gvm_hosts_reverse_lookup_unify_excluded (hosts);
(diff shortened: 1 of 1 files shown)


src/attack.c
```diff
@@ -1083,9 +1083,48 @@ apply_hosts_preferences_ordering (gvm_hosts_t *hosts)
...
 
-static void
+static int
 apply_hosts_reverse_lookup_preferences (gvm_hosts_t *hosts)
 {
+  int hosts_excluded = 0;
+
+#ifdef FEATURE_REVERSE_LOOKUP_EXCLUDED
+
+  const char *exclude_hosts = prefs_get ("exclude_hosts");
```
Kraemii, line 1093: #ifdef FEATURE_REVERSE_LOOKUP_EXCLUDED
  ```suggestion
  int hosts_excluded = 0;
  const char *exclude_hosts = prefs_get ("exclude_hosts");
```
Without it, I get
```
attack.c:1089:7: error: unused variable ‘hosts_excluded’
```

Kraemii (approved):

PR #1517, opened 2023-10-27
Fix: infisto storage, create results file on scan init

Kraemii:
In case the storage for a new scan was initialized, the file for the results was missing Now creating a file for the results with an empty Vec

SC-928


Kraemii:
Closed in favor of https://github.com/greenbone/openvas-scanner/pull/1518

PR #1518, opened 2023-10-27
Fix: panic when trying to append empty results

nichtsfrei:
Under the condition that there is a status change
without any kind results infisto returned an error.

This changes the behavior of append-all to ignore
calls without any data.

### DIFF
--- a/rust/openvasd/src/controller/mod.rs
+++ b/rust/openvasd/src/controller/mod.rs
@@ -277,7 +277,6 @@ mod tests {
             count: Arc::new(RwLock::new(0)),
         };
         let ns = std::time::Duration::from_nanos(10);
-        // let storage = file::encrypted("/tmp/aha", "key").unwrap();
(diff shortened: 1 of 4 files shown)


rust/openvasd/src/controller/mod.rs
```diff
@@ -266,18 +269,19 @@ mod tests {
...
-            resp.reverse();
+            let resp = serde_json::from_slice::<Vec<models::Result>>(&resp).unwrap();
             resp
         }
         let scan: models::Scan = models::Scan::default();
         let scanner = FakeScanner {
             count: Arc::new(RwLock::new(0)),
         };
         let ns = std::time::Duration::from_nanos(10);
+        // let storage = file::encrypted("/tmp/aha", "key").unwrap();
```
Kraemii, line 280:
```suggestion
```

Kraemii (approved):

PR #1527, opened 2023-11-27
Fix: nasl-cli/openvasd: possible stackoverflow on End::Continue

nichtsfrei:
When a NASL script gets executed a lot of parens that includes
SyntaxError then nasl-cli may get stuck in an recursive loop that
overflows the stack therefore it was reported with the severity
`VSS:4.0/AV:L/AC:L/AT:N/PR:N/UI:N/VC:N/VI:L/VA:H/SC:N/SI:L/SA:H`.

To fix it Lexer rembers it's current depth and when it get higher than
the defined MAX_DEPTH it returns an SyntaxError of kind MaxRecursion.

If it is based on an previous error then the iterator implementation of
Lexer simulates an EoF to prevent multiple false positive errors to be
reported.

SC-963


jjnicola (approved):
LGTM

PR #1557, opened 2024-01-17
Adjust host details

Kraemii:
**What**:
Currently the host Details were handled as name value pairs, but they actually contain an internal struct with name, value and source information. This PR adjusts the handling of host details, as well as its documentation including an example.

SC-930

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/osp/src/response.rs
+++ b/rust/osp/src/response.rs
@@ -344,36 +344,25 @@ impl From<&ScanResult> for models::ResultType {
             ("HOST_START", ResultType::Log) => models::ResultType::HostStart,
             ("HOST_END", ResultType::Log) => models::ResultType::HostEnd,
             ("DEADHOST", ResultType::Log) => models::ResultType::DeadHost,
-            (_, ResultType::Log) =>
-            // is either set on nasl-script as proto or on openvas when
-            // a host is dead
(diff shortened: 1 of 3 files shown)


rust/osp/src/response.rs
```diff
@@ -344,36 +344,25 @@ impl From<&ScanResult> for models::ResultType {
...
+struct HostDetail {
+    detail: Vec<models::Detail>,
 }
 
-#[derive(Deserialize, Debug)]
-struct HostDetail {
-    detail: Vec<Detail>,
+impl HostDetail {
+    pub fn extract(&self) -> Option<models::Detail> {
+        self.detail.get(0).cloned()
```
nichtsfrei, line 364: should be `first` instead of `get(0)`.

nichtsfrei: Also wouldn't a Option<&models::Detail> be sufficient as it is stored on the HostDetail anyway?

Kraemii: `HostDetail` will only live within the function it is used in, the details on the other hand are used for the generated result, so we need to clone it either here or after the `extract()` call

jjnicola (approved):

PR #1558, opened 2024-01-17
Add: nasl-cli notus update subcommand

jjnicola:
**What**:
Load up Notus Advisories into redis.
It performs the signature check, the hashsum check and the upload.

Signature check is optional. It must be enabled with the command line option but also the environment variable to the gnupg keyring must be set.

Usage:
`GPGHOME=/path/to/.gnupg nasl-cli notus update --path <path-to-the-advisories> --signature-check`

Jira: SC-997

**Why**:

**How**:

**Checklist**:

- [ ] Tests


rust/models/src/advisories.rs
```diff
@@ -0,0 +1,240 @@
...
+            filename: data.filename.to_string(),
+            refs,
+            family: data.product_data.family.to_owned(),
+            name: data.adv.title.to_owned(),
+            category: "3".to_string(),
+        }
+    }
+}
+
+//impl Vulnerability {
```
nichtsfrei, line 199: can this be deleted?

rust/nasl-cli/src/notusupdate/mod.rs
```diff
@@ -0,0 +1,69 @@
...
+                }
+            };
+
+            let signature_check = args
+                .get_one::<bool>("signature-check")
+                .cloned()
+                .unwrap_or(false);
+
+            let dispatcher =
+                redis_storage::NvtDispatcher::as_dispatcher(&redis, NOTUSUPDATE_SELECTOR)
```
nichtsfrei, line 59: maybe the NvtDispatcher of redis_storage should be renamed?

rust/nasl-cli/src/notusupdate/update.rs~
```diff
@@ -0,0 +1,59 @@
+// SPDX-FileCopyrightText: 2023 Greenbone AG
```
nichtsfrei, line 1: seems like a swap file?

rust/notus/src/update/#mod.rs#
```diff
@@ -0,0 +1,69 @@
+// SPDX-FileCopyrightText: 2023 Greenbone AG
```
nichtsfrei, line 1: seeme like a swap file?

rust/models/src/advisories.rs
jjnicola: ups.. that was my try of serializer for vulnerabilities.

rust/nasl-cli/src/notusupdate/mod.rs
jjnicola: I thought about it and forgot to do it.

rust/notus/src/update/#mod.rs#
jjnicola: ups. Also, clean up other files.

rust/nasl-cli/src/notusupdate/update.rs~
jjnicola: ups

jjnicola:
thanks for the review @nichtsfrei

nichtsfrei:
I have been thinking about the separation between the 'notus' and 'feed' commands. Although it is technically correct to separate them, I'm questioning whether users actually need this differentiation or if they would rather 'verify' and 'update' them in one command?

nichtsfrei:
> I have been thinking about the separation between the 'notus' and 'feed' commands. Although it is technically correct to separate them, I'm questioning whether users actually need this differentiation or if they would rather 'verify' and 'update' them in one command?

maybe something like `nasl-cli feed update` -> notus + nasl, `nasl-cli feed --notus update` -> only notus and the same for nasl?

nichtsfrei (approved):

ArnoStiefvater (approved):

PR #1562, opened 2024-01-31
Add: get_vts path for getting vts from redis cache

jjnicola:
**What**:
Add: get_vts path for getting vts from redis cache
Jira: SC-998

**Why**:

**How**:

**Checklist**:

- [ ] Tests


nichtsfrei (changes requested):
As you have multiple collects, starting at VtHelper. Meaning redis streams, we collect, then we iteratore and collect again, this is done 3 times.

- ListRetriever::retrieve_keys - oids
- ListRetriever::retrieve_keys - notus
- GetVts - for each found key we call retrieve_single_nvt. again in a blocking fashion

This will allocate about ~40MB of memory per request for the whole stream.

Therefore I think it is best to use a Iterator and stream the response rather than collecting on each step.

Addiotionally since we already have an alternative approach of getting oids I think we need to either remove the previous one or use it. 

Having to maintain both ways is unnecessary effort.

nichtsfrei (approved):

nichtsfrei:
Decided to merge and now change it later.

Kraemii (approved):

PR #1563, opened 2024-01-31
Change: update hyper to version 1

nichtsfrei:
This changes the underlying version of hyper to version 1.

The run methods is moved to controller. This hypothetically allows us to
run a complete server within tests.

When TLS is enabled, which should be the case for all production
environments, then it will use http2. On http it uses http1 so that it
is more convenient to use with curl.

The stream handling of results and oids is changed to channel and chunks
as it is easier to understand and requires less code then our previous
Stream approach.

The handling of TLS and insecure mode are now more separated, although
this results in wet code with a few duplication it is believed to be
easier to change when there is a problem within one mode but not the
other.

The client identifier handling is also changed, rather then using
wrapper upon wrapper we use a global arc in run and reset per request.


jjnicola (dismissed):

rust/openvasd/src/controller/mod.rs
```diff
@@ -48,27 +59,112 @@ pub enum ClientIdentifier {
...
 impl<T> Scanner for T where T: ScanStarter + ScanStopper + ScanDeleter + ScanResultFetcher {}
 
+fn retrieve_and_reset(id: Arc<RwLock<ClientIdentifier>>) -> ClientIdentifier {
+    // get client information
+    let mut ci = id.write().unwrap();
+    let cci = ci.clone();
+    // reset client information
+    *ci = ClientIdentifier::Unknown;
+    drop(ci);
+    cci
```
Kraemii, line 72: why overwrite `ci` and drop it directly afterwards manually? Isn't it drop regardless after the end of the function?

nichtsfrei: the handler call may be slow. to remove the lock I drop it immediately so that the next request doesn't have to wait until the handling of the previous request is done.

nichtsfrei: ah nevermind, wrong function. yes, you're right I just drop out of habit.

Kraemii (approved):

PR #1564, opened 2024-02-01
 Change: Ci: cleaning up and restructure gh actions

nichtsfrei:
Changes the Release Trigger Process

To initiate a release, navigate to `Actions -> CI` in the GitHub repository, and
click on `Run workflow`. Choose from the following options:
- `major`: For a major release with incompatible changes.
- `minor`: For a minor release introducing new features.
- `patch`: For a patch release focusing on bug fixes and minor improvements.
- `no_release`: To run the pipeline without releasing, updating the edge image.

Jobs Overview

The CI pipeline incorporates multiple jobs, each with a specific function in the
development lifecycle.

1. Initialization (`init`)

If the initialization fails it will prevent furter execution of `build`.

- **Purpose**: Sets the release type based on the input or event that triggered
  the workflow.
- **Workflow File**: `init.yaml`

2. Unit Tests (`unittests`)
- **Purpose**: Executes unit tests to validate code changes.
- **Workflow File**: `tests.yml`

If the unit tests fails it will prevent furter execution of `build`.

3. Build (`build`)
- **Purpose**: Compiles and builds the project, preparing it for testing and
  deployment.
- **Dependencies**: Requires successful completion of `unittests`.
- **Workflow File**: `build.yml`

If the build fails it will prevent furter execution of `functional`.

4. Linting (`linting`)
- **Purpose**: Ensures code quality and consistency through linting.
- **Workflow File**: `linting.yml`

If linting fails it will not prevent execution of the other steps, as it may be
that newer versions of the used tooling finds new linting issues that are not
affecting the binary as much.

5. Functional Testing (`functional`)
- **Purpose**: Conducts functional tests on the build.
- **Dependencies**: Needs a successful `build`.
- **Workflow File**: `functional.yaml`

If the functional tests fail it will prevent furter execution of
`containerization`.

6. Containerization
- **Purpose**: Packages the build into Docker containers.
- **Jobs**:
  - **Container**: Uses `push-container.yml`.
  - **Container Testing**: Uses `push-container-testing.yml`.
  - **Container Oldstable**: Uses `push-container-oldstable.yml`.
- **Dependencies**: Depends on `build`, `init`, and `functional`.

If the `containerization` fails the smoketests cannot be executed.

7. Smoke Tests (`smoketests`)
- **Purpose**: Conducts tests on helm chart based on the previously pushed
  docker image.
- **Conditions**: Excluded during pull request events.
- **Dependencies**: Relies on `container`, `build`, and `init`.
- **Workflow File**: `smoketest.yaml`

If the smoketests fail the helm chart will not be updated and releases be
prevented.

8. Helm Chart Deployment (`helm`)
- **Purpose**: Deploys Helm chart, assuming `IMAGE_REGISTRY` is configured.
- **Conditions**: Triggered if `IMAGE_REGISTRY` is set.
- **Dependencies**: Depends on `smoketests`, `container`, `build`, and `init`.
- **Workflow File**: `push-helm-chart.yml`

9. Release (`release`)
- **Purpose**: Handles the release process for different version types.
- **Conditions**: Activated based on the release type set in `init`.
- **Dependencies**: Requires `smoketests`, `container`, `build`, and `init`.
- **Workflow File**: `release.yml`

The CI workflow employs GitHub secrets for secure authentication and interaction
with external services such as DockerHub.

- **DOCKERHUB_USERNAME**: DockerHub username.
- **DOCKERHUB_TOKEN**: Token for DockerHub with write access to the registry.
- **GREENBONE_BOT_TOKEN**: Token for Helm chart registry and GitHub repository
  operations.
- **GREENBONE_BOT**: Username for git commits.
- **GREENBONE_BOT_MAIL**: Email address for git commits.


.github/workflows/init.yaml
```diff
@@ -0,0 +1,149 @@
...
+      latest_version: ${{ steps.version.outputs.latest_version }}
+      release_kind:  ${{ steps.version.outputs.release_kind }}
+      release_ref:  ${{ steps.version.outputs.release_ref }}
+      project:  ${{ steps.version.outputs.project}}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: "set IS_VERSION_TAG"
```
bjoernricks, line 51: Would be nice to use https://github.com/greenbone/actions/tree/main/is-latest-tag here instead

.github/workflows/release.yml
```diff
@@ -34,115 +44,30 @@ on:
...
-      RELEASE_REF: ${{needs.calculate_version.outputs.release_ref}}
-      LATEST_VERSION: ${{needs.calculate_version.outputs.latest_version}}
-      NEW_VERSION: ${{needs.calculate_version.outputs.new_version}}
-      PROJECT: ${{needs.calculate_version.outputs.project}}
+      RELEASE_KIND: ${{inputs.release_kind}}
+      RELEASE_REF: ${{inputs.release_ref}}
+      LATEST_VERSION: ${{inputs.latest_version}}
+      NEW_VERSION: ${{inputs.new_version}}
+      PROJECT: ${{inputs.project}}
     steps:
```
bjoernricks, line 56: I really think you should use the https://github.com/greenbone/actions/tree/main/release action here

nichtsfrei: Last time I checked it didn't match all our needs, hence we have that self written approach. We may change it, however based on the massive restructure I would prefer to do that later.

.github/workflows/init.yaml
nichtsfrei: i fail to understand why that would be beneficial.

.github/workflows/release.yml
nichtsfrei: For an example it doesn't seem to create a branch on a major version change. It doesn't support Rust. 

The capability of creating a branch on major version is crucial for us and part of the existing release process. Given the complexity of github action and the inability to validate it on a local machine I really, really think this is not the right PR for requesting release process changes.

bjoernricks: If you really really want to maintain your specific workflows on your own you can of course do whatever you want.

nichtsfrei: we already do that, that's not changed within this PR: https://github.com/greenbone/openvas-scanner/blob/main/.github/workflows/release.yml

I would love to switch, but again, we can't as that action doesn't cover everything we need. I didn't say you're wrong I did say this is not the time to bring it up.

y0urself: I could see you doing - instead of maintaining your own:
a) add the desired features to the action (contribution)
b) ask DevOps to do a)

Would that solve your problem?

nichtsfrei: yes, b. would be, but again: not the right PR to request that. Can we please go down the official route instead of discussing on a PR that DOES NOT change the release process.

.github/workflows/README.md
```diff
@@ -0,0 +1,92 @@
...
+- `patch`: For a patch release focusing on bug fixes and minor improvements.
+- `no_release`: To run the pipeline without releasing, updating the edge image.
+
+## Jobs Overview
+
+The CI pipeline incorporates multiple jobs, each with a specific function in the development lifecycle.
+
+### 1. Initialization (`init`)
+
+If the initialization fails it will prevent furter execution of `build`.
```
Kraemii, line 19:
```suggestion
If the initialization fails it will prevent further execution of `build`.
```
This typo occurs multiple times

.github/workflows/ddependabot.yml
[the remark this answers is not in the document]
nichtsfrei: can be deleted

Kraemii (dismissed):

Kraemii (approved):

PR #1565, opened 2024-02-05
Add: scan scheduler

Kraemii:
Add: scan scheduler

Changes context to contain only a scheduler and not any scanner or
storage directly. This to ensure that each command goes through the
scheduler so that it can keep track if a feed update is running or a
scan is queued.

Additionally the background processes call the scheduler to synchronize
scans and feed. A feed can only be updated when no scans are running and
a scan only started when no feed update is running.

This means essentially that the Scheduler within scheduling plays a core
role in openvasd shifting away from direct scanner access.

Since a scheduler uses only standard interfaces of models::scanner each
scanner needs to keep track by themselves depending on their technical
solution. In a nutshell this means that each scanner must be implemented
within it's crate:
- osp has a scanner implementation
- openvasctl has a scanner implementation

SC-999


rust/openvasctl/src/ctl.rs
```diff
@@ -1,83 +1,198 @@
...
+}
+
+pub struct Scheduler {
+    scans: Scans,
     sudo: bool,
+    config: Config,
 }
 
-impl OpenvasController {
+impl Scheduler {
```
nichtsfrei, line 82: maybe it makese sense to move the scheduler to plugin scheduler and move everything openvas specific to openvasctl struct? That way we could utilize the scheduler with the rust impl.

rust/openvasctl/src/ctl.rs
```diff
@@ -1,83 +1,198 @@
...
     /// Create a new OpenvasController
-    pub fn new() -> Result<Self, OpenvasError> {
+    pub fn new(config: Option<Config>) -> Result<Self, OpenvasError> {
         if !cmd::check() {
             return Err(OpenvasError::MissingExec);
         }
 
         Ok(Self {
             scans: Default::default(),
             sudo: check_sudo(),
```
nichtsfrei, line 91: I think sudo is just needed when executing openvas which is as far as I understand implemented in openvasctl?

rust/openvasctl/src/ctl.rs
```diff
@@ -1,83 +1,198 @@
...
+            // TODO: Handle interrupted scans
+
+            let scans = self.scans.clone();
+
+            loop {
+                match rx.try_recv() {
+                    Ok(command) => match command {
+                        ScanCommand::StartScan(scan) => {
+                            if scan.scan_id.is_none() {
+                                // TODO: Error handling
```
nichtsfrei, line 114:
```
tracing::warn!("scan without id, skipping");
```
should be sufficient?

rust/openvasctl/src/ctl.rs
```diff
@@ -1,83 +1,198 @@
...
+        }
     }
+}
 
-    /// Remove a scan from the init phase
-    fn remove_init(&mut self, id: &String) -> bool {
-        self.scans.lock().unwrap().init.remove(id).is_some()
-    }
+pub trait ScanStarter {
+    fn start_scan(scans: Scans, sudo: bool) -> Result<(), OpenvasError>;
```
nichtsfrei, line 164: I would rewrite this interface to bind to &self so that a scanstarter can utilize implementation specific information by themselves. This way we could implement multiple starter without having to pass context information per call.

rust/openvasctl/src/ctl.rs
```diff
@@ -1,83 +1,198 @@
...
-        if container.init.remove(&id).is_none() {
-            return Ok(false);
-        }
-        let openvas =
-            cmd::start(&id, self.sudo, None).map_err(|_| OpenvasError::UnableToRunExec)?;
-        container.running.insert(id, openvas);
-        Ok(true)
-    }
+pub trait ScanStopper {
+    fn stop_scan(scans: Scans, id: &str, sudo: bool) -> Result<(), OpenvasError>;
```
nichtsfrei, line 168: Same as start

rust/openvasctl/src/ctl.rs
Kraemii: In this case we would have to put the struct into a mutex and lock the whole struct, when calling start or stop scan, which could block for a long time, even though it does not run any operations on shared data.
instead of
```rust
task::spawn(async move { S::start_scan(scans, self.sudo) });
```
we would have to do something like
```rust
let controller = controller.clone();
task::spawn(async move { controller.lock().unwrap().start_scan() });
```
which will block the controller, until start_scan() is finished, which would block all calls to start_scan() and stop_scan().

jjnicola (approved):

nichtsfrei (approved):

PR #1567, opened 2024-02-12
Add: Preference handler for openvasctl

jjnicola:
**What**:
 Add preference handler module

This module gets a scan config and prepare the data to be stored in redis as openvas-scanner expected.
The data is stored in redis as well. For that, a new module openvas_redis was added as well, which adds an implements
the RedisHelper struct.

Jira: SC-1001

**Why**:
Deprecation of ospd-openvas

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/models/src/credential.rs
+++ b/rust/models/src/credential.rs
@@ -17,7 +17,6 @@ pub struct Credential {
     #[cfg_attr(feature = "serde_support", serde(flatten))]
     /// Type of the credential to get access. Different services support different types.
     pub credential_type: CredentialType,
-
 }
 
 impl Credential {
--- a/rust/models/src/credential.rs
+++ b/rust/models/src/credential.rs
@@ -50,6 +50,7 @@ impl Default for Credential {
             credential_type: CredentialType::UP {
                 username: "root".to_string(),
                 password: "".to_string(),
+                privilege_credential: None,
             },
         }
     }
@@ -75,16 +76,12 @@ pub enum Service {
     #[cfg_attr(feature = "serde_support", serde(rename = "snmp"))]
     /// SNMP, supports [SNMP](CredentialType::SNMP)
     SNMP,
-    #[cfg_attr(feature = "serde_support", serde(rename = "privilege_ssh"))]
-    /// Privilege SSH, supports [SSH](CredentialType::UP)
-    PSSH,
 }
 
 impl AsRef<str> for Service {
     fn as_ref(&self) -> &str {
         match self {
             Service::SSH => "ssh",
-            Service::PSSH => "privilege_ssh",
             Service::SMB => "smb",
             Service::ESXi => "esxi",
             Service::SNMP => "snmp",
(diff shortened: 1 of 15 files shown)


rust/openvasctl/src/pref_handler.rs
```diff
@@ -0,0 +1,626 @@
...
+            "internal/123-456/scanprefs",
+            "1.3.6.1.4.1.25623.1.0.103591:1:entry:SSH login name:|||user"
+        ));
+
+        assert!(prefh.prepare_ports_for_openvas().is_ok());
+        assert!(prefh
+            .redis_connector
+            .item_exists("internal/123-456/scanprefs", "PORTS|||tcp:22-25,80,"));
+
+        println!("{:?}", prefh.redis_connector.data);
```
nichtsfrei, line 624: This println seems odd. Was it meant to be rewritten as an assert?

nichtsfrei (dismissed):

rust/openvasctl/src/pref_handler.rs
jjnicola: removed with [6e68d1f](https://github.com/greenbone/openvas-scanner/pull/1567/commits/6e68d1fffc1a18665d4041055ffcf45ce3fa49c9)

nichtsfrei (dismissed):

rust/openvasctl/src/pref_handler.rs
```diff
@@ -0,0 +1,624 @@
...
+const OID_ESXI_AUTH: &str = "1.3.6.1.4.1.25623.1.0.105058";
+const OID_SNMP_AUTH: &str = "1.3.6.1.4.1.25623.1.0.105076";
+const OID_PING_HOST: &str = "1.3.6.1.4.1.25623.1.0.100315";
+
+const BOREAS_ALIVE_TEST: &str = "ALIVE_TEST";
+const BOREAS_ALIVE_TEST_PORTS: &str = "ALIVE_TEST_PORTS";
+const ALIVE_TEST_SCAN_CONFIG_DEFAULT: u8 = 0x00;
+
+fn int_to_bool(value: u8) -> bool {
+    if value == 0 {
```
nichtsfrei, line 24: this can be simplified to 
```
value == 0
```
instead of that if.

Kraemii: I think the whole method is not necessary, as `value == 0` can be used directly where it is needed

rust/models/src/credential.rs
```diff
@@ -75,12 +75,16 @@ pub enum Service {
...
+    #[cfg_attr(feature = "serde_support", serde(rename = "privilege_ssh"))]
+    /// Privilege SSH, supports [SSH](CredentialType::UP)
+    PSSH,
 }
 
 impl AsRef<str> for Service {
     fn as_ref(&self) -> &str {
         match self {
             Service::SSH => "ssh",
+            Service::PSSH => "privilege_ssh",
```
Kraemii, line 87: Add privilege ssh to the ssh implementation, as it can only be used with it.
Also add the privilege ssh to the openapi specification.

nichtsfrei (approved):

Kraemii (approved):

PR #1569, opened 2024-02-15
Change: set Scan ID of Scan Model to non-optional

Kraemii:
As the scan ID was set by the server and not by the client, it was set to Optional. This lead to many unnecessary checks, even though it was not possible, that this value is None.

SC-1021

### DIFF
--- a/rust/models/src/scan.rs
+++ b/rust/models/src/scan.rs
@@ -14,7 +14,7 @@ use super::{scanner_preference::ScannerPreference, target::Target, vt::VT};
 pub struct Scan {
     #[cfg_attr(feature = "serde_support", serde(default))]
     /// Unique ID of a scan
-    pub scan_id: Option<String>,
+    pub scan_id: String,
     /// Information about the target to scan
     pub target: Target,
     #[cfg_attr(feature = "serde_support", serde(default))]
--- a/rust/openvasd/src/controller/entry.rs
+++ b/rust/openvasd/src/controller/entry.rs
@@ -257,14 +257,14 @@ where
                     match crate::request::json_request::<models::Scan, _>(&ctx.response, req).await
                     {
                         Ok(mut scan) => {
-                            if scan.scan_id.is_some() {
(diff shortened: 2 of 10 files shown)


rust/models/src/scan.rs
```diff
@@ -13,9 +13,9 @@ use super::{scanner_preference::ScannerPreference, target::Target, vt::VT};
 )]
 #[cfg_attr(feature = "bincode_support", derive(bincode::Encode, bincode::Decode))]
 pub struct Scan {
-    #[cfg_attr(feature = "serde_support", serde(default))]
+    #[cfg_attr(feature = "serde_support", serde(skip_deserializing, default = "uuid"))]
```
nichtsfrei, line 16: I don't think we can do it that way as we use deserializing and serializing when using infisto. This means that everytime we deserialize we have a new scan_id.

Kraemii: Yes, you are right, I forgot about it...

nichtsfrei: maybe it is better to use `#[serde(default)]` and override the scan_id on post scan inside entry?

rust/openvasd/src/controller/entry.rs
```diff
@@ -256,15 +256,9 @@ where
...
                     {
-                        Ok(mut scan) => {
-                            if scan.scan_id.is_some() {
-                                return Ok(ctx
-                                    .response
-                                    .bad_request("field scan_id is not allowed to be set."));
-                            }
-                            let id = uuid::Uuid::new_v4().to_string();
+                        Ok(scan) => {
+                            let id = scan.scan_id.clone();
```
nichtsfrei, line 260: We could still warn the user about an invalid contract when doing:
```
                            if !scan.scan_id.is_empty() {
                                return Ok(ctx
                                    .response
                                    .bad_request("field scan_id is not allowed to be set."));
                            }
 ```

jjnicola (approved):

PR #1572, opened 2024-02-16
Change: make async the methods for preference handling.

jjnicola:
**What**:
Jira: SC-1026
Also, improve the init of RedisHelper, since at this point the nvti cache and a main kb for the scan should have been created.

Usage example. (Needs a running redis instance with an up-to-date nvticache)

``` rust
use std::sync::{Arc, Mutex};

use models::{AliveTestMethods, Credential, PortRange, Port, Scan, Parameter}; use openvas::{pref_handler::PreferenceHandler, openvas_redis::RedisHelper}; use redis_storage::{NameSpaceSelector, RedisCtx};

async fn main() -> Result<(), std::fmt::Error> {

    // Create an scan config
    let mut scan = Scan::default();
    scan.scan_id = Some("123-456".to_string());
    scan.target.alive_test_methods = vec![AliveTestMethods::Icmp, AliveTestMethods::TcpSyn];
    scan.target.credentials = vec![Credential {
        service: models::Service::SSH,
        port: Some(22),
        credential_type: models::CredentialType::UP {
            username: "user".to_string(),
            password: "pass".to_string(),
            privilege_credential: None,
        },
    }];
    scan.vts.push(models::VT {
        oid: "1.3.6.1.4.1.25623.1.0.112771".to_string(),
        parameters: vec![Parameter {
            id: 1,
            value: "llala".to_string()

        }]
    });
    scan.target.excluded_hosts = vec!["127.0.0.1".to_string()];
    scan.target.hosts = vec!["127.0.0.2".to_string()];
    scan.target.ports = vec![Port {
        protocol: Some(models::Protocol::TCP),
        range: vec![
            PortRange {
                start: 22,
                end: Some(25),
            },
            PortRange {
                start: 80,
                end: None,
            },
        ],
    }];

    let redis_url = "unix:///run/redis-openvas/redis.sock";
    // In this example, a fix db is taken,but the next free can be taken, using the name space selector NameSpaceSelector::Free
    let mut rctx = RedisCtx::open(redis_url, &[NameSpaceSelector::Fix(6)]).unwrap();
    rctx.delete_namespace().unwrap();
    // Create a redis connection to the nvticache, and a connection to the mainkb().
    let cache = RedisCtx::open(redis_url, &[NameSpaceSelector::Key("nvticache")]).unwrap();

    // New redis helper with access to the main kb for storing the preferences, and access to the nvticache
    // for getting info to build preferences.
    let rc = RedisHelper::<RedisCtx>::init(Arc::new(Mutex::new(cache)),Arc::new(Mutex::new(rctx))).unwrap();

    let mut p = PreferenceHandler::new(scan, rc);
    Ok(p.prepare_preferences_for_openvas().await.expect("aaa"))

}
```

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/openvasctl/src/openvas_redis.rs
+++ b/rust/openvasctl/src/openvas_redis.rs
@@ -4,7 +4,7 @@
 
 use redis_storage::{
     dberror::{DbError, RedisStorageResult},
-    NameSpaceSelector, RedisCtx, RedisGetNvt, RedisWrapper,
+    RedisCtx, RedisGetNvt, RedisWrapper,
 };
 use std::{
     collections::HashMap,
@@ -28,16 +28,12 @@ where
     /// Initialize a RedisHelper struct with the connection to access the NVT cache
     /// and a empty task knowledge base to store the scan configuration to be sent to openvas.
     pub fn init(
-        redis_url: &str,
-        selector: &[NameSpaceSelector],
+        nvti_cache: Arc<Mutex<RedisCtx>>,
+        kb_cache: Arc<Mutex<RedisCtx>>,
     ) -> RedisStorageResult<RedisHelper<RedisCtx>> {
-        let mut rctx = RedisCtx::open(redis_url, selector)?;
-        rctx.delete_namespace()?;
-        let cache = RedisCtx::open(redis_url, selector)?;
-
         Ok(RedisHelper::<RedisCtx> {
-            cache: Arc::new(Mutex::new(cache)),
-            task_kb: Arc::new(Mutex::new(rctx)),
+            cache: nvti_cache,
+            task_kb: kb_cache,
         })
     }
 }
--- a/rust/openvasctl/src/openvas_redis.rs
+++ b/rust/openvasctl/src/openvas_redis.rs
@@ -27,14 +27,14 @@ where
 {
     /// Initialize a RedisHelper struct with the connection to access the NVT cache
     /// and a empty task knowledge base to store the scan configuration to be sent to openvas.
-    pub fn init(
+    pub fn new(
         nvti_cache: Arc<Mutex<RedisCtx>>,
         kb_cache: Arc<Mutex<RedisCtx>>,
-    ) -> RedisStorageResult<RedisHelper<RedisCtx>> {
-        Ok(RedisHelper::<RedisCtx> {
+    ) -> RedisHelper<RedisCtx> {
+        RedisHelper::<RedisCtx> {
             cache: nvti_cache,
             task_kb: kb_cache,
-        })
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


nichtsfrei (approved):

rust/openvasctl/src/pref_handler.rs
```diff
@@ -103,11 +103,12 @@ where
                 vts_list.push(vt.oid.clone());
 
                 // prepare vt preferences
-                for pref in vt.parameters.iter() {
+                for pref in vt.parameters.clone() {
```
nichtsfrei, line 106: what's the reason to clone it?

rust/openvasctl/src/openvas_redis.rs
```diff
@@ -28,16 +28,12 @@ where
     /// Initialize a RedisHelper struct with the connection to access the NVT cache
     /// and a empty task knowledge base to store the scan configuration to be sent to openvas.
     pub fn init(
-        redis_url: &str,
-        selector: &[NameSpaceSelector],
+        nvti_cache: Arc<Mutex<RedisCtx>>,
+        kb_cache: Arc<Mutex<RedisCtx>>,
     ) -> RedisStorageResult<RedisHelper<RedisCtx>> {
```
nichtsfrei, line 33: why result? the connection is already initialized and cannot fail.

rust/openvasctl/src/openvas_redis.rs
```diff
@@ -28,16 +28,12 @@ where
     /// Initialize a RedisHelper struct with the connection to access the NVT cache
     /// and a empty task knowledge base to store the scan configuration to be sent to openvas.
     pub fn init(
```
nichtsfrei, line 30: Since the initialization is done previously I think it would be better to rename it to new and remove the result.

nichtsfrei (changes requested):
Some minor things.

rust/openvasctl/src/pref_handler.rs
```diff
@@ -76,7 +76,7 @@ where
         Ok(())
     }
 
-    fn prepare_scan_id_for_openvas(&mut self) -> RedisStorageResult<()> {
+    async fn prepare_scan_id_for_openvas(&mut self) -> RedisStorageResult<()> {
         self.redis_connector.push_kb_item(
```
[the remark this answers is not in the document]
jjnicola: `push_kb_item` doesn't take so long. Its quite fast, since an item can be a vector of items. It would be different if it has to store items one by one and do many calls to the push function. In any case, we can check this later and do it given the case.

nichtsfrei (approved):

Kraemii (approved):

PR #1576, opened 2024-02-23
Add: privilege credential for SSH authenticated scans

jjnicola:
**What**:
Add: privilege credential for SSH authenticated scans

After a successful authentication, the scanner can require privilege access for some test. In this case the privilege credential are used.
This means that the privilege username and password only make sense for SSH service and only if the regular username and password are given

Also, add a small fix for a parameter name in the openvasd/config.rs

Jira: SC-1025

**Why**:
was changed for UP credential and added for USK credential.

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/models/src/credential.rs
+++ b/rust/models/src/credential.rs
@@ -22,7 +22,7 @@ impl Credential {
     /// Maps the password of the credential using the given closure.
     pub fn map_password<F, E>(self, f: F) -> Result<Self, E>
     where
-        F: FnOnce(String) -> Result<String, E>,
+        F: Fn(String) -> Result<String, E>,
     {
         Ok(Credential {
             service: self.service,
@@ -49,10 +49,22 @@ impl Default for Credential {
             credential_type: CredentialType::UP {
                 username: "root".to_string(),
                 password: "".to_string(),
-                privilege_credential: None,
+                privilege: None,
             },
         }
     }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(
+    feature = "serde_support",
+    derive(serde::Serialize, serde::Deserialize)
+)]
+pub struct PrivilegeInformation {
+    #[cfg_attr(feature = "serde_support", serde(rename = "privilege_username"))]
+    pub username: String,
+    #[cfg_attr(feature = "serde_support", serde(rename = "privilege_password"))]
+    pub password: String,
 }
 
 /// Enum of available services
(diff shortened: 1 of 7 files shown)


rust/models/src/credential.rs
```diff
@@ -114,6 +121,14 @@ pub enum CredentialType {
         #[cfg_attr(feature = "serde_support", serde(rename = "private"))]
         /// The private key for authentication.
         private_key: String,
+        /// privilege username only use for SSH service
+        #[serde(skip_serializing_if = "Option::is_none")]
```
nichtsfrei, line 125: This is a personal preference so it can be seen as nitpicking and can be decided by you. But I think that keeping the information as null within the response is better than hiding it as it makes it more obvious that it is an option without having to lookup the api when playing around with openvasd. I am aware that some prefer it, and I have to admit I was one of them, in this case I think it is not necessary to spare 50 bytes.

nichtsfrei: TL;DR: I  would remove `#[serde(skip_serializing_if = "Option::is_none")]`

nichtsfrei (approved):

nichtsfrei:
Looks good, I like that the privileged information are duplicated instead of introducing a new credential type because it is making it easier to use. 

An alternative approach could be to introduce a new struct:
```
struct PrivilegeInformation {
   username: String,
   password: String,
}
``` 

and include that as an option into the enum variants an option while annotating it with serde flatten:

```
...
  #[serde(flatten)]
  privelege: Option<PrivilegeInformation>,
...
```

that way we could ensure that username and password are written in the same  way.

However since we just have two cases, it can be seen as optional :).

rust/models/src/credential.rs
jjnicola: I thought about this as well, but it is not only for SSH service. The problem is that it will be shown for all service credential, even for those which does not allow/support privilege credentials. 
I decide to hide it because it can be confusing an unset privilege credential for e.g. SMB service, which doesn't support it

nichtsfrei: that's a good point.

rust/models/src/credential.rs
```diff
@@ -49,12 +49,22 @@ impl Default for Credential {
...
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(
+    feature = "serde_support",
+    derive(serde::Serialize, serde::Deserialize)
+)]
+pub struct PrivilegeInformation {
+    pub privilege_username: String,
```
nichtsfrei, line 64: when you use `#[serde(rename = "privilege_username")]` you can rename the field to username and password respectively as it is clear within rust that those are privilege from the struct name.

rust/models/src/credential.rs
```diff
@@ -114,6 +127,11 @@ pub enum CredentialType {
         #[cfg_attr(feature = "serde_support", serde(rename = "private"))]
         /// The private key for authentication.
         private_key: String,
+        /// privilege credential only use for SSH service
+        #[serde(skip_serializing_if = "Option::is_none")]
+        #[serde(default)]
```
nichtsfrei, line 132: I think we have to use `#[cfg_attr(feature = "serde_support"` everywhere when we want to use serde. Otherwise compilation fails when serde is not available. Alternatively we mark serde as required in the cargo and drop the compiler declaration for cfg_attr.

Luckily you can combine them:
```
#[cfg_attr(feature = "serde_support", serde(default, flatten, skip_serializing_if = "Option::is_none"))]
```

should be valid.

nichtsfrei (approved):

PR #1581, opened 2024-03-01
Fix: memory check handling in scheduler 

nichtsfrei:
Fix: memory check handling in scheduler

The memory check does start a scan although there is insufficient
memory. This is fixed by moving the check before iterating through the
queued scans.

 Change: sanitize openvasd config

Instead of using wrapper the configuration is using scanner. This
expresses the meaning better than wrapper.

### DIFF
--- a/rust/openvasd/src/scheduling.rs
+++ b/rust/openvasd/src/scheduling.rs
@@ -182,7 +182,7 @@ where
                 let available_memory = sys.available_memory();
                 if available_memory < min_free_memory {
                     tracing::debug!(%min_free_memory, %available_memory, "insufficient memory to start a scan.");
-                    continue;
+                    break;
                 }
             }
             if let Some(scan_id) = queued.pop() {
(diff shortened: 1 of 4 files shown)


rust/openvasd/src/scheduling.rs
```diff
@@ -176,14 +177,15 @@ where
...
-                    let available_memory = sys.available_memory();
-                    if available_memory < min_free_memory {
-                        tracing::debug!(%min_free_memory, %available_memory, %scan_id, "insufficient memory to start scan.");
-                    }
+            sys.refresh_memory();
+            if let Some(min_free_memory) = self.config.min_free_mem {
+                let available_memory = sys.available_memory();
+                if available_memory < min_free_memory {
+                    tracing::debug!(%min_free_memory, %available_memory, "insufficient memory to start a scan.");
+                    continue;
```
Kraemii, line 185:
```suggestion
                    break;
```
Maybe a break would be better, as there won't be enough memory available to start any scan. Also a scan does not use memory immediately, so as long as there is enough memory to start a single scan, all scans in queue will be started at once, which can quickly lead to out of memory issues

nichtsfrei: Regarding your second case: wouldn't be `min_free_memory` the wrong terminology to begin with and shouldn't we extend it to more values like:
- memory_to_keep_free_as_a_buffer
- median_memory_consumption_per_scan
- ...
?

Kraemii (approved):

PR #1585, opened 2024-03-05
Implement Scanner trait for Openvas scanner

jjnicola:
**What**:
Implement traits start, stop and collect results for openvas scanner
Jira: SC-1035

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/models/src/status.rs
+++ b/rust/models/src/status.rs
@@ -68,13 +68,10 @@ impl FromStr for Phase {
     fn from_str(status: &str) -> Result<Phase, ()> {
         match status {
             "requested" => Ok(Phase::Requested),
-            "new" => Ok(Phase::Requested),
             "running" => Ok(Phase::Running),
-            "ready" => Ok(Phase::Running),
             "stopped" => Ok(Phase::Stopped),
             "failed" => Ok(Phase::Failed),
             "succeeded" => Ok(Phase::Succeeded),
-            "finished" => Ok(Phase::Succeeded),
             "stored" => Ok(Phase::Stored),
             _ => Err(()),
         }
--- a/rust/models/src/status.rs
+++ b/rust/models/src/status.rs
@@ -2,7 +2,7 @@
 //
 // SPDX-License-Identifier: GPL-2.0-or-later
 
-use std::fmt::Display;
+use std::{fmt::Display, str::FromStr};
 
 use super::host_info::HostInfo;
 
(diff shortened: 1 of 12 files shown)


rust/models/src/status.rs
```diff
@@ -62,6 +62,25 @@ impl Phase {
...
+        match status {
+            "requested" => Ok(Phase::Requested),
+            "new" => Ok(Phase::Requested),
+            "running" => Ok(Phase::Running),
+            "ready" => Ok(Phase::Running),
+            "stopped" => Ok(Phase::Stopped),
+            "failed" => Ok(Phase::Failed),
+            "succeeded" => Ok(Phase::Succeeded),
+            "finished" => Ok(Phase::Succeeded),
+            "stored" => Ok(Phase::Stored),
```
nichtsfrei, line 78: I think stored is incorrect here, as we use it to indicate that a scan has been stored but not yet started by the user within the /scan endpoint. In this case the user has started the scan so it should either stay on requested or be Succeeded?

jjnicola: Not sure if I understand your point. I use the FromStr trait for converting every possible string in our stack into a valid Phase status. Actually, some strings like `new` (openvas status) and `requested` (openvasd status) corresponds to the same Phase `Requested`.
In this case, `stored` is the an openvasd phase status already defined and will never be set from the openvas side.

nichtsfrei: ah, right, I assumed it is openvas specific as the string cases are but it's an overall fromstr.

jjnicola: To avoid confusion, I moved the openvas specif phases to openvas.rs with https://github.com/greenbone/openvas-scanner/pull/1585/commits/ba24c69898fcab5fd5763816533995570388ba91

nichtsfrei:
The feed update verification step fails on feed-verifier/src/main.rs#174 with no results found. Did the storage of the meta data change?

jjnicola:
> The feed update verification step fails on feed-verifier/src/main.rs#174 with no results found. Did the storage of the meta data change?

Fixed with https://github.com/greenbone/openvas-scanner/pull/1585/commits/b337ef0c939585a6df290a90e95d47f5406f0e18

nichtsfrei (approved):

rust/models/src/target.rs
```diff
@@ -44,9 +44,9 @@ pub struct Target {
...
-    Icmp = 0x01,
-    TcpSyn = 0x02,
-    TcpAck = 0x04,
-    Arp = 0x08,
-    ConsiderAlive = 0x16,
+    TcpAck = 0x01,
+    Icmp = 0x02,
+    Arp = 0x04,
+    ConsiderAlive = 0x08,
+    TcpSyn = 0x10,
```
Kraemii, line 51:
```suggestion
    TcpSyn = 0x16,
```
I guess a typo?

jjnicola: No, it is hexa, which is 16 decimal

Kraemii (approved):

PR #1586, opened 2024-03-06
Fix: Set scan status to failed, if scan cannot be started

Kraemii:
In cases the scan cannot be started, the scan is removed from the queue, but the scan status was still requested. This fixes the issue, by setting the scan status to failed.

### DIFF
--- a/rust/openvasd/src/scheduling.rs
+++ b/rust/openvasd/src/scheduling.rs
@@ -199,7 +199,18 @@ where
                         queued.push(scan_id);
                     }
                     Err(e) => {
-                        tracing::warn!(%scan_id, %e, "unable to start, removing from queue. Verify that scan using the API");
+                        tracing::warn!(%scan_id, %e, "unable to start, removing from queue and set status to failed. Verify that scan using the API");
+                        self.db
+                            .update_status(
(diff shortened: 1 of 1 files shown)


rust/openvasd/src/scheduling.rs
```diff
@@ -199,7 +199,17 @@ where
                         queued.push(scan_id);
                     }
                     Err(e) => {
-                        tracing::warn!(%scan_id, %e, "unable to start, removing from queue. Verify that scan using the API");
+                        tracing::warn!(%scan_id, %e, "unable to start, removing from queue and set status to failed. Verify that scan using the API");
+                        self.update_status(
+                            &scan_id,
+                            models::Status {
+                                start_time: None,
+                                end_time: None,
```
Kraemii, line 207: Not sure, if it is better to set the start and end time, as the scan technically did not start at all

nichtsfrei (approved):

rust/openvasd/src/scheduling.rs
nichtsfrei: I think it's fine to leave start and end time empty.

jjnicola (approved):

PR #1598, opened 2024-03-11
Adjust example config

Kraemii:
Dependent on https://github.com/greenbone/openvas-scanner/pull/1590


jjnicola (dismissed):

rust/examples/openvasd/config.example.toml
```diff
@@ -80,9 +86,16 @@ path = "/var/lib/openvasd/storage"
 # Sets the key used to ecrypt the storage data. It is recommended to set it via the `STORAGE_KEY` environment variable.
 #key = "changeme"
 
+[scheduler]
```
ArnoStiefvater, line 89: Put in the default as currently used as example.

ArnoStiefvater (approved):

PR #1612, opened 2024-04-03
Get scanner details

Kraemii:
Rename scanner preferences to scan preferences
Add get scanner details from OSP to the new scanner API as get scan preferences
Implement the new endpoint into Openvasd

The new endpoint is `/scans/preferences` and a simple get command should get something like:
```json
[{
    "id": "auto_enable_dependencies",
    "name": "Automatic Enable Dependencies",
    "default": true,
    "description": "OpenVAS plugins use the result of each other to execute their job. For instance, a plugin which logs into the remote SMB registry will need the results of the plugin which finds the SMB name of the remote host and the results of the plugin which attempts to log into the remote host. If you want to only select a subset of the plugins available, tracking the dependencies can quickly become tiresome. If you set this option to 'yes', openvas will automatically enable the plugins that are depended on."
}, {
    "id": "cgi_path",
    "name": "CGI Path",
    "default": "/cgi-bin:/scripts",
    "description": "By default, openvas looks for default CGIs in /cgi-bin and /scripts. You may change these to something else to reflect the policy of your site. The syntax of this option is the same as the shell $PATH variable: path1:path2:..."
}, ...]
```
Jira: SC-1040


Kraemii (commented):
~~This PR contains an API change for the rename of `scanner_preference` to `scan_preference` and should only be included for the next Major Release~~

Kraemii:
Moved API changes to a new PR https://github.com/greenbone/openvas-scanner/pull/1615

jjnicola (dismissed):

nichtsfrei (approved):

jjnicola (approved):

PR #1615, opened 2024-04-03
Adjust scanner preferences

Kraemii:
This PR contains two changes to the API:
1. Rename scanner_preferences to scan_preferences within the Scan json
2. Change the contents of scan_preferences to a json dict instead of an array of id-value pairs

This PR should be merged for the next major release.

Jira: SC-1040


ArnoStiefvater:
Will be superseded by #2221

PR #1617, opened 2024-04-08
Fix: TLS support

jjnicola:
**What**:
Fix: TLS support

**Why**:
The server didn't start if the client certs were missing. With this fix, it is possible to start the server in TLS or mTLS modes

**How**:
Start openvasd without client cert options, without a valid directory or an empty directory.

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/openvasd/src/tls.rs
+++ b/rust/openvasd/src/tls.rs
@@ -137,16 +137,13 @@ pub fn config_to_tls_paths(
     //   .expect("When a tls.key is set, a certificate must be set as well");
     let client_certs = match &config.tls.client_certs {
         Some(x) => x,
-        None => return Ok(Some((key_path.to_path_buf(), certs.to_path_buf(), vec![]))),
+        None => {
+            return Ok(Some((key_path.to_path_buf(), certs.to_path_buf(), vec![])));
+        }
     };
     let client_certs = match std::fs::read_dir(client_certs) {
         Ok(x) => x,
-        Err(e) => {
-            tracing::warn!(
-                client_certs = ?client_certs,
-                error = ?e,
-                "unable to load client certificates, continuing without them"
-            );
+        Err(_) => {
             return Ok(Some((key_path.to_path_buf(), certs.to_path_buf(), vec![])));
         }
     };
@@ -175,6 +172,7 @@ pub fn tls_config(config: &crate::config::Config) -> Result<Option<TlsData>, Err
         Some(x) => x,
         None => return Ok(None),
     };
+
     let mut roots = RootCertStore::empty();
     for root in clients.iter().flat_map(load_certs).flatten() {
         roots.add(root)?;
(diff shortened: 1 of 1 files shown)


nichtsfrei (dismissed):

Kraemii (changes requested):
Not sure about this fix, as the root cause why there are no client certs is not stated, even if it is clear, that either no path is configured, the given path is empty or the given path does not exist. Also not giving a client certs path could be wanted, when just TLS is sufficient

jjnicola:
> Not sure about this fix, as the root cause why there are no client certs is not stated, even if it is clear, that either no path is configured, the given path is empty or the given path does not exist. Also not giving a client certs path could be wanted, when just TLS is sufficient

Fix: Improved the warnings and clarify why is running in TLS mode if client certs are not found.

rust/openvasd/src/tls.rs
```diff
@@ -175,24 +172,60 @@ pub fn tls_config(config: &crate::config::Config) -> Result<Option<TlsData>, Err
...
-        .with_single_cert(certs, key)?;
+        config.alpn_protocols = vec![b"h2".to_vec()];
 
-    config.alpn_protocols = vec![b"h2".to_vec()];
+        Ok(Some((client_identifier, config, !clients.is_empty())))
+    } else {
+        match &config.tls.client_certs {
+            Some(clicerts) => {
+                match std::fs::read_dir(clicerts) {
+                    Ok(x) => {
```
Kraemii, line 199:
```suggestion
                    Ok(_) => {
```

Kraemii: cargo clippy is complaining

rust/openvasd/src/tls.rs
```diff
@@ -175,24 +172,60 @@ pub fn tls_config(config: &crate::config::Config) -> Result<Option<TlsData>, Err
...
-        .with_single_cert(certs, key)?;
+        config.alpn_protocols = vec![b"h2".to_vec()];
 
-    config.alpn_protocols = vec![b"h2".to_vec()];
+        Ok(Some((client_identifier, config, !clients.is_empty())))
+    } else {
+        match &config.tls.client_certs {
+            Some(clicerts) => {
+                match std::fs::read_dir(clicerts) {
+                    Ok(_x) => {
```
Kraemii, line 199:
```suggestion
                    Ok(_) => {
```
I would rather remove the x completely, or is there a reason for this?

jjnicola: it was the clippy's suggestion.

Kraemii (approved):

rust/openvasd/src/tls.rs
```diff
@@ -175,24 +172,60 @@ pub fn tls_config(config: &crate::config::Config) -> Result<Option<TlsData>, Err
...
+                        tracing::warn!(
+                            client_certs = ?clicerts,
+                            error = ?e,
+                            "Unable to load client certificates, continuing without them in TLS mode"
+                        );
+                    }
+                };
+            }
+            None => {
+                tracing::warn!(
```
nichtsfrei, line 215: shouldn't that be either a info or even debug as it is configured that way?

rust/openvasd/src/tls.rs
```diff
@@ -175,24 +172,60 @@ pub fn tls_config(config: &crate::config::Config) -> Result<Option<TlsData>, Err
...
+                    "Client certificates option not set, continuing without them in TLS mode."
```
nichtsfrei, line 216: The wording is confusing. This is a normal TLS mode without client verification. So I would suggest to rephrase it to `Client verification disabled`?

rust/openvasd/src/tls.rs
```diff
@@ -175,24 +172,60 @@ pub fn tls_config(config: &crate::config::Config) -> Result<Option<TlsData>, Err
...
-    config.alpn_protocols = vec![b"h2".to_vec()];
+        Ok(Some((client_identifier, config, !clients.is_empty())))
+    } else {
+        match &config.tls.client_certs {
+            Some(clicerts) => {
+                match std::fs::read_dir(clicerts) {
+                    Ok(_) => {
+                        tracing::warn!(
+                            client_certs = ?clicerts,
+                            "No valid client certificates found, continuing without them in TLS mode."
```
nichtsfrei, line 202: What's the benefit of `in TLS mode.`? Isn't it clear based on the protocol information on startup?

ArnoStiefvater (approved):

Kraemii (approved):

PR #1618, opened 2024-04-09
CI: auto labeler, fix release based on label Chart: default option without OSPD, removes notus, cleaned up env handling

nichtsfrei:
- **Add: release tag toml**
- **CI: removes conventional-commits and configured disable-on for labeler**
- **CI: change init.yml to get all information**
- **CI: add debug step**
- **CI: use correct identifier on debug step**
- **CI: set labeler to main**
- **CI: adjust when to run on a PR**
- **CI: change action parameter to configuration-toml**
- **CI: release cannot be dependent on smoketest**
- **CI: utilize github.token to be more flexible**
- **CI: Helm: remove ospd-openvas from chart**
- **Chart: add check if https or http on configuring openvasd_server**


.github/workflows/control.yml
```diff
@@ -17,6 +17,11 @@ on:
     branches: [ main]
     tags: ["v*"]
   pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - closed
```
Kraemii, line 24: Is there a reason why the workflow runs, when the PR is closed?

.github/workflows/control.yml
```diff
@@ -103,25 +103,13 @@ jobs:
...
-    secrets:
-      user: ${{ secrets.GREENBONE_BOT }}
-      token: ${{ secrets.GREENBONE_BOT_PACKAGES_WRITE_TOKEN }}
-    with:
-      registry: ${{ vars.IMAGE_REGISTRY }}
   release:
-    needs: [smoketests, init]
+    # we release after container build so that we can release on a closed pr as we don't push the release container yet
+    # instead it is pushed after the tag is created. Maybe we should either change smoketests to not be dependent on a container push
+    # or push the newest container on release rather than on tag creation.
```
Kraemii, line 113: Keep track of it, so we don't forget about it

release_tag.toml
```diff
@@ -0,0 +1,24 @@
...
+]
+
+# group within groups must be defined in `changelog.toml`
+# label within groups must be defined in `labels`
+groups = [
+  { group = "Added", label = "minor_release" },
+  { group = "Changed", label = "major_release" }, # maybe minor?
+  { group = "Removed", label = "major_release" }, # maybe minor?
+  { group = "Bug Fixes", label = "patch_release" },
+]
```
Kraemii, line 18: I think Changed and Removed should always be major releases and only be used if you change something API (e.g. rename or remove a field) or system related (e.g. deprecation of ospd-openvas)

.github/workflows/control.yml
nichtsfrei: yes, so that the release label is read and interpret.

release_tag.toml
nichtsfrei: Changed or Removed should only be used when it is doing that. Keep in mind those will be rendered in the changelog and is confusing otherwise. I also used it wrongly before.

.github/workflows/control.yml
nichtsfrei: That's the reason why I leave it in draft. I wanted to change release behaviour to immediately push even when triggered by PR close and label instead of manually triggered.

nichtsfrei: For now I changed the comment. This is another issue.  i'll create a ticket for it.

nichtsfrei:
The auto label fail is because it's based on a fork that doesn't share the secrets use to manipulate the PR.

charts/openvasd/templates/deployment.yaml
```diff
@@ -82,7 +82,24 @@ spec:
...
+            mountPath: /mnt/ovc
+        command: ['sh', '-c']
+        args: ["cp /etc/gvm/ospd-openvas.conf /mnt/ovc/ospd-openvas.conf && printf \"disable_notus_hashsum_verification = True\n\" >> /mnt/ovc/ospd-openvas.conf"]
+      - name: create-dummy-openvas-log
+        image: "{{ .Values.ospd.repository }}:{{ .Values.ospd.tag }}"
+        imagePullPolicy: Always
+        volumeMounts:
+          - name: openvas-logs
+            mountPath: /mnt/ovc
+        command: ['sh', '-c']https://conceptboard.com/blog/dns-outage-management-statement/
```
ArnoStiefvater, line 101: delete

charts/openvasd/templates/deployment.yaml
```diff
@@ -135,15 +158,113 @@ spec:
...
+        {{ end }}
+        - name: openvas-logs
+          mountPath: /var/log/gvm
+        securityContext:
+          capabilities:
+            add:
+              - NET_ADMIN
+              - NET_RAW
+      {{ else }}
+     # openvas log repliacte
```
ArnoStiefvater, line 179: typo

ArnoStiefvater (approved):

PR #1637, opened 2024-05-06
Add: methology for simulating forking behaviour

nichtsfrei:
To simulate forking behaviour for sripts that rely on it e.g.:
```
set_kb_item(name: "test", value: 1);
set_kb_item(name: "test", value: 2);
set_kb_item(name: "test", value: 3);
set_kb_item(name: "test", value: 4);
set_kb_item(name: "test", value: 5);
display(get_kb_item("test"));
```

a new NaslValue `Fork` is introduced. When a method returns `Fork` then
the interpreter is cloning the register and creates as many interpreter
instances as elements within `Fork` so that a caller can then call those
interpreter after each statement.

SC-1069


ArnoStiefvater (dismissed):

rust/nasl-interpreter/src/per_statement_interpreter.rs
Kraemii, line 1: Empty file

Kraemii (dismissed):

ArnoStiefvater (approved):

Kraemii (approved):

PR #1649, opened 2024-05-23
Rs sockets

Kraemii:
Basic socket functionalities for the Rust NASL implementation

Jira: SC-1084

For testing here is a small TCP/UDP Server written in rust:
```rust
use std::{
    io::{Read, Write},
    net::TcpListener,
    thread::sleep,
    time::Duration,
};

fn host_udp() {
    use std::net::UdpSocket;

    let address = "127.0.0.1:34254";
    let socket = UdpSocket::bind(address).unwrap();
    println!("listening on {address}");

    loop {
        let mut buf = [0; 10];
        let (amt, src) = socket.recv_from(&mut buf).unwrap();
        println!("got {:?}", buf);
        let buf = &mut buf[..amt];
        buf.reverse();
        socket.send_to(buf, &src).unwrap();
    }
}

fn host_tcp() {
    let listener = TcpListener::bind("127.0.0.1:34254").unwrap();

    for stream in listener.incoming() {
        match stream {
            Ok(mut socket) => {
                println!("new client: {:?}", socket.peer_addr());
                // NOTE: You'll need to handle the fact that a single "message"
                // may come over the wire in many pieces. Some data formats
                // include a length up front that helps, others have specific
                // EOF (end of "message") sequences. There are trade-offs.
                let mut buf = [0; 10];

                let len = match socket.read(&mut buf) {
                    Ok(len) => len,
                    Err(e) => {
                        println!("{e}");
                        return;
                    }
                };
                // Receives a single datagram message on the socket. If `buf` is too small to hold
                // the message, it will be cut off.

                // Redeclare `buf` as slice of the received data and send reverse data back to origin.
                let mut buf = &mut buf[..len];
                println!("{:?}", buf.to_vec());
                buf.reverse();
                // Set number of bytes returned at once
                let mut step = 1;
                while !buf.is_empty() {
                    // Duration between sent messages
                    sleep(Duration::from_secs(2));
                    if step > buf.len() {
                        step = buf.len();
                    }
                    if let Err(e) = socket.write(&buf[0..step]) {
                        println!("{e}");
                        break;
                    }
                    buf = &mut buf[step..];
                }
            }
            Err(e) => println!("couldn't get client: {:?}", e),
        }
    }
}

fn main() {
    // host_udp();
    host_tcp();
}
```
Corresponding NASL files can be found in the examples folder (`tcp.nasl`,`udp.nasl`)


rust/examples/tls/Self-Signed mTLS Method/client_certificates.sh
```diff
@@ -33,6 +33,7 @@ openssl req -nodes \
 
 openssl rsa \
           -in end.key \
+          -passin pass:lol \
```
nichtsfrei, line 36: why is that needed?

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -0,0 +1,731 @@
...
+            }
+        } else {
+            None
+        };
+
+        // Create socket for connection
+        let socket = if let Some(timeout) = timeout {
+            TcpStream::connect_timeout(&sock, timeout)?
+        } else {
+            TcpStream::connect(sock)?
```
nichtsfrei, line 122: probably also done so in the OG implementation: should we work without a timeout at all?

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -0,0 +1,731 @@
...
+            _buffer: buffer,
+        })))
+    }
+
+    fn add(&self, socket: NaslSocket) -> usize {
+        let mut handles = self
+            .handles
+            .write()
+            .expect("Unable to access socket handles");
+        let ret;
```
nichtsfrei, line 154: why not `let ret = { ... }` ?

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -0,0 +1,731 @@
...
+            .get_mut(socket)
+            .ok_or(FunctionErrorKind::WrongArgument(format!(
+                "the given socket FD {socket} does not exist"
+            )))? {
+            NaslSocket::Tcp(conn) => {
+                let fd = conn.socket.as_raw_fd();
+
+                self.wait_before_next_probe();
+
+                let n;
```
nichtsfrei, line 239: why not let `let n = unsafe { ... }`?

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -0,0 +1,731 @@
...
+
+                // We restrict the MTU to 512 - 60 - 8 bytes, as this is the minimum
+                if len > MTU {
+                    return Err(FunctionErrorKind::Dirty(format!(
+                        "udp data exceeds the maximum length of {}",
+                        MTU
+                    )));
+                }
+
+                let n;
```
nichtsfrei, line 261: same remark as previously.

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -0,0 +1,731 @@
...
+    /// - min is the minimum number of data that must be read in case the “magic read function” is activated and the timeout is lowered. By default this is 0. It works together with length. More info https://lists.archive.carbon60.com/nessus/devel/13796
+    /// - timeout can be changed from the default.
+    fn recv(&self, r: &Register, _: &Context) -> Result<NaslValue, FunctionErrorKind> {
+        let socket = super::get_usize(r, "socket")?;
+        let len = super::get_usize(r, "length")?;
+        let _ = super::get_opt_int(r, "min");
```
nichtsfrei, line 289: why do we call get_opt_int when we immediately throw it away?

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -0,0 +1,731 @@
...
+                "the given socket FD {socket} does not exist"
+            )))? {
+            NaslSocket::Tcp(conn) => {
+                let mut old = None;
+                if let Some(timeout) = timeout {
+                    old = conn.socket.read_timeout().unwrap();
+                    conn.socket
+                        .set_read_timeout(Some(Duration::from_secs(timeout as u64)))?;
+                }
+                // TODO: process min for magic read function"
```
nichtsfrei, line 311: please comment out getting min and maybe move it to this TODO

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -0,0 +1,731 @@
...
+                let mut old = None;
+                if let Some(timeout) = timeout {
+                    old = conn.socket.read_timeout().unwrap();
+                    conn.socket
+                        .set_read_timeout(Some(Duration::from_secs(timeout as u64)))?;
+                }
+
+                let mut result = conn.socket.recv(data.as_mut_slice());
+
+                for _ in 0..5 {
```
nichtsfrei, line 336: why 6 times?

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -0,0 +1,731 @@
...
+    /// - Secret/kdc_use_tcp
+    fn open_sock_kdc(
+        &self,
+        _: &Register,
+        context: &Context,
+    ) -> Result<NaslValue, FunctionErrorKind> {
+        let hostname = match get_kb_item(context, "Secret/kdc_hostname")? {
+            NaslValue::String(x) => x,
+            x => {
+                return Err(FunctionErrorKind::Diagnostic(
```
nichtsfrei, line 379: if you rewrite get_kb_item to return a `Result<Option<NaslValue>, FunctionErrorKind>` instead of `Result<NaslValue, FunctionErrorKind>` we could differentiate if it was set to a wrong type or not set at all.

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -0,0 +1,731 @@
...
+                x
+            }
+            x => {
+                return Err(FunctionErrorKind::Diagnostic(
+                    "KB key 'Secret/kdc_port' is either missing or has wrong type".to_string(),
+                    Some(x),
+                ));
+            }
+        };
+        if port <= 0 || port > 65535 {
```
nichtsfrei, line 405: wasn't that already checked?

rust/nasl-builtin-network/src/socket.rs
nichtsfrei: That even had the potential to create a second method `get_required_kb_item` with a &str parameter to build the FunctionErrorKind accordingly.

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -0,0 +1,731 @@
...
+    fn open_sock_tcp(
+        &self,
+        register: &Register,
+        context: &Context,
+    ) -> Result<NaslValue, FunctionErrorKind> {
+        // Get port
+        let port = get_pos_port(register)?;
+        let timeout = super::get_opt_int(register, "timeout");
+        let transport = super::get_opt_int(register, "transport").unwrap_or(-1);
+        let _priority = super::get_named_value(register, "priority")
```
nichtsfrei, line 449: why is not used?

nichtsfrei (changes requested):

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -0,0 +1,731 @@
...
+            ret = handles.handles.len();
+            handles.handles.push(socket);
+        }
+        ret
+    }
+
+    /// Close a given file descriptor taken as an unnamed argument.
+    fn close(&self, r: &Register, _: &Context) -> Result<NaslValue, FunctionErrorKind> {
+        let args = r.positional();
+        let socket = match args.get(1) {
```
nichtsfrei, line 168: must be `args.get(0)` otherwise:
When I execute:
```
# let socket = UdpSocket::bind("127.0.0.1:34254")?;
data = "foo bar";
soc = open_sock_udp(34254);
n = send(socket:soc, data: data);
display(n);
close(soc);
```

with `scannerctl execute --target 127.0.0.1 udp.nasl`

I get:
```
7
2024-07-08T12:28:55.913279Z  WARN scannerctl: script error, udp.nasl, line: 146, col: 151: close(soc);: close: expected 1 arguments but got 1
```

The close error is misleading and it happens regardless if there is udp server running or not.

nichtsfrei:
The recv method on udp doesn't work like I would suspect when using it:
```
# let socket = UdpSocket::bind("127.0.0.1:34254")?;
data = "foo bar";
soc = open_sock_udp(34254);
n = send(socket:soc, data: data);
data = recv(socket: soc, length: n);
display(data);
display(n);
close(soc);
```

with a server that does immediately response:
```
use std::net::UdpSocket;

fn main() -> std::io::Result<()> {
    {
        let address = "127.0.0.1:34254";
        let socket = UdpSocket::bind(address)?;
        println!("lestening on {address}");

        loop {
            // Receives a single datagram message on the socket. If `buf` is too small to hold
            // the message, it will be cut off.
            let mut buf = [0; 10];
            let (amt, src) = socket.recv_from(&mut buf)?;
            println!("got {:?}", buf);
            // Redeclare `buf` as slice of the received data and send reverse data back to origin.
            let buf = &mut buf[..amt];
            buf.reverse();
            socket.send_to(buf, &src)?;
        }
    } // the socket is closed here
    Ok(())
}
```

the nasl script doesn't receive the response `rab oof` although it does with openvas-scanner.

rust/examples/tls/Self-Signed mTLS Method/client_certificates.sh
Kraemii: I was playing a bit around with certificates and forgot to remove it again

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -0,0 +1,731 @@
...
+
+#[derive(Default)]
+pub struct NaslSockets {
+    handles: RwLock<Handles>,
+    interval: Option<RwLock<Interval>>,
+}
+
+impl NaslSockets {
+    fn open_udp(addr: &str, port: u16) -> Result<NaslSocket, FunctionErrorKind> {
+        let socket = UdpSocket::bind("0.0.0.0:0")?;
```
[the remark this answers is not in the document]
Kraemii: This lets the OS decide which IP and Port to use for UDP communication, as we need to bind a local address to receive data. For the c implementation we take an IP set for the scanner globally, but not sure about the port. We could also do it configurable

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -0,0 +1,731 @@
...
+        socket.connect(format!("{addr}:{port}"))?;
```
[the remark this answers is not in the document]
Kraemii: No, the bind is regarding our own system and the connect is regarding the remote system we are scanning

rust/nasl-builtin-network/src/socket.rs
Kraemii: No sure, if they are needed. They are also really messy...
In the c implementation: if the given timeout is negative, it is set to 10, if it was not given, it is set to the double of the globally defined recv timeout. But I also know about services with a "sleep mode" when no one is connected for a specified time and wakes up on an incoming connection. But this normally leads to the first connection try to fail. In such cases a timeout could make sense. On the other hand it is not used in the feed, so it will always be the recv timeout times 2...

rust/nasl-builtin-network/src/socket.rs
Kraemii: We resend the data up to 5 times. It is the same, as in the c implementation, except there we don't check if we can receive data after the last send. The c implementation also checks with `select` if there is data  available to receive, but I don't see any reason to do so, as it should be the same, as `recv` regarding the timeout. Maybe you have any idea?
https://github.com/greenbone/openvas-scanner/blob/main/nasl/nasl_socket.c#L790

rust/nasl-builtin-network/src/socket.rs
Kraemii: I am currently not sure what to do with it. In the feed it is not used and also `rustls` does not support them. This is pure and deep GnuTLS stuff and is only used to define a custom TLS connection, that e.g. uses a different crypto provider. Not sure if we still should support it.

nichtsfrei: ah, ok, wouldn't it better to comment it and just don't get it?

rust/nasl-builtin-network/src/socket.rs
Kraemii: I will set the default timeout to 10 seconds for now

rust/nasl-builtin-network/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+            Ok(num as usize)
+        }
+        _ => Err(FunctionErrorKind::WrongArgument(
+            "Wrong type for argument, expected a number".to_string(),
+        )),
+    }
+}
+
+fn get_data(r: &Register) -> Result<Vec<u8>, FunctionErrorKind> {
+    Ok((&get_named_value(r, "data")?).into())
```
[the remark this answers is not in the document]
Kraemii: why not, it is easier to parse that way, as you don't need type annotations for the `into()`

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -0,0 +1,790 @@
...
+                }
+
+                let n = unsafe {
+                    libc::send(
+                        fd,
+                        data.as_ptr() as *const libc::c_void,
+                        len,
+                        flags.unwrap_or_default() as i32,
+                    )
+                };
```
Tehforsch, line 290: See comment above.

rust/nasl-builtin-network/src/socket.rs
Tehforsch: Since I had the same question as @nichtsfrei, I think the `5` at least deserves to be a named constant like `NUM_TIMES_TO_RESEND` or whatever.

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -0,0 +1,790 @@
...
+                }
+                // TLS/SSL
+                Some(tls_version) => match tls_version {
+                    OpenvasEncaps::Tls12 | OpenvasEncaps::Tls13 => {
+                        // if let Ok(fd) =
+                        //     self.open_sock_tcp_tls(context, addr, port, bufsz, timeout, vhost)
+                        // {
+                        //     fds.push(self.add(fd))
+                        //     // TODO: Set port transport
+                        // }
```
Tehforsch, line 566: Remove?

Tehforsch (changes requested):
Looks good except for a few minor nitpicks and concerns about the `unsafe blocks` in `send`.

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -0,0 +1,790 @@
...
+                    let mut ret = 0;
+                    while !data.is_empty() {
+                        let n = unsafe {
+                            libc::send(
+                                fd,
+                                data.as_ptr() as *const libc::c_void,
+                                len,
+                                flags.unwrap_or_default() as i32,
+                            )
+                        };
```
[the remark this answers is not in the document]
Kraemii: Internally `Write` use lib and the `send` method as well. The issue here is however, that `Write` does not support to set flags. I will go with your 2nd suggestion then and will comment, why it is needed and how safety is guaranteed.

ArnoStiefvater:
I removed the "minor_release" label because we are going to release this together with a few other minor_release changes early next week. It can still be merged anytime before that.

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -0,0 +1,790 @@
...
+use nasl_syntax::NaslValue;
+use pkcs8::der::Decode;
+use rustls::{
+    pki_types::{PrivateKeyDer, PrivatePkcs8KeyDer, ServerName},
+    RootCertStore,
+};
+
+use crate::{get_kb_item, get_pos_port, OpenvasEncaps};
+
+const MTU: usize = 512 - 60 - 8;
```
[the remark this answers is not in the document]
Kraemii: The MTU is the maximum size of an IP Packet that can be send to a destination without fragmentation. Each network device from source to destination has an MTU and the MTU, that must be used is the minimum of that devices. The issue is, that the path a packet travel can change and so does the MTU to the destination. I chose a general value, that most network devices support and subtract 60 for the maximum size of an IP header and 8 for the UDP header. The MTU is also only relevant for UDP, as it is not allowed to fragment UDP packages.
I will write a comment about this value and add an TODO for calculating the MTU right before sending an UDP package.

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -0,0 +1,790 @@
...
+    /// - socket: the socket, of course.
+    /// - data: the data block. A string is expected here (pure or impure, this does not matter).
+    /// - length: is optional and will be the full data length if not set
+    /// - option: is the flags for the send() system call. You should not use a raw numeric value here.
+    ///
+    /// On success the number of sent bytes is returned.
+    fn send(&self, r: &Register, _: &Context) -> Result<NaslValue, FunctionErrorKind> {
+        let socket = super::get_usize(r, "socket")?;
+        let data = super::get_data(r)?;
+        let flags = super::get_opt_int(r, "option");
```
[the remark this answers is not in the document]
Kraemii: I will leave those, as they are removed anyway in https://github.com/greenbone/openvas-scanner/pull/1697

rust/nasl-builtin-network/src/socket.rs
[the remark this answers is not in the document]
Tehforsch: Thanks for the explanation!

Tehforsch (approved):
Thanks!

PR #1657, opened 2024-06-05
Fix: Set the right status when a scan fails because an invalid host list

jjnicola:
**What**:
Fix: Set the right status when a scan fails because an invalid host list
Jira: SC-1089

**Why**:
Running a scan with a host with unsupported chars (e.g. examplö.com) the openvas scanner finishes immediately but the status is set as Succeeded in openvasd. Also set a start scan timestamp and cleanups the host information.

**How**:
start a scan via openvasd with a host with unsupported chars.

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/openvas/src/openvas.rs
+++ b/rust/openvas/src/openvas.rs
@@ -359,7 +359,7 @@ impl ScanResultFetcher for Scanner {
                     host_info: Some(hosts_info),
                 };
 
-                let scan_res = ScanResults {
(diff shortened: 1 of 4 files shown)


nichtsfrei:
Please just use `change:` when it is an API or a not downwards compatible change. This is not the case.

nichtsfrei (approved):

Kraemii (approved):

PR #1661, opened 2024-06-12
Add: nasl function support for IPv6 packet forgery, TCP and UDP over IPv6 and ICMPv6

jjnicola:
**What**:
Add: nasl function support for IPv6 packet forgery, TCP and UDP over IPv6 and ICMPv6
Jira: SC-1096

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/src/nasl/builtin/raw_ip/packet_forgery.rs
+++ b/rust/src/nasl/builtin/raw_ip/packet_forgery.rs
@@ -3122,7 +3122,7 @@ fn insert_tcp_v6_options(register: &Register, _configs: &Context) -> Result<Nasl
     ori_tcp = packet::tcp::MutableTcpPacket::new(&mut new_buf)
         .ok_or_else(|| error("No possible to create a packet from buffer".to_string()))?;
 
-    // At this point, opts len is a 4bytes multiple and the ofset is expressed in 32bits words
+    // At this point, opts len is a 4bytes multiple and the offset is expressed in 32bits words
     ori_tcp.set_data_offset(5 + opts_len as u8 / 4);
     if !opts.is_empty() {
         ori_tcp.set_options(&opts);
@@ -3398,15 +3398,16 @@ fn set_udp_v6_elements(register: &Register) -> Result<NaslValue, FnError> {
     let chksum = match register.named("uh_sum") {
         Some(ContextType::Value(NaslValue::Number(x))) if *x != 0 => (*x as u16).to_be(),
         _ => {
-            let pkt = packet::ipv6::Ipv6Packet::new(&buf)
-                .ok_or_else(|| error("No possible to create a packet from buffer".to_string()))?;
-            let udp_aux = UdpPacket::new(ori_udp.packet())
-                .ok_or_else(|| error("No possible to create a packet from buffer".to_string()))?;
+            let pkt = packet::ipv6::Ipv6Packet::new(&buf).ok_or_else(|| {
+                error("No possible to create an IPv6 segment from buffer".to_string())
+            })?;
+            let udp_aux = UdpPacket::new(ori_udp.packet()).ok_or_else(|| {
+                error("No possible to create an UDP datagram from buffer".to_string())
+            })?;
             pnet::packet::udp::ipv6_checksum(&udp_aux, &pkt.get_source(), &pkt.get_destination())
         }
     };
     ori_udp.set_checksum(chksum);
-
     // Create a owned copy of the final udp segment, which will be appended as payload to the IP packet.
     let mut fin_udp_buf: Vec<u8> = vec![0u8; udp_total_length];
     let buf_aux = <&[u8]>::clone(&ori_udp.packet()).to_owned();
@@ -3415,13 +3416,18 @@ fn set_udp_v6_elements(register: &Register) -> Result<NaslValue, FnError> {
     // Create a new IP packet with the original IP header, and the new UDP payload
     let mut new_ip_buf = vec![0u8; iph_len];
     //new_ip_buf[..].copy_from_slice(&buf[..iph_len]);
-    safe_copy_from_slice(&mut new_ip_buf[..], 0, iph_len, &buf, 0, iph_len)?;
-    new_ip_buf.append(&mut fin_udp_buf.to_vec());
+    safe_copy_from_slice(
+        &mut new_ip_buf[..],
+        0,
+        buf.len() - 1,
+        &buf,
+        0,
+        buf.len() - 1,
+    )?;
 
-    let l = new_ip_buf.len();
+    let l = fin_udp_buf.len();
     let mut pkt = packet::ipv6::MutableIpv6Packet::new(&mut new_ip_buf)
         .ok_or_else(|| error("No possible to create a packet from buffer".to_string()))?;
-
     pkt.set_payload_length(l as u16);
     pkt.set_payload(&fin_udp_buf);
 
--- a/rust/src/nasl/builtin/raw_ip/packet_forgery.rs
+++ b/rust/src/nasl/builtin/raw_ip/packet_forgery.rs
@@ -5,12 +5,13 @@
 //! Defines NASL packet forgery functions
 
 use std::{
-    net::{Ipv4Addr, Ipv6Addr, SocketAddr},
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
     str::FromStr,
 };
 
 use super::{
-    raw_ip_utils::{get_interface_by_local_ip, get_source_ip, islocalhost}, PacketForgeryError, RawIpError
+    raw_ip_utils::{get_interface_by_local_ip, get_source_ip, islocalhost},
+    PacketForgeryError, RawIpError,
 };
 
 use super::super::host::get_host_ip;
@@ -30,13 +31,14 @@ use pnet::packet::{
             MutableNeighborAdvertPacket, MutableNeighborSolicitPacket, MutableRouterAdvertPacket,
             MutableRouterSolicitPacket,
         },
-        Icmpv6Types,
+        Icmpv6Code, Icmpv6Packet, Icmpv6Types,
     },
     ip::{IpNextHeaderProtocol, IpNextHeaderProtocols},
     ipv4::{checksum, Ipv4Packet, MutableIpv4Packet},
-    ipv6::MutableIpv6Packet,
+    ipv6::{Ipv6Packet, MutableIpv6Packet},
     tcp::{TcpOption, TcpOptionNumbers, TcpPacket, *},
-    udp::UdpPacket,
+    udp::{MutableUdpPacket, UdpPacket},
+    Packet, PrimitiveValues,
 };
 
 use pnet_macros_support::types::u9be;
@@ -54,8 +56,7 @@ macro_rules! custom_error {
 }
 
 /// Default Timeout for received
-const DEFAULT_TIMEOUT: i32 = 5000;
-
+const DEFAULT_TIMEOUT_SEC: i32 = 5;
 /// Define IPPROTO_RAW
 const IPPROTO_RAW: i32 = 255;
 /// Define IPPROTO_IP for dummy tcp . From rfc3542:
@@ -63,6 +64,7 @@ const IPPROTO_RAW: i32 = 255;
 // This should not be a problem since IPPROTO_IP is used only with IPv4
 // sockets and IPPROTO_HOPOPTS only with IPv6 sockets.
 const IPPROTO_IP: u8 = 0;
+const IPPROTO_IPV6: u8 = 6;
 /// Reserved fragment flag
 const IP_RF: i64 = 0x8000;
 /// Dont fragment flag
@@ -71,6 +73,52 @@ const IP_DF: i64 = 0x4000;
 const IP_MF: i64 = 0x2000;
 /// Mask for fragmenting bits
 const IP_OFFMASK: i64 = 0x1fff;
+/// Minimum header length for IPv4 Packet Header
+const MIN_IPV4_HEADER_LENGTH: usize = 20;
+/// Fixed header length for IPv6 Packet Header
+const FIX_IPV6_HEADER_LENGTH: usize = 40;
+/// Total length for TCP Packet
+const TOTAL_TCP_HEADER_LENGTH: u16 = 40;
+/// Minimum header length for UDP Packet Header
+const MIN_UDP_HEADER_LENGTH: usize = 8;
+/// Minimum header length for ICMP Packet Header
+const MIN_ICMP_HEADER_LENGTH: usize = 8;
+/// Default header length for IPv4 Packet Header in chunk of 32bits.
+/// Used in the header length field. This the minimum 20bytes.
+const DEFAULT_IPV4_HEADER_LENGTH_32BIT_INCREMENTS: u8 = 5;
+/// Used in the header length field. This the minimum 20bytes.
+const DEFAULT_TCP_DATA_OFFSET_32BIT_INCREMENTS: u8 = 5;
+/// Default ttl value is inherit from NASL C
+const DEFAULT_TTL: u8 = 0x40;
+const TH_SYN: u16 = 0x02;
+
+#[derive(Default)]
+struct PacketPayload {
+    data: Vec<u8>,
+}
+
+impl From<Vec<u8>> for PacketPayload {
+    fn from(data: Vec<u8>) -> Self {
+        Self { data }
+    }
+}
+
+impl From<PacketPayload> for Vec<u8> {
+    fn from(data: PacketPayload) -> Self {
+        data.data
+    }
+}
+
+impl<'a> FromNaslValue<'a> for PacketPayload {
+    fn from_nasl_value(data: &'a NaslValue) -> Result<Self, FnError> {
+        match data {
+            NaslValue::Data(d) => Ok(d.clone().into()),
+            NaslValue::String(d) => Ok(d.as_bytes().to_vec().into()),
+            NaslValue::Number(d) => Ok(d.to_be_bytes().to_vec().into()),
+            _ => Ok(Vec::<u8>::new().into()),
+        }
+    }
+}
 
 /// print the raw packet
 pub fn display_packet(vector: &[u8]) {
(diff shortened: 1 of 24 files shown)


jjnicola:
https://github.com/rust-lang/socket2/pull/518

jjnicola:
waiting for https://github.com/rust-lang/socket2/pull/535

Tehforsch (commented):
A few comments before I do a detailed review:

I think nasl function argument handling could be used to remove a ton of boilerplate in this code and make the actual logic stand out more. This primarily means moving the primitive types (flags etc) to arguments given by `nasl_function` and removing all of the custom `register.get(...)` calls which are boilerplaty and have error handling that is less helpful and inconsistent with the majority of the builtins.

For example this code:

```rust
#[nasl_function]
fn get_udp_element(register: &Register) -> Result<NaslValue, FnError> {
    let buf = match register.named("udp") {
        Some(ContextType::Value(NaslValue::Data(d))) => d.clone(),
        _ => {
            return Err(FnError::missing_argument("udp"));
        }
    };
    let ip = packet::ipv4::Ipv4Packet::new(&udp)
        .ok_or_else(|| error("No possible to create a packet from buffer".to_string()))?;
    let udp = packet::udp::UdpPacket::new(ip.payload())
        .ok_or_else(|| error("No possible to create a packet from buffer".to_string()))?;
    ...
}
```

could be replaced (as a first step) by
```rust
#[nasl_function(named(udp))]
fn get_udp_element(udp: &[u8], ...) -> Result<NaslValue, FnError> {
    let ip = packet::ipv4::Ipv4Packet::new(&udp)
        .ok_or_else(|| error("No possible to create a packet from buffer".to_string()))?;
    let udp = packet::udp::UdpPacket::new(ip.payload())
        .ok_or_else(|| error("No possible to create a packet from buffer".to_string()))?;
    ...
}
```

However, I think we can go further than this and do the rusty thing of moving the parsing logic to the type system via something like this (note that `UdpPacket` uses the owned version here to avoid some borrow checker difficulties, but I think this could be fixed if it is a performance issue and we decide to go this way):

```rust
impl<'a> FromNaslValue<'a> for Ipv4Packet<'a> {
    fn from_nasl_value(val: &'a NaslValue) -> Result<Self, FnError> {
        let buf: &[u8] = <&[u8]>::from_nasl_value(val)?;
        let ip = packet::ipv4::Ipv4Packet::new(&buf)
            .ok_or_else(|| PacketForgeryError::CreatePacket)?;
        Ok(ip)
    }
}

impl<'a> FromNaslValue<'a> for UdpPacket<'a> {
    fn from_nasl_value(val: &'a NaslValue) -> Result<Self, FnError> {
        let ip: Ipv4Packet = Ipv4Packet::from_nasl_value(val)?;
        let udp = packet::udp::UdpPacket::owned(ip.payload().to_vec())
            .ok_or_else(|| PacketForgeryError::CreatePacket)?;
        Ok(udp)
    }
}
```

Doing this for the various packet types will allow us to simply write something like this in virtually all of the builtin functions:

```rust
#[nasl_function(named(udp, ...))]
fn get_udp_element(udp: UdpPacket, ...) -> Result<NaslValue, FnError> {
    ...
}
```


I think this will help separate argument handling from the code logic, parse as early as possible and allow reusing those packet types in various places.

Finally, I think a similar strategy could be used for the different elements of the packets which are currently only represented by the `element` argument string. To clarify what I mean, here is the example of the `get_udp_element` method:

```rust
pub enum UdpElement {
    SourcePort,
    DestinationPort,
    Length,
    Checksum,
}

// This is not very inspiring to write, and could even be done by a macro if we want to
impl FromStr for UdpElement {
    type Err = ArgumentError;
   
    fn from_str(s: &str) -> Result<Self, Self::Err> {
        match s {
            "uh_sport" => Ok(Self::SourcePort),
            "uh_dport" => Ok(Self::DestinationPort),
            "uh_len" => Ok(Self::Length),
            "uh_sum" => Ok(Self::Checksum),
            s => Err(ArgumentError::WrongArgument(format!("Invalid element for UDP packet: {}", s)).into()),
        }
    }
}

trait PacketElement {
    type Packet<'a>: pnet::packet::Packet;
    fn get_element<'a>(&self, packet: &Self::Packet<'a>) -> i64;
}

impl PacketElement for UdpElement {
    type Packet<'a> = packet::udp::UdpPacket<'a>;

    fn get_element<'a>(&self, packet: &Self::Packet<'a>) -> i64 {
        match self {
            UdpElement::SourcePort => packet.get_source() as i64,
            UdpElement::DestinationPort => packet.get_destination() as i64,
            UdpElement::Length => packet.get_length() as i64,
            UdpElement::Checksum => packet.get_checksum() as i64,
        }
    }
}

/// Represents the fact that the `element` argument to
/// the nasl functions below can often be of value "data",
/// meaning that the function should simply return the
/// payload of the packet.
enum ElementOrData<T> {
    Element(T),
    Data,
}

impl<'a, T> FromNaslValue<'a> for ElementOrData<T>
where
    T: FromStr<Err=ArgumentError>,
{
    fn from_nasl_value(value: &NaslValue) -> Result<Self, FnError> {
        let s = String::from_nasl_value(value)?;
        Ok(match s.as_str() {
            "data" => Self::Data,
            s => Self::Element(
                T::from_str(&s)?
            ),
        })
    }
}

impl<T: PacketElement> ElementOrData<T> {
    fn get<'a>(self, packet: &T::Packet<'a>) -> NaslValue {
        match self {
            ElementOrData::Element(element) => element.get_element(packet).into(),
            ElementOrData::Data => packet.payload().into(),
        }
    }
}


/// Get an UDP element from a IP datagram. It returns a data block or an integer, according to the type of the element. Its arguments are:
/// - udp: is the IP datagram.
/// - element: is the name of the field to get
///
/// Valid IP elements to get are:
/// - uh_sport
/// - uh_dport
/// - uh_ulen
/// - uh_sum
/// - data
#[nasl_function(named(udp, element))]
fn get_udp_element(udp: UdpPacket, element: ElementOrData<UdpElement>) -> Result<NaslValue, FnError> {
    Ok(element.get(&udp))
}
```

Using these element types again allows us to separate the parsing logic (which is often done twice for "setter" and "getter" builtins respectively) from the actual interesting logic, make future changes easier and avoid accidentally introducing typos or something similar that would break functionality.

rust/src/nasl/builtin/raw_ip/packet_forgery.rs
```diff
@@ -1763,74 +1759,47 @@ pub mod igmp {
...
     let l = ip_buf.len();
-    let mut pkt = packet::ipv4::MutableIpv4Packet::new(&mut ip_buf)
-        .ok_or(PacketForgeryError::CreatePacket)?;
+    let mut pkt =
+        MutableIpv4Packet::new(&mut ip_buf).ok_or(PacketForgeryError::CreateIpv4Packet)?;
     pkt.set_total_length(l as u16);
-    match register.named("update_ip_len") {
-        Some(ContextType::Value(NaslValue::Boolean(l))) if !(*l) => {
+    match update_ip_len {
+        Some(l) if !(l) => {
```
Tehforsch, line 1802: Is this the correct logic? If update_ip_len is unset or false, we update the packet length?

rust/src/nasl/builtin/raw_ip/packet_forgery.rs
```diff
@@ -1852,12 +1821,26 @@ fn new_raw_socket() -> Result<Socket, FnError> {
...
+        Domain::IPV6,                      // 10
+        socket2::Type::RAW,                // 3
+        Some(Protocol::from(IPPROTO_RAW)), // 255
+    ) {
+        Ok(s) => Ok(s),
+        Err(e) => Err(error(format!(
+            "new_raw_ipv6_socket: Not possible to create a raw socket: {}",
+            e
+        ))),
+    }
```
Tehforsch, line 1835: Not a required change but `.map_err` is more idiomatic and clearer imo

rust/src/nasl/builtin/raw_ip/packet_forgery.rs
```diff
@@ -2200,43 +2103,1273 @@ fn nasl_send_capture(register: &Register, configs: &Context) -> Result<NaslValue
...
+        &mut new_buf[..],
+        40,
+        hl_valid_data,
+        &opt_buf,
+        0,
+        opt_buf.len(),
+    )?;
+
+    let mut new_pkt = MutableIpv6Packet::new(&mut new_buf)
+        .ok_or_else(|| error("No possible to create a packet from buffer".to_string()))?;
```
Tehforsch, line 2357: I think this cannot return `Err`, since we always have at least `IPV6_HEADER_LENGTH`, right? So we could also unwrap. Not a required change though

rust/src/nasl/builtin/raw_ip/packet_forgery.rs
```diff
@@ -2200,43 +2103,1273 @@ fn nasl_send_capture(register: &Register, configs: &Context) -> Result<NaslValue
...
-            NaslValue::Number(TcpOptionNumbers::WSCALE.to_primitive_values().0 as i64),
+        _ => println!(
+            "\tip6_nxt   : {:?}",
+            pkt.get_next_header().to_primitive_values().0
+        ),
+    };
+}
+/// Receive a list of IP v6 packets and print them in a readable format in the screen.
+#[nasl_function]
+fn dump_ip_v6_packet(positional: CheckedPositionals<Ipv6Packet>) -> Result<NaslValue, FnError> {
```
Tehforsch, line 2385: Not a fallible function, doesn't need a return type.

rust/src/nasl/builtin/raw_ip/packet_forgery.rs
```diff
@@ -2200,43 +2103,1273 @@ fn nasl_send_capture(register: &Register, configs: &Context) -> Result<NaslValue
...
+        Icmpv6Types::EchoRequest => {
+            icmp_pkt_size = MutableEchoRequestPacket::minimum_packet_size();
+            total_length = icmp_pkt_size + data.len();
+            icmp_buf = vec![0; total_length];
+            let mut icmp_pkt =
+                packet::icmpv6::MutableIcmpv6Packet::new(&mut icmp_buf).ok_or_else(|| {
+                    error(
+                        "EchoRequest: No possible to create an icmp packet from buffer".to_string(),
+                    )
+                })?;
```
Tehforsch, line 2806: The typical comment: Here and below, we always fulfill the minimum buffer size requirement, so we can unwrap, right?

rust/src/nasl/builtin/raw_ip/packet_forgery.rs
```diff
@@ -2200,43 +2103,1273 @@ fn nasl_send_capture(register: &Register, configs: &Context) -> Result<NaslValue
...
+
+            icmp_pkt.set_icmpv6_type(icmp_type);
+            icmp_pkt.set_icmpv6_code(icmp_code);
+
+            if !data.is_empty() {
+                //buf[8..].copy_from_slice(&data[0..]);
```
Tehforsch, line 2812: remove this comment (and the ones below)?

rust/src/nasl/builtin/raw_ip/packet_forgery.rs
```diff
@@ -2200,43 +2103,1273 @@ fn nasl_send_capture(register: &Register, configs: &Context) -> Result<NaslValue
...
+
+            if let Some(x) = icmp_seq {
+                //buf[6..8].copy_from_slice(&x.to_le_bytes()[0..2]);
+                safe_copy_from_slice(&mut icmp_buf, 6, 8, &x.to_le_bytes()[..], 0, 2)?;
+            }
+        }
+        Icmpv6Types::RouterSolicit => {
+            icmp_pkt_size = MutableRouterSolicitPacket::minimum_packet_size();
+            total_length = icmp_pkt_size + data.len();
+            icmp_buf = vec![0; total_length];
```
Tehforsch, line 2829: I think extracting these three lines from each of the match arms would help with readability a bit. Sure, the types are different in all of the arms, but you could do something like
```rust
let icmp_pkt_size = match icmp_type { 
    EchoRequest => MutableEchoRequestPacket::minimum_packet_size(),
    ...
};
```
let total_length = icmp_pkt_size + data.len();
let icmp_buf = vec![0; total_length];

match icmp_type {
    ...
}

rust/src/nasl/builtin/raw_ip/packet_forgery.rs
```diff
@@ -2200,43 +2103,1273 @@ fn nasl_send_capture(register: &Register, configs: &Context) -> Result<NaslValue
...
+
+            if let Some(x) = icmp_seq {
+                //buf[6..8].copy_from_slice(&x.to_le_bytes()[0..2]);
+                safe_copy_from_slice(&mut icmp_buf, 6, 8, &x.to_le_bytes()[..], 0, 2)?;
+            }
+        }
+    }
+
+    if !data.is_empty() {
+        //buf[8..].copy_from_slice(&data[0..]);
```
Tehforsch, line 2953: remove this comment?

rust/src/nasl/utils/function/from_nasl_value.rs
```diff
@@ -98,6 +109,118 @@ impl<'a> FromNaslValue<'a> for bool {
...
+#[cfg(feature = "nasl-builtin-raw-ip")]
+impl<'a> FromNaslValue<'a> for Icmpv6Packet<'a> {
+    fn from_nasl_value(val: &'a NaslValue) -> Result<Self, FnError> {
+        let payload = get_payload_from_packet(val)?;
+        let icmp = Icmpv6Packet::owned(payload).ok_or(BuiltinError::RawIp(
+            RawIpError::PacketForgery(PacketForgeryError::CreateIcmpPacket),
+        ))?;
+        Ok(icmp)
+    }
+}
```
Tehforsch, line 196: Is there a specific reason why these impls have to be here? If not, I'd move them to the packet forgery code, so that we don't need all the feature gates here.

rust/src/nasl/builtin/raw_ip/packet_forgery.rs
```diff
@@ -554,96 +555,68 @@ fn insert_ip_options(register: &Register, _configs: &Context) -> Result<NaslValu
...
+    let mut tcp_seg = MutableTcpPacket::new(&mut tcp_buf)
+        .ok_or_else(|| error("No possible to create a packet from buffer".to_string()))?;
+    // Calculate checksum for TCP segment now, because it depends on the dst and src addresses
+    let chksum = match th_sum {
+        Some(x) if x != 0 => (x).to_be(),
+        _ => pnet::packet::tcp::ipv4_checksum(
+            &tcp_seg.to_immutable(),
+            &pkt.get_source(),
+            &pkt.get_destination(),
+        ),
```
Tehforsch, line 610: This type of code appears a lot. Creating a proper trait just to replace this expression is overkill, but I think it would be nice to create helper functions that do this for at least `tcp::ipv4_checksum`, i.e. something like `update_checksum_ipv4(tcp_seg, pkt)`

rust/src/nasl/builtin/raw_ip/packet_forgery.rs
jjnicola: The pnet library requires that you set the right payload length before setting the payload. So, you have to update the length always. After that, may be you want to send a packet with a wrong length (case `update_ip_len = false`). 
This statement just check if we have to leave the original payload length, after we use the new payload length to avoid an error.

```
// set the new payload
set_new_payload_length(new_len); // necessary to do it before setting payload to avoid crashing
set_payload(data);

// Default is true. So, not being set, or true, we leave the new payload, already set above
// set the original if it is not desired to update the len
if (update_payload_len == false)
    set_new_payload_length (original_len)
```

I know it is confusing, but unfortunatelly, pnet requires to set the new length before setting the payload. Then, given the case, we need to revert the len set to the orginal one.

jjnicola: I have rewrite the code. I hope it is easier to understand now with
` if !update_ip_len.unwrap_or(true) {...}`

Tehforsch: Got it, thanks for the explanation

rust/src/nasl/utils/function/from_nasl_value.rs
Tehforsch: The orphan rule doesn't apply on a module level, but on a crate level. It's totally fine to implement the `FromNaslValue` trait anywhere we want within scannerlib, even for foreign structs. We only run into restrictions if we want to implement it for foreign structs in foreign crates (say, `scannerctl` or if we exposed it publically and a 3rd party would try to implement it). So moving the trait impls to `packet_forgery.rs` is not a problem.

Tehforsch:
Thanks for the changes!

I have nothing new to add, I'll leave the `FromNaslValue` trait impl comment unresolved, since I think it would be valuable to do - it could be done easily without requiring any rebasing in a separate commit. 

There are also some formatting fixes needed for the CI.
Finally, `--features experimental` doesn't build on this branch. This is due to the `libssh-rs` version being too high - fixing it to `0.2.4` should work.

Tehforsch (dismissed):
Thanks for the changes!

I have nothing new to add, I'll leave the FromNaslValue trait impl comment unresolved, since I think it would be valuable to do - it could be done easily without requiring any rebasing in a separate commit.

There are also some formatting fixes needed for the CI.
Finally, --features experimental doesn't build on this branch. This is due to the libssh-rs version being too high - fixing it to 0.2.4 should work.

Tehforsch (approved):

rust/src/nasl/builtin/raw_ip/packet_forgery.rs
```diff
@@ -764,212 +797,214 @@ fn get_tcp_option(register: &Register, _configs: &Context) -> Result<NaslValue,
...
-        ));
-    }
+    let data: Vec<u8> = data.unwrap_or(orig_payload).into();
 
     let mut opts: Vec<TcpOption> = vec![];
     let mut opts_len = 0;
-    let mut opts_iter = positional.iter();
+    let mut opts_iter = tcp_opts.iter();
     loop {
         match opts_iter.next() {
```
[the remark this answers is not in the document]
jjnicola: To be done in another PR.

PR #1664, opened 2024-06-21
Add: NASL builtin functions for rsa

puethenn:
Add: NASL builtin functions for rsa

Jira-Ticket: sc-1050


nichtsfrei:
Possibly last words: I don't think a side channel attack is a high thread considering how those rsa functions are usually used within a larger scan. There I think we can ignore thise audit outcome.

doc/manual/nasl/built-in-functions/cryptographic/rsa_public_decrypt.md
```diff
@@ -10,7 +10,7 @@
 
 **rsa_public_decrypt** decrypts provided data with the public RSA key given by e and n. Returns the decrypted data.
 
-## DESCRIPTION
+## DESCRIPTIONN
```
jjnicola, line 13: Small typo
```suggestion
## DESCRIPTION
```

jjnicola (changes requested):

nichtsfrei (dismissed):

rust/nasl-builtin-cryptographic/src/rsa.rs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+fn rsa_private_decrypt(
+    register: &Register,
+    _: &Context,
+) -> Result<nasl_syntax::NaslValue, FunctionErrorKind> {
+    let data = get_required_named_data(register, "data")?;
+    let n = get_required_named_data(register, "n")?;
+    let e = get_required_named_data(register, "e")?;
+    let d = get_required_named_data(register, "d")?;
```
Kraemii, line 40: Same for the parameter `pad` as above. In case of padding is disabled, use `rsa::hazmat::rsa_decrypt()` instead

Kraemii (changes requested):
There are few things to do here. If you are not sure, how to implement specific stuff (e.g. the signing) you can look into the corresponding nasl function within the c implementation: [rsa_sign](https://github.com/greenbone/openvas-scanner/blob/main/nasl/nasl_crypto2.c#L1094)

In addition we now have switched to another approach regarding parameters for NASL functions. You can look into https://github.com/greenbone/openvas-scanner/pull/1682 for reference, if you want to use them instead

Kraemii:
A tip: it is useful to regularly call
```
git fetch
git rebase origin/main
```
to stay up to date with the main branch.

PR #1665, opened 2024-06-24
Add: push to greenbone registry

pascalholthaus:
**What**:
Add: push to greenbone registry

**Why**:
DEVOPS-1073


nichtsfrei:
do we need to push to docker hub and our internal registry?

pascalholthaus:
As far as I know, everything should be moved away from docker hub

Balou9 (approved):

pascalholthaus:
Everything piece by piece, the registry is very new to us. If everything goes well, the rest will come.

nichtsfrei (approved):

PR #1669, opened 2024-06-30
Fix: Prevent endless loop on circular dependency

nichtsfrei:
In circular dependencies, when trying to fetch the dependencies
an endless loop can occur when trying to load the VT.

To fix this, we track handled dependencies to ensure we
don't reload them, allowing scheduler to then return
MissingDependency error instead of looping indefinitely.

### DIFF
--- a/rust/nasl-interpreter/src/scheduling/mod.rs
+++ b/rust/nasl-interpreter/src/scheduling/mod.rs
@@ -7,7 +7,10 @@ mod wave;
 
 use std::{collections::HashMap, fmt::Display};
 
-use storage::item::Nvt;
+use storage::{
+    item::{NVTField, Nvt},
+    Retrieve, Retriever, StorageError,
+};
 use thiserror::Error;
 pub use wave::WaveExecutionPlan;
 /// Error cases for VTFetcher
@@ -15,7 +18,7 @@ pub use wave::WaveExecutionPlan;
 pub enum VTError {
     /// Underlying DB error.
     #[error("data-base error: {0}")]
-    DB(#[from] storage::StorageError),
+    DB(#[from] StorageError),
     #[error("{0} misses required dependencies {1:?}")]
     /// Will be returned when Scheduler tries to schedule a VT with missing dependencies
     MissingDependencies(Nvt, Vec<String>),
@@ -109,7 +112,7 @@ pub trait ExecutionPlaner {
     ///
     /// # Example
     ///
-    /// This examples shows the usage of the default implementation for a storage::Retriever to may
+    /// This examples shows the usage of the default implementation for a Retriever to may
     /// help you understanding the behavior for your own implementation or when using it.
     ///
     /// ```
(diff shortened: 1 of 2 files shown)


rust/nasl-interpreter/src/scheduling/mod.rs
```diff
@@ -284,40 +284,49 @@ where
...
                     tracing::trace!(?stage, oid = x.oid, "adding script_dependency");
-                    ret.extend(x.dependencies.iter().map(|x| {
-                        storage::Field::NVT(storage::item::NVTField::FileName(x.to_string()))
-                    }));
+                    ret.extend(
+                        x.dependencies
+                            .iter()
+                            .filter(|x| !found_dependencies.contains_key(*x))
+                            .map(|x| {
+                                storage::Field::NVT(storage::item::NVTField::FileName(
```
Tehforsch, line 312: Just out of curiosity: Is there a particular reason why you qualify these (`storage::item::NVTField`) instead of importing them via `use` or just a matter of taste?

rust/nasl-interpreter/src/scheduling/mod.rs
```diff
@@ -284,40 +284,49 @@ where
...
+                                    x.to_string(),
+                                ))
+                            }),
+                    );
                     //results[usize::from(stage)].append_vt(x.clone(), None)?;
```
Tehforsch, line 317: I realize its not a new change, but is this line still relevant?

rust/nasl-interpreter/src/scheduling/mod.rs
```diff
@@ -272,8 +272,8 @@ where
             .collect::<Vec<_>>();
         let mut results = core::array::from_fn(|_| E::default());
         let mut vts = Vec::new();
-        let mut unresolved_dependencies = Vec::new();
-        let mut resolved_dependencies = HashMap::new();
+        let mut unknown_dependencies = Vec::new();
+        let mut found_dependencies = HashMap::new();
```
Tehforsch, line 276: Nitpicking, but maybe rename this to `known_dependencies`

rust/nasl-interpreter/src/scheduling/wave.rs
```diff
@@ -486,6 +486,28 @@ mod tests {
...
+    #[tracing_test::traced_test]
+    fn circular_dependency() {
+        let generator = NvtGenerator {
+            discovery: 3,
+            nonevasive: 0,
+            exhausting: 0,
+            end: 0,
+        };
+        let mut vts = generator.generate_pyramid();
+        //let with_dependency = vts.iter().find(|x|!x.dependencies.is_empty()).unwrap();
```
Tehforsch, line 499: delete?

Tehforsch (approved):
Only had a few minor suggestions

rust/nasl-interpreter/src/scheduling/mod.rs
nichtsfrei: no, I find it sometimes easier to identify something when I am looking for something when I type in the full qualified name

rust/nasl-interpreter/src/scheduling/mod.rs
nichtsfrei: nope, going to delete.

rust/nasl-interpreter/src/scheduling/mod.rs
nichtsfrei: fair enough :)

rust/nasl-interpreter/src/scheduling/mod.rs
Tehforsch: I always just let rust-analyzers `replace qualified path with use` code action do its job afterwards

Tehforsch (approved):

PR #1671, opened 2024-07-09
Draft: Add `openvasd` scanner

Tehforsch:
Jira: SC-1071

**What**:
- Very rudimentary implementation of `ScanStarter`, within `nasl-interpreter`. Also adds `todo!()` implementations of `ScanStopper`, `ScanDeleter` and `ScanResultFetcher`.
- Add `ScannerType::Openvasd` and call the new scanner impl if it is specified.
- Refactor `openvasd::main` to simplify the setup of the three scanner types.

**Todo:**
- [ ] Proper error handling (see TODO in comments)
- [x] What is the proper value of `dirty` for `DefaultDispatcher`?
- [x] What is the correct root path to pass to the `FSPluginLoader`?
- [ ] Impl `can_start_scan` properly
- [x] Properly implement `ScanDeleter` (that is, delete the scans)
- [x] Implement `ScanResultFetcher` once results are there

SC-1105


rust/nasl-syntax/src/loader.rs
```diff
@@ -93,11 +93,8 @@ impl Default for Box<dyn Loader> {
 /// So when the root path is `/var/lib/openvas/plugins` than it will be extended to
 /// `/var/lib/openvas/plugins/plugin_feed_info.inc`.
 #[derive(Debug, Clone)]
-pub struct FSPluginLoader<P>
-where
-    P: AsRef<Path>,
-{
-    root: P,
+pub struct FSPluginLoader {
+    root: PathBuf,
```
Tehforsch, line 97: I allowed myself to remove the generic here to simplify things downstream a bit. The `new` method is still generic over some `P: AsRef<Path>`, so outward interfaces shouldnt have changed much.

rust/openvasd/src/storage/mod.rs
```diff
@@ -223,3 +223,11 @@ impl<T> Storage for T where
...
 {
 }
+
+/// A storage type that can be created from a Config and a list of feeds.
+pub trait FromConfigAndFeeds: Storage + Sized {
+    fn from_config_and_feeds(
+        config: &Config,
+        feeds: Vec<FeedHash>,
+    ) -> Result<Self, Box<dyn std::error::Error + Send + Sync>>;
+}
```
Tehforsch, line 233: Not super happy about this name or even the existence of this trait, but I felt like it would be nice to move the internal construction logic of each individual storage type out of main, so that we can just call `from_config_and_feeds(...)?` in main and be done with it. This could also have been a `TryFrom<(&Config, Vec<FeedHash>)>` impl or a method on the `Storage` trait itself.

Tehforsch:
Closing since its superseded by #1698

PR #1672, opened 2024-07-10
Add mTLS to authentication methods in response header

Tehforsch:
Jira: SC-1106

**What**:
Adds `mTLS` to authentication methods in the response header:

```
> curl --head --insecure --cert client.cert --key client.key --cacert CA/ca.cert --request GET https://localhost:3000/scans -H "X-API-KEY: mtls_is_preferred"  

HTTP/2 401 
authentication: mTLS
api-version: 1
feed-version: UNDEFINED
date: Wed, 10 Jul 2024 10:00:51 GMT
```

**How**:
The authentication methods are updated via the `ContextBuilder::add_authentication` method which was previously only called from `ContextBuilder::api_key`. The problem is that previously, the TLS setup was done in `run`, after the `Context` had already been built. Adding the authentication method at this point would mean manually modifying the `ctx.response` which is a little ugly. Instead, I decided to move the TLS setup to the `ContextBuilder::build` method, which is called just before `run` anyway. The `ContextBuilder` now has all the information to decide on the proper response.

I'm not sure what the proper response is if both api key and TLS are used, since the api key will be internally disabled. Should it only be TLS, or both?

I also made the type alias `TlsData` a proper struct `TlsConfig` (renamed for consistency), because I felt that now that it is being used more often, it deserved being its own proper type.

### DIFF
--- a/rust/openvasd/src/main.rs
+++ b/rust/openvasd/src/main.rs
@@ -24,7 +24,7 @@ fn create_context<DB, ScanHandler>(
     db: DB,
     sh: ScanHandler,
     config: &config::Config,
-) -> controller::Context<ScanHandler, DB>
+) -> Result<controller::Context<ScanHandler, DB>, Box<dyn std::error::Error + Send + Sync>>
 where
     ScanHandler: ScanStarter
         + ScanStopper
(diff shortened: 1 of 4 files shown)


rust/openvasd/src/main.rs
```diff
@@ -24,7 +24,7 @@ fn create_context<DB, ScanHandler>(
     db: DB,
     sh: ScanHandler,
     config: &config::Config,
-) -> controller::Context<ScanHandler, DB>
+) -> Result<controller::Context<ScanHandler, DB>, Box<dyn std::error::Error + Send + Sync>>
```
Tehforsch, line 27: One slightly annoying consequence is that `create_context` is now fallible.

Tehforsch:
Changed it so that only mTLS would be set, even if api is also given.

nichtsfrei (approved):

nichtsfrei:
Can you squash those commits and create a `Fix:` statement?

jjnicola (dismissed):

jjnicola (approved):

PR #1675, opened 2024-07-11
Fix unused code warning for `decode_hex`

Tehforsch:
Nobody really asked for this but seeing this on every test build annoyed me:

```
warning: function `decode_hex` is never used
 --> nasl-builtin-cryptographic/tests/helper/mod.rs:7:8
  |
7 | pub fn decode_hex(s: &str) -> Result<Vec<u8>, ParseIntError> {
  |        ^^^^^^^^^^
  |
  = note: `#[warn(dead_code)]` on by default
```

I also found a documentation string in `lib.rs` which documented this function (but it wasn't there), so I moved it to the function and removed the unnecessary TODO.

Note: Replacing the current 
`cargo clippy -- -D warnings`
 with 
`cargo clippy --tests -- -D warnings` 
in CI would have caught this.

### DIFF
--- a/rust/nasl-builtin-cryptographic/src/lib.rs
+++ b/rust/nasl-builtin-cryptographic/src/lib.rs
@@ -23,9 +23,6 @@ enum Crypt {
     Decrypt,
 }
 
-/// Decodes given string as hex and returns the result as a byte array
-// TODO only used in tests, move tests to its own module and define there
-
 pub(crate) fn lookup(function_name: &str) -> Option<NaslFunction> {
     aes_ccm::lookup(function_name)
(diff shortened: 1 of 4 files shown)


nichtsfrei (approved):

nichtsfrei:
Can you change the commit to not include `Fix:` I don't think that it is change log relevant. I think it is sufficient to remove the `:`

Tehforsch:
> Can you change the commit to not include Fix: I don't think that it is change log relevant. I think it is sufficient to remove the :

Done

nichtsfrei (approved):

ArnoStiefvater (approved):

PR #1680, opened 2024-07-16
Use `thiserror` in all libraries

Tehforsch:
Jira: SC-1118

**What**:

Move all library error types to `thiserror` and remove the `Display` / `Error` impls.

Open questions:
- [x] Is the `Diagnostic` variant on `FunctionErrorKind` used for what I think it is? If so should it really be in the enum or handled separately as part of the `Ok` variant of whatever functions use it?
- [ ] Potential bug in `impl KbAccess for RedisHelper<...>` in `openvas_redis.rs` (see TODO comment)


nichtsfrei:
I think `Diagnostic` is correctly grouped in the `FunctionErrorKind` as it is used as an abort Error from the function point of view but is treated as a diagnostic tool from an interpreter point of view combined with an Optional result type for the user that defaults to `NaslValue::Null`. That said the description is probably misleading as it isn't just used for diagnostic but also for a abort condition that may interpreter but not user relevant.

rust/openvas/src/openvas_redis.rs
```diff
@@ -62,50 +74,37 @@ pub trait KbAccess {
 impl KbAccess for RedisHelper<RedisCtx> {
     /// Provide access to the cache
     fn kb_id(&self) -> RedisStorageResult<u32> {
-        let cache = Arc::as_ref(&self.task_kb)
-            .lock()
-            .map_err(|e| DbError::SystemError(format!("{e:?}")))?;
-        Ok(cache.db)
+        // TODO: Should this really be self.lock_task_kb? This seems
```
nichtsfrei, line 77: @jjnicola can you look into that?

nichtsfrei (approved):

nichtsfrei (changes requested):
Can you fix the clippy warnings?

Tehforsch:
> I think Diagnostic is correctly grouped in the FunctionErrorKind as it is used as an abort Error from the function point of  view but is treated as a diagnostic tool from an interpreter point of view combined with an Optional result type for the user that defaults to NaslValue::Null. That said the description is probably misleading as it isn't just used for diagnostic but also for a abort condition that may interpreter but not user relevant.

Ok - thanks for the explanation.

> Can you fix the clippy warnings?

Done

nichtsfrei (approved):

ArnoStiefvater (approved):

nichtsfrei (approved):

Kraemii (approved):

rust/openvas/src/openvas_redis.rs
jjnicola: It is ok as it is. Just the method documentation is wrong/confusing. It returns the redis namespace ID of the current host kb that must be stored as running. Cache is related to the feed, where the vts metadata is stored. So, feel free to update the functions docs

jjnicola (approved):

PR #1682, opened 2024-07-22
Nasl argument handling

Tehforsch:
Jira: SC-1116

This introduces ergonomics improvements for defining NASL functions, in particular for the handling of arguments and return values. Error messages for missing arguments and wrong argument types are automatically generated. This is just a rough proposal, suggestions very welcome.

This also implements the remaining NASL builtin string functions.

Before:
```rust
fn stridx(register: &Register, _: &Context) -> Result<NaslValue, FunctionErrorKind> {
    let positional = resolve_positional_arguments(register);
    let haystack = match positional.first() {
        Some(NaslValue::String(x)) => x,
        x => return Err(("0", "string", x).into()),
    };
    let needle = match positional.get(1) {
        Some(NaslValue::String(x)) => x,
        x => return Err(("1", "string", x).into()),
    };
    let offset = match positional.get(2) {
        Some(NaslValue::Number(x)) => *x as usize,
        _ => 0_usize,
    };
    Ok(match &haystack[offset..].find(needle) {
        Some(index) => NaslValue::Number(*index as i64),
        None => NaslValue::Number(-1),
    })
}
```
After:
```rust
#[nasl_function]
fn stridx(haystack: String, needle: String, offset: Option<usize>) -> i64 {
    let offset = offset.unwrap_or(0);
    match &haystack[offset..].find(&needle) {
        Some(index) => *index as i64,
        None => -1,
    }
}
```

Examples of how the implementations are provided in `nasl_builtin_string` and `nasl_builtin_http`.

# Specifics

The attribute-style proc macro `nasl_function` takes a function with any number of arguments and a normal return value, and transforms it into a function with signature
`fn(&Context, &Register) -> Result<NaslValue, FunctionErrorKind>`.
which can be used in a `NaslFunctionExecuter`, just as before.
(Note: If the function receives a `&self` argument, it is transformed into a function with signature 
`fn(&self, &Context, &Register) -> Result<NaslValue, FunctionErrorKind>` to support use cases such as in `nasl_builtin_http`)

In order for this to work, each argument type has to implement the trait `FromNaslValue` which attempts converting from `NaslValue` to the corresponding type, returning a `FunctionErrorKind` if the conversion failed. Similarly, the return type has to implement `ToNaslResult` which performs the conversion back to `Result<NaslValue, FunctionErrorKind>`. Most primitive types can easily support these conversions, for example:

```rust
impl<'a> FromNaslValue<'a> for &'a str {
    fn from_nasl_value(value: &'a NaslValue) -> Result<Self, FunctionErrorKind> {
        match value {
            NaslValue::String(string) => Ok(string),
            _ => Err(FunctionErrorKind::WrongArgument(
                "Expected string.".to_string(),
            )),
        }
    }
}

impl ToNaslResult for String {
    fn to_nasl_result(self) -> Result<NaslValue, FunctionErrorKind> {
        Ok(NaslValue::String(self))
    }
}
```

However, it is also possible to implement to implement these traits wherever needed for more complex types. For example, we could do something like `impl FromNaslValue for MacAddress` to move all the verification logic whether the given string is indeed a correct mac address out of the nasl function itself and into the argument handling.

# Special arguments
## Optional arguments
In order to receive optional arguments, where omitting them during the function call should not result in an error, we can simply make the function receive `Option<T>`. For example:

```rust
#[nasl_function]
fn hexstr(s: Option<NaslValue>) -> Option<String> {
    match s? {
        NaslValue::String(s) => Some(encode_hex(s.as_bytes())),
        NaslValue::Data(bytes) => Some(encode_hex(&bytes)),
        _ => None,
    }
}
```

## Named arguments
In order to receive named arguments, additional attributes can be given to the `nasl_function` macro. For example:

```rust
#[nasl_function(maybe_named(length), named(data))]
fn crap(length: usize, data: Option<&str>) -> String {
    let data = data.unwrap_or("X");
    data.repeat(length)
}
```

In this example, `length` may be either positional or named, which is why it is of `maybe_named` type. Data may only be given as a named argument, so it is `named`. Note that `data` is of type `Option<&str>`, making it optional.

## Maybe arguments
I didn't really know what to call these, but there is an additional pattern that seems to occur every now and then: A given argument should be of some type `T`, but if it isn't we will assume a default value (or return `Null`) instead of failing with an error. I added the `Maybe` type to represent this case:

```rust
fn data_to_hexstr(bytes: Maybe<&[u8]>) -> Option<String> {
    bytes.map(encode_hex)
}
```

## `Context`
The `Context` can still be obtained in a function, simply by adding it as an argument:
```rust
#[nasl_function]
fn foo(context: &Context, ...) -> ... {
}
```
The `nasl_function` macro will then pass it along automatically.

## `Positionals`
For functions that receive lists of positional arguments, the `Positionals` and `CheckedPositionals` types are provided (not sure about the names). `Positionals` simply provides an iterator over the arguments with item type `Result<T, FunctionErrorKind>`,  meaning that the conversion is done while iterating over the arguments:

```rust
#[nasl_function]
fn foo(positional: Positionals<&NaslValue>) -> Result<(), FunctionErrorKind> {
    for arg in positional.iter() {
        do_something(arg?) // Conversion would fail here
    }
}
```

The `CheckedPositionals` type checks all arguments before the inner function is even called (and returns an error if one fails to convert), which is convenient but also potentially slow:

```rust
#[nasl_function]
fn foo(positional: CheckedPositionals<&NaslValue>) {
    for arg in positional.iter() {
        do_something(arg)
    }
}
```


nichtsfrei:
There is one think I would miss.

Imagine you need to store a socket look up or numbering specifically for a function call within a spawn of a scan. Currently you would implement like this:

```
#[derive(Debug, Clone, Default)]
/// The description builtin function
pub struct Reporting {
    id: Arc<RwLock<usize>>,
}

impl Reporting {
    fn id(&self) -> usize {
        let mut id = self.id.as_ref().write().expect("expected write lock");
        let result = *id;
        *id += 1;
        result
    }

    fn store_result(
        &self,
        typus: ResultType,
        register: &Register,
        context: &Context,
    ) -> Result<NaslValue, FunctionErrorKind> {
        ...
        let result = models::Result {
            id: self.id(),
           .. Default::default(),
        };
        context.dispatcher().retry_dispatch(
            5,
            context.key(),
            storage::Field::Result(result.into()),
        )?;
        Ok(NaslValue::Null)
    }

    fn log_message(
        &self,
        register: &Register,
        context: &Context,
    ) -> Result<NaslValue, FunctionErrorKind> {
         ...
        let result = models::Result {
            id: self.id(),
           .. Default::default(),
        };
        ...
        Ok(NaslValue::Null)
    }
}

impl nasl_builtin_utils::NaslFunctionExecuter for Reporting {
    fn nasl_fn_execute(
        &self,
        name: &str,
        register: &Register,
        context: &Context,
    ) -> Option<nasl_builtin_utils::NaslResult> {
        match name {
            "log_message" => Some(self.log_message(register, context)),
            "security_message" => Some(self.security_message(register, context)),
            "error_message" => Some(self.error_message(register, context)),
            _ => None,
        }
    }

    fn nasl_fn_defined(&self, name: &str) -> bool {
        matches!(name, "log_message" | "security_message" | "error_message")
    }
}
```

Although I think your solution is awesome for things that don't need to cache implementation specific details I believe the mast majority would need it. Therefore I have a feeling if we just approve it as is it promotes the tendency to just put those things into the Context. That would not just pollute it but also make it very hard to switch the builtin implementation to another one. 

I think a better construct would be:
```
/// Generates the NaslFunctionExecutor for  reporting
#[derive(NaslFunctionRegister)]
pub struct Reporting {
    id: Arc<RwLock<usize>>,
}

impl Reporting {
  #[nasl_function]
  fn log_message(&self, data: &[u8], port: String, proto: String) -> i64 {
      ...
  }
}
```

that way we would still enforce a struct definition to move away from the 'I have to put everything in Context` insensitive.

Tehforsch:
I'm not sure if I understood you correctly, so feel free to ignore the rest of this comment if it doesn't address your concerns properly:

I'm a little confused because I think the definition of the nasl_function itself works exactly as you propose in the better construct (this is quite similar to the `close_handle` example in `nasl_builtin_http`):

```rust
pub struct Reporting {
    id: Arc<RwLock<usize>>,
}

impl Reporting {
  #[nasl_function]
  fn log_message(&self, data: &[u8], port: String, proto: String) -> i64 {
      ...
  }

  #[nasl_function]
  fn error_message(&self, data: &[u8], port: String, proto: String) -> i64 {
      ...
  }
}
```

However, one still has to do the `NaslFunctionExecutor` somewhat by hand:

```rust
impl nasl_builtin_utils::NaslFunctionExecuter for Reporting {
    fn nasl_fn_execute(
        &self,
        name: &str,
        register: &Register,
        context: &Context,
    ) -> Option<nasl_builtin_utils::NaslResult> {
        match name {
            "log_message" => Some(self.log_message(register, context)),
            "security_message" => Some(self.security_message(register, context)),
            "error_message" => Some(self.error_message(register, context)),
            _ => None,
        }
    }

    fn nasl_fn_defined(&self, name: &str) -> bool {
        matches!(name, "log_message" | "security_message" | "error_message")
    }
}
```

So the current proposal simplifies the definition of `log_message` (and `error_message`, ...) and cache-specific things can be passed via the `&self` parameter.
Now, it's true that the current proposal doesn't really affect the `NaslFunctionExecutor` implementation. Adding a `derive(NaslFunctionExecutor)` macro would be great, but it would need to be a little more complicated because we'd still need to specify what the names of the functions are, since a derive macro on the struct itself cant query the names of the functions in the impl blocks of the same struct (although that would be awesome sometimes).
One way to do this automatically that I thought of at one point but decided it would make things worse rather than better would be something like
```rust
#[nasl_function_block]
impl Reporting {
    #[nasl_function]
    fn log_message(&self, data: &[u8], port: String, proto: String) -> i64 {
        ...
    }

    #[nasl_function]
    fn error_message(&self, data: &[u8], port: String, proto: String) -> i64 {
        ...
    }
}
```

after which the `nasl_function_block` macro would automatically generate the `NaslFunctionExecutor` impl for the `Reporting` struct with all of the functions (or at least the ones with the `#[nasl_function]` attribute) contained in the impl block as members of it. This would be super inflexible though because we could not even split the implementation of the nasl functions over multiple impl blocks, let alone modules, since proc macros cant share state as far as I know. Could be that I haven't thought this through fully and there is a much more convenient way to do this. At the very least something like this should work

```rust
#[derive(NaslFunctionExecutor)]
#[nasl_functions(log_message, error_message, ...)]
struct Reporting {
    ...
}
```

nichtsfrei:
My main concern was the possibility to bind them to a struct as I didn't see it in the examples of the PR description. We can think about removing the trait boilerplate later.

nichtsfrei (approved):

PR #1685, opened 2024-07-26
Fix: Feed is loaded many times in readis after openvasd restart

jjnicola:
**What**:
Check if the current version is already in the cache before uploading. Also remove the old entries before uploading, if the entry is present to avoid the amount if items inside keys (duplication).
Jira: SC-1122

**Why**:
After reloading openvasd, the feed is completely uploaded without deleting the present entries which increases the memory usages.
Also, If the feed does not changed (same feed version), the feed shouldn't be uploaded.

**How**:
restart openvasd  many times and check the redis cache

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/openvasd/src/storage/mod.rs
+++ b/rust/openvasd/src/storage/mod.rs
@@ -186,6 +186,11 @@ pub trait NVTStorer {
 
     /// Returns the currently stored feed hash.
     async fn feed_hash(&self) -> Vec<FeedHash>;
+
+    /// Returns the current feed version, if any.
+    async fn current_feed_version(&self) -> Result<String, Error> {
+        Ok("".to_string())
+    }
 }
 
 #[async_trait]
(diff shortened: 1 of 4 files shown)


rust/openvasd/src/storage/mod.rs
```diff
@@ -186,6 +186,11 @@ pub trait NVTStorer {
 
     /// Returns the currently stored feed hash.
     async fn feed_hash(&self) -> Vec<FeedHash>;
+
+    /// Returns the current feed version, if any.
+    async fn current_feed_version(&self) -> Result<String, Error> {
+        Ok("".to_string())
+    }
```
Tehforsch, line 193: Just a question about this: Do the other storage types not have any information about the feed version?

Tehforsch (dismissed):

rust/openvasd/src/storage/mod.rs
jjnicola: It seems that the metadata is completely replace in the other storage types, both `memory` and `fs` when the daemon is restarted.

Tehforsch (approved):

PR #1688, opened 2024-07-31
Refactor SSH, add `russh` as alternative to `libssh`

Tehforsch:
Jira: SC-1124

Todo:
- [x] Make sure we return diagnostics instead of other errors wherever appropriate.
- [x] Check return codes.
- [x] Make tests more robust.

Compatibility things that can become separate issues:
- [x] Support custom sockets.
- [x] server/client and client/server ciphers are not differentiated in russh (which makes sense to me). Currently we use the intersection of the two, which should be fine. However, these keys are never actually used in the feed, so perhaps propose a removal of these altogether, to simplify the code.
- [x] ed25519 currently unsupported for key exchange, will be fixed by increasing the version to include https://github.com/Eugeny/russh/pull/360
- [x] Support public key auth by creating the `KeyPair` in russh impl
- [x] Fix stderr for `call` in russh impl.
- [x] Document issues with blocking libssh_rs calls being "fake" async. Document the proposed solution which is wrapping `libssh_rs::Session` (and the channel) into a `Arc<Mutex>` and wrapping all of its methods in a `spawn_blocking` call. This isn't pretty but should solve the issue and provide proper concurrency.


Tehforsch:
Blocked by #1671

rust/Cargo.toml
```diff
@@ -52,6 +52,8 @@ rand = "0.8.5"
 redis = "0.22.3"
 regex = "1.10.6"
 ripemd = "0.1.3"
+russh = "0.45.0"
+russh-keys = "0.45.0"
```
Tehforsch, line 56: Having these here even if compiling with `nasl-builtin-libssh` is a little unfortunate, but the alternative is to

1. Add a feature `nasl-builtin-russh`.
2. Activate that feature by default
3. Toggle the `russh` dependency on that feature.

but then specifying `--features experimental` will not work directly, since it would not actually turn the `nasl-builtin-russh` feature off, so we'd have to do `--no-default-features --features experimental,and,all,other,features,that,are,in,default` which is very annoying.

rust/Cargo.toml
```diff
@@ -97,6 +99,7 @@ members = [
 [dev-dependencies]
 tracing-test = "0.2.5"
 criterion = "0"
+once_cell = "1.20.1"
```
Tehforsch, line 102: This is needed to lock the test server behind a static mutex. I think we very likely indirectly depend on this anyways, since it is a pretty common library.

rust/crates/nasl-function-proc-macro/src/codegen.rs
```diff
@@ -195,6 +195,7 @@ impl<'a> ArgsStruct<'a> {
         let mangled_ident = Ident::new(&mangled_name, ident.span());
         let inner_call = self.get_inner_call_expr(&mangled_ident, asyncness);
         quote! {
+            #[allow(clippy::too_many_arguments)]
```
Tehforsch, line 198: I added this because we cannot really do anything about the number of arguments in our builtin functions anyways and without it adding it in the macro, there is no way for the user of `nasl_function!` to add this `#[allow(...)]` attribute themselves.

rust/src/nasl/builtin/string/mod.rs
```diff
@@ -19,26 +19,6 @@ use std::num::ParseIntError;
...
-        match value {
-            NaslValue::String(string) => Ok(Self(string.clone())),
-            NaslValue::Data(buffer) => Ok(Self(bytes_to_str(buffer))),
-            _ => Err(FunctionErrorKind::WrongArgument(
-                "Expected string or byte buffer.".to_string(),
-            )),
-        }
-    }
-}
-
```
Tehforsch, line 41: Although this will probably not exist if we ever autoconvert from bytes to string, this is quite useful for now, so I moved it to `nasl/function/types.rs`

rust/src/nasl/builtin/README.md
```diff
@@ -62,19 +62,3 @@ and then extend the register builder within the [nasl_std_variables] with the im
...
-
-```toml
-
-nasl-builtin-ssh = {path = "../nasl-builtin-ssh", optional = true}
-
-[features]
-experimental = ["nasl-builtin-ssh"]
-```
-
-It is recommended to toggle on the crate name and not on experimental to also enable toggling those without using experimental.
```
Tehforsch, line 80: I removed this because it is out of date, but I can also correct it and add it back in if thats what we want.

rust/src/nasl/builtin/ssh/libssh/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+use super::Ssh;
+
+pub use libssh_rs::AuthMethods;
+pub use session::SshSession;
+
+pub type SessionId = i32;
+pub type Socket = std::os::raw::c_int;
+
+pub fn get_log_level() -> LogLevel {
+    let verbose = std::env::var("OPENVAS_LIBSSH_DEBUG")
```
nichtsfrei, line 29: q: what's the reasoning to not follow to overall logging level?

rust/src/nasl/builtin/ssh/libssh/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+        None
+    } else {
+        Some(
+            items
+                .iter()
+                .map(|name| name.as_ref().to_string())
+                .collect::<Vec<_>>()
+                .join(","),
+        )
+    }
```
nichtsfrei, line 53: nit: I think reduce would have been a more approachable choice:
```
    items
        .iter()
        .map(|name| name.as_ref().to_string())
        .reduce(|acc, name| acc + ", " + &name)
```

rust/src/nasl/builtin/ssh/libssh/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+
+type BorrowedSession<'a> = MutexGuard<'a, SshSession>;
+
+impl Ssh {
+    pub async fn find_id<'a>(
+        &'a self,
+        f: impl for<'b> Fn(&BorrowedSession<'b>) -> bool,
+    ) -> Result<Option<SessionId>> {
+        for id in self.ids() {
+            let session = self.get_by_id(*id).await?;
```
nichtsfrei, line 64: q: under which criteria does get_by_id fail? I'm asking because it is hard to tell if the result of that would be always Err when ids are available or not.

rust/src/nasl/builtin/ssh/libssh/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+    pub async fn add_new_session(
+        &mut self,
+        f: impl Fn(&mut BorrowedSession) -> Result<()>,
+    ) -> Result<SessionId> {
+        let id = self.next_session_id()?;
+        let session = Mutex::new(SshSession::new(id)?);
+        {
+            let mut borrowed_session = session.lock().await;
+            if let Err(e) = f(&mut borrowed_session) {
+                borrowed_session.disconnect().await?;
```
nichtsfrei, line 83: q: I don't understand why you need to disconnect immediately.

rust/src/nasl/builtin/ssh/libssh/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+            if let Some(scciphers) = to_comma_separated_string(&scciphers) {
+                session.set_option(SshOption::CiphersSC(scciphers))?;
+            }
+            session.set_option(SshOption::Port(port))?;
+
+            if let Some(socket) = socket {
+                // This is a fake raw socket.
+                // TODO: implement openvas_get_socket_from_connection()
+                let my_sock = UdpSocket::bind("127.0.0.1:0").unwrap();
+                debug!(
```
nichtsfrei, line 125: please use tracing

rust/src/nasl/builtin/ssh/libssh/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+            if let Some(csciphers) = to_comma_separated_string(&csciphers) {
+                session.set_option(SshOption::CiphersCS(csciphers))?;
+            }
+            if let Some(scciphers) = to_comma_separated_string(&scciphers) {
+                session.set_option(SshOption::CiphersSC(scciphers))?;
+            }
+            session.set_option(SshOption::Port(port))?;
+
+            if let Some(socket) = socket {
+                // This is a fake raw socket.
```
nichtsfrei, line 122: q: what's a fake raw socket?

rust/src/nasl/builtin/ssh/libssh/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+                // TODO: implement openvas_get_socket_from_connection()
+                let my_sock = UdpSocket::bind("127.0.0.1:0").unwrap();
```
nichtsfrei, line 124: q: should that always be '127.0.0.1:0'?

rust/src/nasl/builtin/ssh/libssh/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+                debug!(
+                    ip_str,
+                    sock_fd = my_sock.as_raw_fd(),
+                    nasl_sock = socket,
+                    "Setting SSH fd for socket",
+                );
+                session.set_option(SshOption::Socket(my_sock.as_raw_fd()))?;
+            }
+            debug!(
```
nichtsfrei, line 133: please use tracing::

rust/src/nasl/builtin/ssh/russh/session.rs
```diff
@@ -0,0 +1,235 @@
...
+
+        loop {
+            // There's an event available on the session channel
+            let Some(msg) = channel.wait().await else {
+                break;
+            };
+            match msg {
+                // Write data to the terminal
+                ChannelMsg::Data { ref data } => {
+                    stdout.push_str(&bytes_to_str(data));
```
nichtsfrei, line 140: q: did you verify that this is just stdout? what happens with stderr data?

rust/src/nasl/builtin/ssh/russh/session.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    pub async fn auth_password(&mut self, login: &str, password: &str) -> Result<(), SshError> {
+        self.session
+            .authenticate_password(login, password)
+            .await
+            .map_err(|_| SshErrorKind::UserAuthPassword.with(self.id))
+            .map(|_| ())
+    }
+
+    pub async fn auth_public_key(
```
nichtsfrei, line 165: q: do we support that currently? When I look at https://docs.greenbone.net/API/OSP/osp-22.4.html#element_credential or https://github.com/greenbone/openvas-scanner/blob/main/rust/src/models/credential.rs  i don't see a possibility to provide key pairs?

rust/src/nasl/utils/context.rs
```diff
@@ -398,6 +399,15 @@ impl<'a> Context<'a> {
         &self.target
     }
 
+    /// Get the target host
+    pub fn target_ip(&self) -> Result<IpAddr, AddrParseError> {
+        match self.target() {
+            x if !x.is_empty() => x.to_string(),
+            _ => "127.0.0.1".to_string(),
```
nichtsfrei, line 406: q: do we want to use localhost if there is no target provided? Or shouldn't we enforce that a target is set when we start to run?

rust/src/nasl/builtin/ssh/libssh/mod.rs
Tehforsch: I don't know. I did this because it was implemented in the [original implementation](https://github.com/greenbone/openvas-scanner/blob/e6abbc2a30283a3f036ba3ccf24bb8a82ac3354e/rust/src/nasl/builtin/ssh/mod.rs#L467-L469).

rust/src/nasl/builtin/ssh/libssh/mod.rs
Tehforsch: Note that we have to do
```rust
    items
        .iter()
        .map(|name| name.as_ref().to_string())
        .reduce(|acc, name| acc + ", " + &name)
        .unwrap_or("".to_string())
```

to deal with empty iterators, so I am not sure how much cleaner it is. Alternatively, `fold` with `""` as starting value works too. A third alternative is `itertools::join(..., ",")` which I think is the cleanest and avoids collecting altogether (although I think internally it probably just does something similar to your reduce call anyways).

rust/src/nasl/builtin/ssh/libssh/mod.rs
Tehforsch: `get_by_id` only fails if the ID does not exist (the locking could throw an error with std mutexes but not with tokio async mutexes).

rust/src/nasl/builtin/ssh/libssh/mod.rs
Tehforsch: I did this to replicate the original code: libssh requires you to manually call the `disconnect` method if the session is dropped (as opposed to russh where RAII does its job). 
So the `add_new_session` method is effectively just making sure that the connection is closed if any error occurs during initialization (which effectively means that we will disconnect the session if any `set_option` or the `connect` call fails).

However, I now realize that this method is a little overkill - in the original implementation it was called from multiple places but I refactored it so that only `connect` calls this method anyways, so I should probably simplify it by replacing the closure with the code within `connect`, which should make it a little easier to understand.

rust/src/nasl/builtin/ssh/libssh/mod.rs
Tehforsch: This is a remnant from the original implementation. I think the comment is intended to clarify that currently we completely ignore the socket argument to this function and instead just hope that we can bind to `127.0.0.1:0`. This is clearly not the intended behavior, see the TODO comment below.

rust/src/nasl/builtin/ssh/libssh/mod.rs
Tehforsch: See comment above. It shouldnt.

rust/src/nasl/builtin/ssh/russh/session.rs
Tehforsch: I had a todo a few lines above because I hadn't implemented stderr properly, but now I implemented stderr for russh (turns out that for stderr, you receive an `ExtendedData` variant with a specific number set to exactly 1. Love the SSH protocol)  and added a "test" for it.

I also added a comment explaining that the compatibility mode for `ssh_request_exec` is currently not implemented as described (and also wasn't before this PR). We should check whether this is ever used in the described way and then either get rid of it or implement it properly.

rust/src/nasl/builtin/ssh/russh/session.rs
Tehforsch: Good point. It existed in the original code, so I kept it around in the refactoring and then wanted to add it in russh as well but didn't really know how to implement it. Maybe @jjnicola knows whether this should exist or not. I'll get rid of it if it shouldnt.

rust/src/nasl/utils/context.rs
Tehforsch: I kept this as in the original code, but I agree with you and would much prefer doing any kind of logic like this once when we start the run and then never touch it again. I think @Kraemii wanted to change this behavior originally, but I don't remember the reason why this was put on hold.

rust/src/nasl/builtin/ssh/libssh/mod.rs
Tehforsch: This is a little more annoying than I would like - as far as I can tell there is no way to retrieve the active level filter from `tracing_subscriber`, so we'll have to remember the logging level and store it in the context, for example. I'm not sure this is what we want to do.

Tehforsch: Maybe https://docs.rs/tracing/latest/tracing/level_filters/struct.LevelFilter.html#method.current helps

Kraemii: Maybe `tracing_subscriber::filter::LevelFilter::current() ` will work

rust/src/nasl/builtin/ssh/russh/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+use russh::cipher;
+use russh_keys::key;
+
+use super::error::Result;
+
+use super::sessions::SshSessions;
+
+pub type SessionId = i32;
+pub type Port = u16;
+// TODO: Fix this
```
Kraemii, line 17: What should be fixed here?

rust/src/nasl/utils/context.rs
Kraemii: The reason this was put on hold, was that I wanted to enforce an IP address for the target provided within the context. It was said to discuss this first, as it is may be wanted to also have a host name as a target

Tehforsch: Opened jira issue for this

ArnoStiefvater (approved):

PR #1695, opened 2024-08-09
Fix: change HEAD endpoint

jjnicola:
**What**:
Fix: change HEAD endpoint
- change HEAD to not allow invalid urls anymore. It return 200 OK as before for any Unknown path.
- change `HEAD /scans` to return 204 when a client certificate is known and 401 when not

Jira SC-1125

**Why**:
 To test if an instance is allowed to communicate with a sensor

**How**:
`curl --verbose --request HEAD https://localhost:3000/scans`
`curl --verbose --request HEAD https://localhost:3000/invalid`
`curl --verbose --cacert /home/jnicola/install/etc/openvasd/tls/server.pem --cert /home/jnicola/install/etc/openvasd/tls/clients/client.pem --cert-type PEM --key /home/jnicola/install/etc/openvasd/tls/clients/client.rsa --request HEAD https://localhost:3000/scans`
`curl --verbose --cacert /home/jnicola/install/etc/openvasd/tls/server.pem --cert /home/jnicola/install/etc/openvasd/tls/clients/client.pem --cert-type PEM --key /home/jnicola/install/etc/openvasd/tls/clients/client.rsa --request HEAD https://localhost:3000/invalid`

**Checklist**:

- [ ] Tests
- [X] PR merge commit message adjusted


rust/openvasd/src/controller/entry.rs
```diff
@@ -185,10 +185,6 @@ where
         let cid = self.cid.clone();
         Box::pin(async move {
             use KnownPaths::*;
-            // on head requests we just return an empty response without checking the api key
```
nichtsfrei, line 188: I think that effectively disables HEAD for all endpoints but for `/scans`.

nichtsfrei (changes requested):

rust/openvasd/src/controller/entry.rs
jjnicola: Fixed. It behaves as before, except for /scans, which checks for authentication. It responds with 204 or 401 given the case.

nichtsfrei (approved):

PR #1697, opened 2024-08-12
Use nasl function macro for builtin socket functions

Kraemii:
Use nasl function macro for builtin socket functions and use network utility functions for parsing ports and IP strings.
Also:
- Fix for creating UDP sockets when using IPv6.
- adds an dns lookup for kdc, which is given as a hostname.

Depends on https://github.com/greenbone/openvas-scanner/pull/1693
Jira: SC-1084

### DIFF
--- a/rust/nasl-builtin-network/src/socket.rs
+++ b/rust/nasl-builtin-network/src/socket.rs
@@ -5,7 +5,7 @@
 use std::{
     fs,
     io::{self, BufReader, Read, Write},
-    net::{TcpStream, ToSocketAddrs, UdpSocket},
+    net::{IpAddr, TcpStream, ToSocketAddrs, UdpSocket},
     os::fd::AsRawFd,
     sync::{Arc, RwLock},
     thread::sleep,
@@ -21,7 +21,7 @@ use rustls::{
     ClientConfig, ClientConnection, RootCertStore, Stream,
 };
 
-use crate::{get_kb_item, mtu, verify_port, OpenvasEncaps};
+use crate::{get_kb_item, mtu, network_utils::ipstr2ipaddr, verify_port, OpenvasEncaps};
 
 // Number of times to resend a UDP packet, when no response is received
 const NUM_TIMES_TO_RESEND: usize = 5;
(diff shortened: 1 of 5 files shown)


rust/nasl-builtin-network/src/socket.rs
```diff
@@ -133,9 +147,17 @@ pub struct NaslSockets {
...
-        socket.connect(format!("{addr}:{port}"))?;
+    fn open_udp(addr: IpAddr, port: u16) -> Result<NaslSocket, FunctionErrorKind> {
+        let sock_addr = (addr, port).to_socket_addrs()?.next().ok_or_else(|| {
+            FunctionErrorKind::Diagnostic(
+                format!(
+                "the given address and port do not correspond to a valid address: {addr}:{port}",
+            ),
+                None,
+            )
+        })?;
```
Tehforsch, line 158: Since this is used multiple times: Extract into `resolve_sock_addr(addr, port)`?

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -144,18 +166,19 @@ impl NaslSockets {
...
-        let sock = format!("{ip}:{port}").to_socket_addrs()?.next().ok_or(
-            FunctionErrorKind::WrongArgument(format!(
+        let sock = (ip, port)
+            .to_socket_addrs()?
+            .next()
+            .ok_or(FunctionErrorKind::WrongArgument(format!(
                 "the given address and port do not correspond to a valid address: {ip}:{port}"
-            )),
-        )?;
+            )))?;
```
Tehforsch, line 181: See above

rust/nasl-builtin-network/src/socket.rs
```diff
@@ -336,14 +356,23 @@ impl NaslSockets {
...
+    ) -> Result<NaslValue, FunctionErrorKind> {
+        let min = if let Some(min) = min {
+            if min < 0 {
+                len
+            } else {
+                min as usize
+            }
+        } else {
+            len
+        };
```
Tehforsch, line 375: I think the previous version of this logic is easier to read:
```rust
let min = min
            .map(|min| if min < 0 { len } else { min as usize })
            .unwrap_or(len);
```

Tehforsch (changes requested):
Requested a few minor cosmetic changes

Tehforsch (approved):
Thanks!

PR #1698, opened 2024-08-12
Async interpreter

Tehforsch:
**What**:
This PR implements three major new features:
- Adds a scanner type for the rust NASL interpreter to `openvasd`
- Adds storage functionality for `openvasd`
- Makes the NASL interpreter async.

Jira: SC-1071

## Structure of Scanner Code
This PR adds the `Scanner` type in the `nasl_interpreter` crate. (It would be nicer to have this in a different crate but I don't want to add a new one. Maybe in the future if we ever merge some of the crates, this can be moved/renamed. ) The `Scanner` is the single instance managing all scans during a run with `Openvasd` scanner type. To do so, it starts a number of `RunningScan`s, each of which is responsible for a single `Scan`. 
The `RunningScan` is responsible for computing the execution plan for this particular `Scan` and running it to completion. It also takes care of controlling the status of the scan and stopping it if necessary. Internally, it runs code via the `ScanRunner` (sorry, I ran out of names) which is responsible for performing all VTs in all stages on each target host. It is also responsible for sticking to the scheduling requirements. I suspect that this part of the code is where we should be able to easily improve performance by making the scripts run concurrently.
Finally, for a given VT and a given Host, the VT is then run to completion using the `VTRunner`.

## Tests
Since all of our NASL test employed the `Iterator` impl on `CodeInterpreter` which cannot exist anymore (since the `CodeInterpreter` is now async), they all needed to be changed. I took this as an opportunity to move the tests to a more unified structure, so that it will be easier to 1. add new tests and 2. change how the tests are done internally without having to make changes all over the code base.

Here is an example how the tests work now:
```rust
let mut t = TestBuilder::default();
t.run("l = [1,2,3,4,5];");
t.ok("max_index(l);", 5);
```
The `TestBuilder` struct can be used to set up the proper `Context`, `Storage` and function `Loader` and has a number of utility functions to perform common tests.
Lines of code can be added to the `TestBuilder` one-by-one, along with certain requirements on their results (For example, the `t.ok("max_index(l)", 5)` above will check that the result of the `max_index(l)` call is `NaslValue::Number(5)`.). At the end of the test, the `TestBuilder` will automatically check that all of the requirements are fulfilled. Even though lines are added individually, the `TestBuilder` keeps track of the state of the `Context` during execution, so tests that require multiple lines to set up (like the one above) will still work as intended.

To test that the correct `Err` variant is returned in certain lines, the `check_err_matches!` macro can be used as before, by passing it the `TestBuilder` as an argument:

```rust
check_err_matches!(
    t,
    r#"typeof(23,76);"#,
    FunctionErrorKind::TrailingPositionalArguments { .. }
);
```

To quickly check individual lines, `check_ok` still exists.

## Function executor
Previously, every set of NASL functions was added to the standard library by implementing the `NaslFunctionExecutor` trait for a struct (that might be either a marker ZST or a proper stateful struct). This approach was difficult to incorporate with the new `async` interpreter where NASL functions might be both sync or async, since async functions can't really be referred to by a simple type like `fn(&Context, &Register) -> NaslResult` like sync functions can.
Instead, the executor part of the `Context` is now an actual struct: `Executor`. This internally tracks all the registered functions and keeps all the necessary state (such as open ssh connections) around. Moreover, it provides the same functionality that `NaslFunctionExecutor` provided previously (namely executing functions by name).

The `Executor` can store multiple `FunctionSet`s which are collections as functions. A new `FunctionSet` can be created as such:
```rust
pub struct Array;
function_set! {
    Array,
    sync_stateless,
    (
        make_array,
        make_list,
        (nasl_sort, "sort"),
        keys,
        max_index,
    )
}
```
which automatically implements the `IntoFunctionSet` trait on `Array` and adds the given 5 functions. The functions will automatically be entered with their own name unless specified otherwise (as for `nasl_sort` here). This replaces the `lookup` functions we had to write before. 

At this point, NASL functions come in four flavors: 

1. `async_stateful` (for `async fn(&S, &Register, &Context)`)
2. `sync_stateful` (for `fn(&S, &Register, &Context)`)
3. `async_stateless` (for `async fn(&Register, &Context)`)
4. `sync_stateless` (for `fn(&Register, &Context)`)

At the moment, the user (i.e. the person using the `function_set!` macro) unfortunately needs to specify what flavour of function the set contains. (This is because the rust compiler does not know that something implementing `Fn(A, B) -> Z` will never also implement `Fn(A, B, C) -> Z`, so that implementing a common trait for all four flavours is impossible since the compiler will think that we implemented the trait multiple times for a single object. Unfortunately, specialization is not yet a thing in Rust.).

In the future, I think it should be possible to do something very cool with the `Executor`: Previously, we only got read access to the state (i.e. to the SSH handles), so we had to employ interior mutability in order to work with this state if we ever needed mutable access (in order to add a new SSH handle to our list of handles, for example). Since all the states are now managed by a `StatefulFunctionSet`, and we can tell whether a given function needs mutable access or not by its signature, I think we should be able to use an external RwLock and improve this situation with some careful logic. Concretely, this could mean having `Vec<Arc<Mutex<SSHSession>>>` instead of `Arc<Mutex<Vec<SSHSession>>>` which we previously needed, which would enable tons of concurrency that wasn't previously possible. This is pretty futuristic at this point though.

## Regressions
- Removed ability to have a generic reader in `HashSumLoader`. This was done for simplicity and since we currently don't use this ability, however it can be reverted if necessary by adding `Send` bounds on the trait object.
- Removed the ability to have a generic `NaslFunctionExecutor` and replaced it with a `Executor` which is a concrete struct. We still have the ability to only register any subset of functions that we want, so I don't think this is a true regression, but I'm writing it down for completeness' sake.
- Previously, the `NaslFunctionExecutor` trait had a `nasl_fn_cache_clear` method. This wasn't really called properly anywhere other than in `scannerctl` and wasn't handled properly in the `std` library either. I don't think there are currently any instances that needed this functionality, since RAII should take care of it. However, if we ever need this, it is very easy to add back in again.
- I had to revamp how `HostInfo` works to clean up the logic and make it possible to deal with asynchronous scans. First of all, I noticed that `HostInfo` internally stored something which I believe was supposed to be some kind of percentage progress. However, this progress was never actually updated or treated in any way. I removed this percentage value for now. This has an impact on how the update of a `HostInfo` works when a new result comes in and `scanner.do_addition()` returns true. I don't know what the expected behavior is there and since all tests still pass, I will just leave this as it is for now. Secondly, the new `HostInfo` stores the number of leftover VTs for each target host. In this way, it can properly track which hosts still have running VTs and which ones are finished. This was necessary because we now start multiple VTs at the same time, whereas previously they were done one by one.

**TODO**:
- [x] Fill in long list of todos
- [x] Add documentation for public methods
- [x] Address every new TODO
- [x] Reactivate commented tests
- [x] Fix conventional commits
- [x] Check if `RunningScan` needs to exist. Can it be replaced by `ScanRunner` altogether?
- [x] Rebase / merge
- [x] Split `executor.rs` in `nasl_builtin_utils`
- [x] Think about stateful/stateless sets again. Can we really not get rid of this? Stateless ones really only existed to be able to merge them without running into issues and because the functions don't have a self parameter.
- [x] Fix `HostInfo` mess
- [x] Improve `run` on `ScanRunner`.
- [x] Fix `unwrap` call in `ScanRunner::new`.

Before merging/closing:
- [ ] Add Jira ticket for the possibility of allowing `Executor` to take `&mut State`.
- [ ] Add Jira ticket for proper concurrent execution of scripts
- [ ] Add Jira ticket for builtin libraries that can now simply be async
- [ ] Add Jira ticket for TODO in Vt runner


rust/feed/src/update/mod.rs
```diff
@@ -154,30 +166,46 @@ where
...
-    R: Read + 'a,
-{
-    type Item = Result<String, Error>;
+    /// Run the feed update and log each result with the
+    /// given log level. If an error occurs, return it.
+    pub async fn perform_update(self, log_level: Level) -> Result<(), Error> {
+        let results = self.stream().collect::<Vec<_>>().await;
+        for result in results.iter() {
+            match log_level {
+                Level::TRACE => trace!(?result),
```
nichtsfrei, line 182: what's the reasoning for adding a log level into this function? I'd guess you only need the result of it in tracing?

rust/feed/src/update/mod.rs
```diff
@@ -154,30 +166,46 @@ where
...
+                Level::DEBUG => debug!(?result),
+                Level::WARN => warn!(?result),
+                Level::ERROR => error!(?result),
+                _ => todo!(),
+            }
+        }
+        results.into_iter().collect::<Result<Vec<_>, Error>>()?;
```
nichtsfrei, line 189: couldn't that be done in the for loop immediately?

rust/feed/src/update/mod.rs
Tehforsch: I was trying to keep the same behavior as before and in scannerctl we called this with `WARN` log level. Would happily remove this though.

rust/feed/src/update/mod.rs
Tehforsch: For some reason I thought we'd want to log all results before returning a potential error, but the collect was ugly anyways. Fixed to immediately return errors as they occur

rust/feed/src/update/mod.rs
nichtsfrei: please do, the WARN log level was by accident.

Tehforsch: Done

nichtsfrei:
we forgot to do that when introducing the openvas scanner type as well, but I think it would be helpful to explain the possible scanner types within `rust/examples/openvasd/config.example.toml` `scanner#type`

rust/infisto/src/serde.rs
```diff
@@ -44,7 +52,11 @@ where
     type Error = Error;
 
     fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
-        match rmp_serde::from_slice(&value) {
+        // match serde_json::from_slice(&value) {
```
nichtsfrei, line 55: please remove.

rust/models/src/host_info.rs
```diff
@@ -4,29 +4,87 @@
...
+    alive: u64,
+    queued: u64,
+    finished: u64,
+    // Hosts that are currently being scanned. The second entry is the number of
+    // remaining VTs for this host.
+    scanning: HashMap<String, usize>,
+}
+
+impl HostInfo {
+    pub fn new(all: u64, excluded: u64, dead: u64, alive: u64, queued: u64, finished: u64) -> Self {
```
nichtsfrei, line 28: feel free to ignore:

I know that a lot of people prefer constructor methods over a constructor but for this amount of parameter I would prefer reading:
```
HostInfo { 
    all: 1,
    excluded: 1,
    .. Default::default(),
}
```
over 
```
 HostInfor::new(1, 1, 0, 0, 0, 0)
```

rust/nasl-builtin-cryptographic/tests/hmac.rs
```diff
@@ -5,110 +5,54 @@
...
-        HMAC_MD5(key: "my_shared?key", data: "so much wow");
-        "#;
-        let register = Register::default();
-        let binding = ContextFactory::default();
-        let context = binding.build(Default::default(), Default::default());
-        let mut parser = CodeInterpreter::new(code, register, &context);
-        assert_eq!(
-            parser.next(),
-            Some(Ok("815292959633f0e63666d90d6f47cb79".into()))
+        check_ok(
```
nichtsfrei, line 19: I think the name for the macro `check_ok` is a bit too vague as it does actually spin up a nasl-interpreter, runs the given code and verifies the Result, that's a bit more then just "check if output is Ok" which I would guess based on the name.

rust/nasl-interpreter/src/scanner/running_scan.rs
```diff
@@ -0,0 +1,297 @@
...
+            &self.scan,
+        )
+        .map_err(make_scheduling_error)
+    }
+
+    async fn run_to_completion<'a>(&self, runner: ScanRunner<'a, S>) -> Phase {
+        let mut end_phase = Phase::Succeeded;
+        let mut stream = Box::pin(runner.stream());
+        while let Some(it) = stream.next().await {
+            // TODO: check for error and abort, we need to keep track of the state
```
nichtsfrei, line 116: isn't that already implemented?

rust/nasl-interpreter/src/scanner/running_scan.rs
```diff
@@ -0,0 +1,297 @@
+use std::{
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc, RwLock,
```
nichtsfrei, line 4: shouldn't we use tokio RwLock when we run in async contexts?

rust/nasl-interpreter/tests/local_var.rs
```diff
@@ -6,31 +6,31 @@
...
-            .map(|stmt| match stmt {
-                Ok(stmt) => interpreter.retry_resolve_next(&stmt, 1),
-                Err(r) => Err(InterpretError::from(r)),
-            })
-            .last()
-            // for the case of NaslValue that returns nothing
-            .unwrap_or(Ok(NaslValue::Exit(0)));
-        assert_eq!(results, Ok(NaslValue::Number(1)));
-    }
+    //     use nasl_builtin_std::ContextFactory;
```
nichtsfrei, line 9: It would make sense to reactivate those tests as they actually are regression tests

rust/nasl-builtin-cryptographic/tests/aes_cbc.rs
```diff
@@ -5,120 +5,71 @@
...
-        assert_eq!(
-            parser.next(),
-            Some(Ok(NaslValue::Data(
-                decode_hex("275cfc0413d8ccb70513c3859b1d0f72").unwrap()
-            )))
+        let mut t = TestBuilder::default();
+        t.run(r#"key = hexstr_to_data("000000000000000000000000000000000000000000000000");"#);
+        t.run(r#"data = hexstr_to_data("1b077a6af4b7f98229de786d7516b639");"#);
+        t.run(r#"iv = hexstr_to_data("00000000000000000000000000000000");"#);
+        t.ok(
```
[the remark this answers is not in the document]
nichtsfrei: Therefore I'm wondering if it would be plausible to remove the Default implementation of `let mut t = TestBuilder::default();` and replace it with a macro additionally adding file!() and line!() into TestBuilder. Although it would still shadow the exact line of the failing test at least someone could immediately figure out where the TestBuilder got created and with the help of the printed nasl code immediately find it.

nichtsfrei:
Very demure, very mindful.

Actually it is impressive, I have minor feedback for the TestBuilder as it would be nice to store somewhere where it got created and print that additionally to the nasl code line so that it is easier to figure out the position of that failing test.

I don't like the name `check_ok!` as a macro name as it hides its actual complexity.

And it would be great if you reenable [rust/nasl-interpreter/tests/local_var.rs](https://github.com/greenbone/openvas-scanner/pull/1698/files/cc938002f11009a30f80639d7f1443953ca72a1c#diff-95ed8183612c78b095c9fd55950f34601037c802088730134dab096d1545e8a6)

The rest can be ignored.

rust/nasl-builtin-cryptographic/tests/aes_cbc.rs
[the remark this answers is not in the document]
Tehforsch: >  Therefore I'm wondering if it would be plausible to remove the Default implementation of let mut t = TestBuilder::default(); and replace it with a macro additionally adding file!() and line!() into TestBuilder. Although it would still shadow the exact line of the failing test at least someone could immediately figure out where the TestBuilder got created and with the help of the printed nasl code immediately find it.

That's a really nice idea. I'll look into it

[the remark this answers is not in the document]
Tehforsch: Found a really awesome mechanism in rust to do this even more cleanly:

https://rustc-dev-guide.rust-lang.org/backend/implicit-caller-location.html

Using this, the new error messages point directly to the line of code adding the line of NASL code that fails:

```
Mismatch at nasl-builtin-regex/tests/regex.rs:137:11.
In code "eregmatch(string: string, pattern: "Bar|true", find_all: TRUE);":
Expected: Ok(Array([String("Baz"), String("true")]))
Found:    Ok(Array([String("Bar"), String("true")]))
```

rust/nasl-builtin-cryptographic/tests/hmac.rs
Tehforsch: Suggestion for a more descriptive name: `check_code_result`. Is that ok?

rust/nasl-interpreter/src/scanner/running_scan.rs
Tehforsch: oh - good catch. we should!

rust/nasl-interpreter/src/scanner/vt_runner.rs
```diff
@@ -0,0 +1,237 @@
...
+                result_err(e)
+            }
+        };
+        match result {
+            None => Ok(()),
+            Some(x) => Err(x),
+        }
+    }
+
+    fn check_keys(&self, vt: &storage::item::Nvt) -> Result<(), ScriptResultKind> {
```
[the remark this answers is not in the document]
Tehforsch: I'm not sure what you mean - should this become async? If that's what you mean then we would probably need to make the retriever async, which will be a bit of work. In that case I suggest adding a task for it for the future.

nichtsfrei (approved):

PR #1699, opened 2024-08-14
Fix: Never type fallback change

jjnicola:
**What**:
Fix: Never type fallback change
Jira: SC-1127

**Why**:
Never type (!) to any type (`never-to-any`) coercions fall back to never type (!) rather than to unit type (()).

**How**:
Replace `stable` with `nightly` in the `rust-toolchain` file and run `cargo build`

**Checklist**:

- [ ] Tests


Tehforsch (approved):
If anyone is as confused about all of this as me (in particular, why the type is inferred to be `!` in the first place), see

https://github.com/rust-lang/rust/issues/51125

for a tiny bit of clarity.

PR #1703, opened 2024-08-22
Add: nasl builtin functions for kb manipulation

jjnicola:
**What**:
Add nasl builtin functions for kb manipulation:
- replace_kb_item()
- get_kb_list()

Also, use nasl_function builtin macro in the existing function in this crate.

SC-1141

**Why**:
missing functions

**How**:

**Checklist**:

- [x] Tests

### DIFF
--- a/rust/nasl-builtin-knowledge-base/src/lib.rs
+++ b/rust/nasl-builtin-knowledge-base/src/lib.rs
@@ -11,8 +11,8 @@ use nasl_builtin_utils::{Context, Register};
 use nasl_function_proc_macro::nasl_function;
 use nasl_syntax::NaslValue;
 
-/// NASL function to set a value under item in a knowledge base
-/// If the already exists, it will be remove first to avoid duplications.
+/// NASL function to set a value under name in a knowledge base
+/// Only pushes unique values for the given name.
 #[nasl_function(named(name, value, expires))]
 fn set_kb_item(
     name: NaslValue,
@@ -68,7 +68,7 @@ fn get_kb_item(key: &str, c: &Context) -> Result<NaslValue, FunctionErrorKind> {
         .map_err(|e| e.into())
 }
 
-/// NASL function to replace an item in a KB.
+/// NASL function to replace a kb list
 #[nasl_function(named(name, value, expires))]
 fn replace_kb_item(
     name: NaslValue,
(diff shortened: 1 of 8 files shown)


rust/storage/src/lib.rs
```diff
@@ -366,6 +374,19 @@ impl DefaultDispatcher {
         Ok(())
     }
 
+    fn replace_kb(&self, scan_id: &str, kb: Kb) -> Result<(), StorageError> {
+        let mut data = self.kbs.as_ref().write()?;
+        if let Some(scan_entry) = data.get_mut(scan_id) {
+            scan_entry.remove(&kb.key);
+            scan_entry.insert(kb.key.clone(), vec![kb]);
```
Kraemii, line 381: The call to remove is not necessary, as insert replaces the value automatically and returns the replaced value. I got two solutions:
```suggestion
            if let Some(kb_entry) = scan_entry.get_mut(&kb.key) {
                *kb_entry = vec![kb];
            } else {
                scan_entry.insert(kb.key.clone(), vec![kb]);
            }
```
```suggestion
            scan_entry.insert(kb.key.clone(), vec![kb]);
```
the second one seems a lot cleaner, but is slightly slower, as we do not need the `clone` operation in the first one. I tested a little bit and if the replace function always replaces a value instead of inserting a new one, the first solution is always about 10% faster.

rust/nasl-builtin-knowledge-base/src/lib.rs
```diff
@@ -46,33 +52,66 @@ fn set_kb_item(register: &Register, c: &Context) -> Result<NaslValue, FunctionEr
...
+                key: name.to_string(),
+                value: value.clone().as_primitive(),
+                expire: None,
+            }),
+        )
+        .map(|_| NaslValue::Null)
+        .map_err(|e| e.into())
+}
+
+/// NASL function to replace an item in a KB.
```
Kraemii, line 91:
```suggestion
/// NASL function retrieve a kb list
```

rust/nasl-builtin-knowledge-base/src/lib.rs
```diff
@@ -4,26 +4,32 @@
...
-/// NASL function to set a knowledge base
-fn set_kb_item(register: &Register, c: &Context) -> Result<NaslValue, FunctionErrorKind> {
-    let name = get_named_parameter(register, "name", true)?;
-    let value = get_named_parameter(register, "value", true)?;
-    let expires = match get_named_parameter(register, "expires", false) {
-        Ok(NaslValue::Number(x)) => Some(*x),
-        Ok(NaslValue::Exit(0)) => None,
-        Ok(x) => {
+/// NASL function to set a value under item in a knowledge base
+/// If the already exists, it will be remove first to avoid duplications.
```
Kraemii, line 15:
```suggestion
/// NASL function to set a value under name in a knowledge base.
/// If the name already exists, the value is added to its list.
```
Not quite sure about that, but I guess we just push the value to the existing entry.

PR #1704, opened 2024-08-26
Add: nasl built regex library, regex nasl functions and tests.

jjnicola:
**What**:
Add nasl built regex library, regex nasl functions and tests

Jira: SC-1143
Jira: SC-1144

**Why**:

**How**:
Since the output of display() differs among  C and Rust implementations, use the following script for debugging with `gdb` and `rust-gdb` respectively. Also unit test where added/updated.
``` 
#split 
string = "a\nb\nc";
a = split(string);
display("1. " + a[0]);
display("2. " + a[1]);
display("3. " + a[2]);

string = 'hola\r\n1234';
a =split(string, sep: "\r");
display(a[0]);

#replace
display(str_replace(string: string, find: '\r\n', replace: "  --  "));

#strstr
display(strstr("hola mundo", "mundo"));
display(strstr("hola\nmundo", "mundo"));
display(strstr('hola\nmundo', "mundo"));
display(strstr('hola\nmundo', "\nmundo"));
display(strstr("hola\nmundo", 'mundo'));

#int
display(int(43));

#insstr
display(insstr("foo bar", "123456789", 4, 5));
display(insstr("foo\nbar", "123456789", 4, 5));
display(insstr('foo\nbar', "123456789", 4, 5));

## token
strstr('Hello \\\'you\\\'!', "aa");
display('Hello \'you\'!');
strstr("Hello \'you\'!", "aa");
display("Hello \'you\'!");

## match
display(match(string:"aaa\nb", pattern: "*a\nb", icase: FALSE ));
display(match(string:'aaa\nb', pattern: '*a\nb', icase: FALSE ));

## str_to_int
display(int("62"));
display(int('62'));

# Ord
display(ord("abcd"));
display(ord("\nabcd"));
display(ord('abcd'));
display(ord(1));

#decode_hex
display("\n\n####  DECODE HEX ");
display(hexstr("4bb3c4a4f893ad8c9bdc833c325d62b3"));
display(hexstr("4bb3c4a4f893ad8c9bdc833c325d62b3"));
display(hexstr('4bb3c4a4f893ad8c9bdc833c325d62b3'));
display(hexstr('4bb3c4a4f893ad8c9bdc833c325d62b3'));

#strlen
display(strlen('hallo\n'));
display(strlen("hallo\n"));


a = hexstr_to_data("aa");
display(hexstr(a));
display (data_to_hexstr(a));


#chomp
display(chomp("abcd\n"));
display(chomp('abcd\n'));

## stridx
display(stridx("abcd\n1234", "d\n"));
display(stridx('abcd\n1234', 'd\n'));
```

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/nasl-builtin-regex/Cargo.toml
+++ b/rust/nasl-builtin-regex/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "nasl-builtin-regex"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+nasl-builtin-utils = {path = "../nasl-builtin-utils"}
+nasl-syntax = { path = "../nasl-syntax" }
+nasl-function-proc-macro = {path = "../nasl-function-proc-macro"}
+storage = {path = "../storage"}
+lazy-regex = "3.1.0"
+regex = "1.10.6"
+
+[dev-dependencies]
+nasl-interpreter = { path = "../nasl-interpreter" }
+nasl-builtin-std = {path = "../nasl-builtin-std"}
--- a/rust/nasl-builtin-regex/Cargo.toml
+++ b/rust/nasl-builtin-regex/Cargo.toml
@@ -10,7 +10,6 @@ nasl-builtin-utils = {path = "../nasl-builtin-utils"}
 nasl-syntax = { path = "../nasl-syntax" }
 nasl-function-proc-macro = {path = "../nasl-function-proc-macro"}
 storage = {path = "../storage"}
-lazy-regex = "3.1.0"
 regex = "1.10.6"
 
 [dev-dependencies]
--- a/rust/nasl-builtin-regex/src/lib.rs
+++ b/rust/nasl-builtin-regex/src/lib.rs
@@ -10,7 +10,7 @@ use regex::RegexBuilder;
 
 /// Matches a string against a regular expression.
 /// - string  String to search the pattern in
-/// - pattern the patern that should be matched
+/// - pattern the pattern that should be matched
 /// - icase   case insensitive flag
 /// - rnul    replace the null char in the string. Default TRUE.
 /// - multiline Is FALSE by default (string is truncated at the first
@@ -66,7 +66,7 @@ fn ereg(
 
 /// Search for a pattern in a string and replace it.
 /// - string  String to search the pattern in
-/// - pattern patern to search in the string for
+/// - pattern pattern to search in the string for
 /// - replace string to replace the pattern with
 /// - icase   case insensitive flag
 /// - rnul    replace the null char in the string. Default TRUE.
@@ -116,7 +116,7 @@ fn ereg_replace(
 /// Looks for a pattern in a string, line by line.
 ///
 /// - string  String to search the pattern in
-/// - pattern the patern that should be matched
+/// - pattern the pattern that should be matched
 /// - icase   case insensitive flag
 /// - rnul    replace the null char in the string. Default TRUE.
 ///
@@ -168,7 +168,7 @@ fn egrep(
 ///
 /// - pattern An regex pattern
 /// - string A string
-/// - icase Boolean, for case sensitve
+/// - icase Boolean, for case sensitive
 /// - find_all Boolean, to find all matches
 /// - rnul replace the null char in the string. Default TRUE.
 ///
--- a/rust/nasl-builtin-regex/src/lib.rs
+++ b/rust/nasl-builtin-regex/src/lib.rs
@@ -6,7 +6,32 @@ use nasl_builtin_utils::{error::FunctionErrorKind, NaslFunction};
 use nasl_builtin_utils::{Context, Register};
 use nasl_function_proc_macro::nasl_function;
 use nasl_syntax::NaslValue;
-use regex::RegexBuilder;
+use regex::{Regex, RegexBuilder};
+
+fn parse_search_string(mut s: &str, rnul: bool, multiline: bool) -> &str {
+    if !rnul {
+        s = s.split('\0').next().unwrap();
+    }
+    if !multiline {
+        s = s.split('\n').next().unwrap();
+    }
+
+    s
+}
+
+fn make_regex(pattern: &str, icase: bool, multiline: bool) -> Result<Regex, FunctionErrorKind> {
+    match RegexBuilder::new(pattern.to_string().as_str())
+        .case_insensitive(icase)
+        .multi_line(multiline)
+        .build()
+    {
+        Ok(re) => Ok(re),
+        Err(e) => Err(FunctionErrorKind::Dirty(format!(
+            " Error building regular expression pattern: {}",
+            e
+        ))),
+    }
+}
 
 /// Matches a string against a regular expression.
 /// - string  String to search the pattern in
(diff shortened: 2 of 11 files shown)


rust/nasl-builtin-regex/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+///                “end of line”), and can be set to TRUE for multiline search.
+/// Return true if matches, false otherwise
+#[nasl_function(named(string, pattern, icase, rnul, multiline))]
+fn ereg(
+    string: NaslValue,
+    pattern: NaslValue,
+    icase: Option<bool>,
+    rnul: Option<bool>,
+    multiline: Option<bool>,
+) -> Result<NaslValue, FunctionErrorKind> {
```
Tehforsch, line 26: You can change the return type to `Result<bool, FunctionErrorKind>` and simply return a bool.

rust/nasl-builtin-regex/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+        (str_aux, _) = str_aux.split_at(s);
+    }
+
+    if !multiline {
+        let s = match str_aux.chars().position(|c| c == '\n') {
+            Some(i) => i,
+            None => str_aux.len(),
+        };
+        (str_aux, _) = str_aux.split_at(s);
+    }
```
Tehforsch, line 48:
```suggestion
    if !rnul {
        str_aux = str_aux.split('\0').next().unwrap();
    }
    if !multiline {
        str_aux = str_aux.split('\n').next().unwrap();
    }
```

Tehforsch: Also, this logic appears many times throughout the code. Please extract it into a function:

```rust
fn parse_search_string(s: &str, rnul: bool, multiline: bool) -> &str {
    ...
}
```

rust/nasl-builtin-regex/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+///
+/// Return the new string with the pattern replaced with replace.
+#[nasl_function(named(string, pattern, replace, icase, rnul))]
+fn ereg_replace(
+    string: NaslValue,
+    pattern: NaslValue,
+    replace: NaslValue,
+    icase: Option<bool>,
+    rnul: Option<bool>,
+) -> Result<NaslValue, FunctionErrorKind> {
```
Tehforsch, line 82: You can change the return type to `String`

rust/nasl-builtin-regex/src/lib.rs
```diff
@@ -0,0 +1,256 @@
...
+        .build()
+    {
+        Ok(re) => re,
+        Err(e) => {
+            return Err(FunctionErrorKind::Dirty(format!(
+                " Error building regular expression pattern: {}",
+                e
+            )));
+        }
+    };
```
Tehforsch, line 108: Similar logic also appears many times throughout the code. Please extract it into a function so that it becomes just

```rust
fn make_regex(pattern: &str, icase: bool, multiline: bool) -> Result<Regex, FunctionErrorKind> {
    ...
}
```

which can then simply be called as
```
make_regex(pattern, icase, multline)?
```

Tehforsch (changes requested):
In addition to the comments, I think we should consider changing the `FromNaslValue` impl for String to make a lot of the changes to nasl-builtin-string unnecessary.

rust/nasl-builtin-regex/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+nasl-builtin-utils = {path = "../nasl-builtin-utils"}
+nasl-syntax = { path = "../nasl-syntax" }
+nasl-function-proc-macro = {path = "../nasl-function-proc-macro"}
+storage = {path = "../storage"}
+lazy-regex = "3.1.0"
```
Tehforsch, line 13: Remove? As far as I can tell this is never used within `nasl-builtin-regex`.

Tehforsch:
I added SC-1148 to address the string conversion behavior and resolved the corresponding conversations. All other suggestions still apply.

rust/nasl-builtin-regex/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+fn parse_search_string<'a>(s: &'a mut &str, rnul: bool, multiline: bool) -> &'a str {
+    if !rnul {
+        *s = s.split('\0').next().unwrap();
+    }
+    if !multiline {
+        *s = s.split('\n').next().unwrap();
+    }
+
+    s
+}
```
Tehforsch, line 20: This can be simplified by removing the double reference. This may be a bit confusing but what we want to do is really just change to which part of the original `String` the `&str` should point to, so we don't actually need a mutable reference here - instead we just return a new `&str` that points to the right place. This is not a costly operation since `&str` is a tiny type.

```rust
fn parse_search_string(mut s: &str, rnul: bool, multiline: bool) -> &str {
    if !rnul {
        s = s.split('\0').next().unwrap();
    }
    if !multiline {
        s = s.split('\n').next().unwrap();
    }

    s
}
```

Looks like a small difference, but it means that we don't need to pass `&mut &str` to this function, which simplifies code further downstream. Also since its just a single reference, the compiler can now tell what the lifetime of the returned reference is.

rust/nasl-builtin-regex/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+    let rnul = rnul.unwrap_or(true);
+    let multiline = multiline.unwrap_or(false);
+
+    let str_bind = string.to_string();
+    let mut str_aux = str_bind.as_str();
+
+    let str_aux = parse_search_string(&mut str_aux, rnul, multiline);
+
+    let re = make_regex(&pattern.to_string(), icase, multiline)?;
+    Ok(re.is_match(str_aux))
```
Tehforsch, line 62: With the change to `parse_search_string` above, we can simplify this even further and get rid of `str_aux` and `str_bind`:
```rust
    let string = string.to_string();
    let string = parse_search_string(&string, rnul, multiline);

    let re = make_regex(&pattern.to_string(), icase, multiline)?;
    Ok(re.is_match(string))
```
Doing it this way avoids the explicit call to `.as_str()` (which is needed only rarely). 

Small digression: The reason this works is that the compiler will see that we are calling a function (`parse_search_string`) with a `&String` argument, but the function takes a `&str`. What happens next is that the compiler checks the `Deref` impl of `String`. `String` implements the `Deref` trait with `Target = &str`, and so the compiler knows that it can turn a `&String` into a `&str`, which it will automatically do. This is a pretty confusing part of Rust but quite important for lots of convenient syntax. This magic is called deref coercion and happens automatically in function calls - see https://doc.rust-lang.org/std/ops/trait.Deref.html#deref-coercion

rust/nasl-builtin-regex/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    let str_bind = string.to_string();
+    let mut str_aux = str_bind.as_str();
+
+    let str_aux = parse_search_string(&mut str_aux, rnul, true);
+    let re = make_regex(&pattern.to_string(), icase, false)?;
+
+    let out = re
+        .replace_all(str_aux, replace.to_string().as_str())
+        .to_string();
```
Tehforsch, line 92: We can do the same thing as above here

rust/nasl-builtin-regex/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+
+    let str_bind = string.to_string();
+    let mut str_aux = str_bind.as_str();
+    let str_aux = parse_search_string(&mut str_aux, rnul, true);
+    let re = make_regex(&pattern.to_string(), icase, true)?;
+
+    let lines: Vec<&str> = str_aux
+        .split_inclusive('\n')
+        .filter(|l| re.is_match(l))
+        .collect();
```
Tehforsch, line 122: Also as above

rust/nasl-builtin-regex/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+    let matches = match find_all {
+        true => re
+            .find_iter(str_aux)
+            .map(|m| NaslValue::String(m.as_str().to_string()))
+            .collect(),
+        false => match re.find(str_aux) {
+            Some(s) => vec![NaslValue::String(s.as_str().to_string())],
+            None => vec![],
+        },
+    };
```
Tehforsch, line 165: Also as above

Tehforsch (approved):
Thanks - just a few minor comments.

I think we can significantly clean up the string implementations but I'll defer this until we decided on SC-1148.

Tehforsch (approved):
Thanks!

PR #1707, opened 2024-08-29
Fix: improve scannerctl for using redis as storage.

jjnicola:
**What**:
Check the current feed version and only loads it up if it is empty or outdated. Currently, it takes too long for running a single script, if scannerctl is called with the parameter --path.

Also, add the command line  option to pass the redis url.
Jira: SC-1145

**Why**:
It takes to long if --path argument is given. 

**How**:
`scannerctl execute script --redis=unix:///var/run/redis-openvas/redis.sock -p /home/jnicola/install/var/lib/openvas/plugins some_script.nasl`

**Checklist**:

- [ ] Tests


jjnicola:
close in favor of #1708

PR #1708, opened 2024-09-02
Fix: load the feed metadata from an existing json file

jjnicola:
**What**:
Fix: load the feed metadata from an existing json file
 Jira: SC-1145

**Why**:
 Currently, it takes too long for running a single script, if scannerctl is called with the parameter --path

**How**:
`scannerctl execute script  -p /home/jnicola/install/var/lib/openvas/plugins some_script.nasl`
vs passing the already existing feed json file generated with scannerctl feed transform 
`scannerctl execute script  -p /tmp/feed.json some_script.nasl`

**Checklist**:

- [ ] Tests


rust/scannerctl/src/interpret/mod.rs
```diff
@@ -193,8 +224,15 @@ pub fn run(
...
-            builder.storage(storage).build().run(script)
+            
+            if let Some(_jsonfile) = path.extension() {
+                let builder = RunBuilder::default().loader(create_loader_from_feed_json(&storage, path)?);
+                builder.storage(storage).build().run(script)
+            
+            } else {
+                let builder = RunBuilder::default().loader(create_fp_loader(&storage, path)?);
+                builder.storage(storage).build().run(script)
+            }
```
Tehforsch, line 235: We can minimize the size of the if expression here:

```rust

            let loader = if let Some(_jsonfile) = path.extension() {
                create_loader_from_feed_json(&storage, path)?
            } else {
                create_fp_loader(&storage, path)?
            };
            let builder = RunBuilder::default().loader(loader);
            builder.storage(storage).build().run(script)
```

Also, this doesn't actually check if the file is a json, right? So maybe we should do something like

```rust
            let is_json = path.extension().map(|ext| ext == "json").unwrap_or(false);
            let loader = if is_json {
                create_loader_from_feed_json(&storage, path)?
            } else {
                create_fp_loader(&storage, path)?
            };
```

rust/scannerctl/src/interpret/mod.rs
```diff
@@ -171,6 +171,37 @@ where
...
+        }
+    };
+
+    let mut buf = String::new();
+    if let Err(err) = file.read_to_string(&mut buf) {
+            return Err(CliError{
+                filename: path.to_owned().to_string_lossy().to_string(),
+                kind: CliErrorKind::LoadError(LoadError::Dirty(err.to_string()))
+            });
+        }
```
Tehforsch, line 194: I think this kind of error conversion logic really distracts from the actual thing going on. I initially thought this conversion should just become a `From<io::Error>` impl for `CliError` but I now see that this already exists (and converts to a `Corrupt` variant. I don't know if this is intended, but let's not touch that for now).

I personally would prefer a `load_error` constructor on `CliError` that takes an `io::Error`. If we then also use `fs::read_to_string` which does exactly what `File::open` and `File::read_to_string` do, but in a single function, we end up with

```rust
    let buf = fs::read_to_string(&path).map_err(|e| CliError::load_error(e, &path))?;
    let vts: Vec<storage::item::Nvt> = serde_json::from_str(&buf)?;
```
which I find much much easier to read.

The only thing thats needed is adding the `load_error` method on `CliError`:
```rust
    pub fn load_error(err: std::io::Error, path: &Path) -> Self {
        Self {
            filename: path.to_owned().to_string_lossy().to_string(),
            kind: CliErrorKind::LoadError(LoadError::Dirty(err.to_string())),
        }
    }
```

Tehforsch (changes requested):

jjnicola:
Hi @Tehforsch! thanks a lot for the detailed review. I really appreciate it.

rust/scannerctl/src/interpret/mod.rs
```diff
@@ -171,6 +174,23 @@ where
     Ok(result)
 }
 
+fn create_loader_from_feed_json(
+    store: &DefaultDispatcher,
+    path: PathBuf,
+) -> Result<FSPluginLoader<PathBuf>, CliError> {
+    let result = FSPluginLoader::new(path.clone());
+    let buf = fs::read_to_string(&path).map_err(|e| CliError::load_error(e, &path))?;
+    let vts: Vec<storage::item::Nvt> = serde_json::from_str(&buf)?;
```
nichtsfrei, line 183: The approach of loading the file as a string into heap, use that heap to serialize into a struct by using serde_json::from_str and then iterate to map it seems relatively expensive.

A better approach could be to use the file reader inside serde directly and may using non collecting approach to immediately map it into a map.

I'm going to provide an example later, writing it now so that I don't forget it.

rust/scannerctl/src/interpret/mod.rs
```diff
@@ -193,7 +213,15 @@ pub fn run(
         }
         (Db::InMemory, Some(path)) => {
             let storage = DefaultDispatcher::new(true);
-            let builder = RunBuilder::default().loader(create_fp_loader(&storage, path)?);
+
+            let is_json = path.extension().map(|ext| ext == "json").unwrap_or(false);
```
nichtsfrei, line 217: seems convoluted, going to provide an example later.

rust/storage/src/lib.rs
```diff
@@ -316,6 +316,12 @@ impl DefaultDispatcher {
         }
     }
 
+    /// Stores an already existing Vts structure.
+    pub fn set_vts(&self, vts: Vts) -> Result<(), StorageError> {
```
nichtsfrei, line 320: wouldn't it bet better to transform a given reader into Vts instead of letting the caller do that?

rust/scannerctl/src/interpret/mod.rs
```diff
@@ -171,6 +174,23 @@ where
     Ok(result)
 }
 
+fn create_loader_from_feed_json(
+    store: &DefaultDispatcher,
+    path: PathBuf,
+) -> Result<FSPluginLoader<PathBuf>, CliError> {
```
nichtsfrei, line 180: To return a FSPluginLoader doesn't make sense here, as the given path is a path to a json file and hence cannot be used as a FSPLuginLoader. An alternative would be either:
- return FSPluginLoader pointing to path.parent()
- don't return a FSPluginLoader and change interprete::mod::run method to call `set_vts` when it's a json and only otherwise return a FSPluginLoader.

nichtsfrei: I think I would rewrite `create_loader_from_feed_json` like:
```
fn create_loader_from_feed_json(
    store: &DefaultDispatcher,
    path: PathBuf,
) -> Result<FSPluginLoader<PathBuf>, CliError> {
    let file = File::open(&path)?;
    let mut reader = BufReader::new(file);
    let vts: Vec<storage::item::Nvt> = serde_json::from_reader(&mut reader)?;
    let all_vts = vts.into_iter().map(|v| (v.filename.clone(), v)).collect();

    store.set_vts(all_vts).map_err(|e| CliError {
        filename: path.to_owned().to_string_lossy().to_string(),
        kind: e.into(),
    })?;
    match path.parent() {
        None => Err(CliError {
            filename: path.to_str().map(|x| x.to_string()).unwrap_or_default(),
            kind: CliErrorKind::Corrupt("Does not have a parent".to_string()),
        }),
        Some(p) => Ok(FSPluginLoader::new(p.to_path_buf())),
    }
}
```

That way the FSPluginLoader shows to the parent of the json feed meta information which in most scenarios should be correct.However I think moving this functionality to the DefaultDispatcher and replace `set_vts` to `load_from_feed_json` using that logic there would be better as we could reconstruct the run method to use the feed path to construct the FSPluginLoader and it has a feed.json or so then it can load it without having it to parse each and every script but can fallback to it.

That way we could reuse this functionality when there is a proper feed that has the json in it later on.

rust/scannerctl/src/interpret/mod.rs
nichtsfrei: after testing serde_json::from_str and iterate on that is faster than creating a bufreader and use from_reader.

rust/scannerctl/src/interpret/mod.rs
nichtsfrei: convoluted was a bit dramatic, but I think:
```
            let is_json = path.extension().filter(|ext| *ext == "json").is_some();
```
is a bit easier to read, even when inlined in the if statement instead of a variable.

nichtsfrei:
Another proposal could be:

```diff
diff --git a/rust/scannerctl/src/interpret/mod.rs b/rust/scannerctl/src/interpret/mod.rs
index 0b303542..6bc27e65 100644
--- a/rust/scannerctl/src/interpret/mod.rs
+++ b/rust/scannerctl/src/interpret/mod.rs
@@ -156,29 +156,25 @@ fn create_redis_storage(
     redis_storage::CacheDispatcher::as_dispatcher(url, FEEDUPDATE_SELECTOR).unwrap()
 }

-fn create_fp_loader<S>(storage: &S, path: PathBuf) -> Result<FSPluginLoader<PathBuf>, CliError>
+fn load_feed_by_exec<S>(storage: &S, pl: &FSPluginLoader<PathBuf>) -> Result<(), CliError>
 where
     S: storage::Dispatcher,
 {
     // update feed with storage

     tracing::info!("loading feed. This may take a while.");
-    let result = FSPluginLoader::new(path);
-    let verifier = feed::HashSumNameLoader::sha256(&result)?;
-    let updater = feed::Update::init("scannerctl", 5, &result, storage, verifier);
+    let verifier = feed::HashSumNameLoader::sha256(pl)?;
+    let updater = feed::Update::init("scannerctl", 5, pl, storage, verifier);
     for u in updater {
-        tracing::warn!(updated=?u);
+        tracing::trace!(updated=?u);
         u?;
     }
     tracing::info!("loaded feed.");
-    Ok(result)
+    Ok(())
 }

-fn create_loader_from_feed_json(
-    store: &DefaultDispatcher,
-    path: PathBuf,
-) -> Result<FSPluginLoader<PathBuf>, CliError> {
-    let result = FSPluginLoader::new(path.clone());
+fn load_feed_by_json(store: &DefaultDispatcher, path: &PathBuf) -> Result<(), CliError> {
+    tracing::info!(path=?path, "loading feed via json. This may take a while.");
     let buf = fs::read_to_string(&path).map_err(|e| CliError::load_error(e, &path))?;
     let vts: Vec<storage::item::Nvt> = serde_json::from_str(&buf)?;
     let all_vts = vts.into_iter().map(|v| (v.filename.clone(), v)).collect();
@@ -187,8 +183,8 @@ fn create_loader_from_feed_json(
         filename: path.to_owned().to_string_lossy().to_string(),
         kind: e.into(),
     })?;
-
-    Ok(result)
+    tracing::info!("loaded feed.");
+    Ok(())
 }

 pub fn run(
@@ -208,18 +204,20 @@ pub fn run(
         (Db::InMemory, None) => builder.build().run(script),
         (Db::Redis(url), Some(path)) => {
             let storage = create_redis_storage(url);
-            let builder = RunBuilder::default().loader(create_fp_loader(&storage, path)?);
+            let loader = FSPluginLoader::new(path);
+            load_feed_by_exec(&storage, &loader)?;
+            let builder = RunBuilder::default().loader(loader);
             builder.storage(storage).build().run(script)
         }
         (Db::InMemory, Some(path)) => {
             let storage = DefaultDispatcher::new(true);
-
-            let is_json = path.extension().map(|ext| ext == "json").unwrap_or(false);
-            let loader = if is_json {
-                create_loader_from_feed_json(&storage, path)?
+            let guessed_feed_json = path.join("feed.json");
+            let loader = FSPluginLoader::new(path.clone());
+            if guessed_feed_json.exists() {
+                load_feed_by_json(&storage, &guessed_feed_json)?
             } else {
-                create_fp_loader(&storage, path)?
-            };
+                load_feed_by_exec(&storage, &loader)?
+            }

             let builder = RunBuilder::default().loader(loader);
             builder.storage(storage).build().run(script)
```

This way we don't need to change that much,  we just use the `feed.json` when it is available and if not we execute like previously.

That way the user doesn't have when to use which path with `-p` arguments, we could potentially introduce an additional arg for the use case that `feed.json` is not within path but elsewhere and the bug that the FSPluginLoader is incorrectly set is fixed.

nichtsfrei (approved):

rust/storage/src/lib.rs
jjnicola: Can be done within another PR

Tehforsch (approved):

PR #1709, opened 2024-09-02
Add: NASL bultin functions for isotime

Kraemii:
This includes

- isotime_add
- isotime_is_valid
- isotime_now
- isotime_print
- isotime_scan

SC-1146

Many cases contain unit tests. For testing the `isotime_now` function I used a small NASL script for validation:
```
display(isotime_now());
```
If found some results, that are different than in the c NASL implementation:
1. We now always allow giving the isotime format (`yyyymmddThhmmss`) and the readable form (`yyyy-mm-dd hh:mm:ss`) and always output in isotime format (except for `isotime_print`). Before it was not allowed for `isotime_add`
2. We now support all dates from year 0000 to 9999, which was a TODO in the old implementation. Before only dates from `1582-10-15 00:00:00` were supported
3. We now also support subtracting years, days and seconds when given a negative numbers within `isotime_add`
4. When adding a year to a date on a 29th of February the day was set to the 1st of March and is now set to the 28th of Februrary
5. The c implementation allows times like `20161231T235961` (61 seconds)

Would be nice to have feedback, if some of these cases should be adjusted to match the exact c implementation.

### DIFF
--- a/rust/nasl-builtin-isotime/src/lib.rs
+++ b/rust/nasl-builtin-isotime/src/lib.rs
@@ -1,0 +1,130 @@
+// SPDX-FileCopyrightText: 2023 Greenbone AG
+//
+// SPDX-License-Identifier: GPL-2.0-or-later WITH x11vnc-openssl-exception
+
+//! Defines NASL functions regarding isotime.
+
+use chrono::{Datelike, Months, NaiveDate, NaiveDateTime, TimeDelta};
+use nasl_builtin_utils::{function_set, FunctionErrorKind};
+use nasl_function_proc_macro::nasl_function;
+
+const ISOFORMAT: &str = "yyyymmddThhmmss";
+const READABLEFORMAT: &str = "yyyy-mm-dd hh:mm:ss";
+
+fn parse_isotime(time: &str) -> Option<NaiveDateTime> {
+    NaiveDateTime::parse_from_str(time, "%Y%m%dT%H%M%S").ok()
+}
+
+fn parse_readable_time(time: &str) -> Option<NaiveDateTime> {
+    if let Ok(time) = NaiveDateTime::parse_from_str(time, "%Y-%m-%d %H:%M:%S") {
+        return Some(time);
+    }
+    if let Ok(time) = NaiveDateTime::parse_from_str(time, "%Y-%m-%d %H:%M") {
+        return Some(time);
+    }
+    if let Some((date, hours)) = time.split_once(" ") {
+        if let Ok(date) = NaiveDate::parse_from_str(date, "%Y-%m-%d") {
+            if let Ok(hours) = hours.parse::<u32>() {
+                if let Some(time) = date.and_hms_opt(hours, 0, 0) {
+                    return Some(time);
+                }
+            }
+        }
+    }
+    if let Ok(date) = NaiveDate::parse_from_str(time, "%Y-%m-%d") {
+        // Cannot fail, since we add no time to the date
+        return Some(date.and_hms_opt(0, 0, 0).unwrap());
+    }
+
+    None
+}
+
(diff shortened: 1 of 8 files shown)


rust/nasl-builtin-isotime/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+        if let Ok(date) = NaiveDate::parse_from_str(date, "%Y-%m-%d") {
+            if let Ok(hours) = hours.parse::<u32>() {
+                if let Some(time) = date.and_hms_opt(hours, 0, 0) {
+                    return Some(time);
+                }
+            }
+        }
+    }
+    if let Ok(date) = NaiveDate::parse_from_str(time, "%Y-%m-%d") {
+        // Impossible to fail, as we add not time to the date
```
Tehforsch, line 35:
```suggestion
        // Cannot fail, since we add no time to the date
```

rust/nasl-builtin-isotime/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+use chrono::{Datelike, Months, NaiveDate, NaiveDateTime, TimeDelta};
+use nasl_builtin_utils::{Context, FunctionErrorKind, NaslFunction, Register};
+use nasl_function_proc_macro::nasl_function;
+
+const ISOFORMAT: &str = "yyyymmddThhmmss";
+const READABLEFORMAT: &str = "yyyy-mm-dd hh:mm:ss";
+
+fn parse_isotime(time: &str) -> Option<NaiveDateTime> {
+    NaiveDateTime::parse_from_str(time, "%Y%m%dT%H%M%S").ok()
+}
```
Tehforsch, line 16: Just a comment, no need to do anything:

I think it would be much cleaner if we could do
```rust
impl<'a> FromNaslValue<'a> for NaiveDateTime {
    fn from_nasl_value(value: &'a nasl_syntax::NaslValue) -> Result<Self, FunctionErrorKind> {
        let time_str = <&str>::from_nasl_value(value)?;
        parse_time(time_str)
    }
}
```
and then directly take `NaiveDateTime` in the functions. This has two benefits: It reduces the amount of times we need to call `parse_time` manually, but more importantly, it allows defining some of the NASL functions below as infallible (since they are only fallible due to the parsing of the isotime, which would otherwise be handled by `nasl_function`), which would be quite nice.

However, due to the current crate split situation, this isn't possible because of the orphan rule (`FromNaslValue` is not defined in `nasl-builtin-isotime`), requiring us to do this by defining a wrapper type (like `IsoTime(NaiveDateTime)`) which reduces the ergonomics by quite a lot.

Anyways, feel free to resolve this, because I don't think we can improve this situation in this PR.

Tehforsch (dismissed):
Looks good, just two minor comments.

rust/nasl-builtin-isotime/src/lib.rs
nichtsfrei: Would it be worth to resolve the crate split situation and move to a module approach?

Tehforsch (approved):

Tehforsch (approved):

rust/nasl-builtin-isotime/tests/isotime.rs
```diff
@@ -0,0 +1,121 @@
+// SPDX-FileCopyrightText: 2023 Greenbone AG
+//
+// SPDX-License-Identifier: GPL-2.0-or-later WITH x11vnc-openssl-exception
+#[cfg(test)]
+mod tests {
+    use nasl_interpreter::{check_err_matches, test_utils::check_ok, FunctionErrorKind};
```
Tehforsch, line 6: `check_ok` was renamed to `check_code_result`

Tehforsch (approved):

PR #1711, opened 2024-09-05
Initial proposal of a Dev container

nichtsfrei:
As more and more people are using VSCode on a non linux machine we would
like to make it easier to start developing on openvas as a whole.

With this and an extension `Dev Container` a developer can work within
an already setup environment.

Additionally it is designed so that someone on a linux machine can also
use it with Distrobox or with a direct mounted docker command.

For that the GID as well as UID can be set via build parameter for
convenience there is a Makefile creating the image with shared UID and
GID gatherred from `id`.

To build a new image for those purposes run `make build`.

Inside the container there a multiple helper integrated, run:
- start_redis -- to start a redis instance for openvas-scanner
- greenbone-nvt-sync -- to get a community feed
- build-openvas -- to rebuild openvas-scanner
- build-gvm-libs -- to rebuild gvm-libs

Additionally there are github specialized clone helper scripts, for an
example executing:

- github-clone greenbone/openvas-scanner
- github-clone greenbone/gvm-libs

will clone the openvas-scanner into
/workspaces/greenbone/openvas-scanner and /workspaces/greenbone/gvm-libs
respectively so that they can be immediately found by the build-
scripts.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -2,7 +2,21 @@ FROM rust:latest
 ARG UID=1000
 ARG GID=1000
 
-RUN apt-get update && apt-get install -y sudo git rsync pipx redis-server clangd
+RUN apt-get update && apt-get install -y \
+    sudo \ 
+    git \ 
+    rsync \ 
+    pipx \ 
+    redis-server \ 
+    clangd \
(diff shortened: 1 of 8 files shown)


nichtsfrei:
already merged through another PR.

PR #1715, opened 2024-09-06
Fix: set tls configuration in openvasd context builder.

jjnicola:
**What**:
Fix: set tls configuration in openvasd context builder.
SC-1154

**Why**:
It was starting only in http mode, since the tls certificate were not set, even if they were set in the config file.

**How**:
Set/unset TLS certs in the openvasd.toml configuration file and start openvasd

**Checklist**:

- [ ] Tests


Tehforsch (approved):
Thanks!

PR #1717, opened 2024-09-09
Single crate

Tehforsch:
Jira: SC-1150

**What**:

Change rust project structure to a main crate with most of the previous crates being modules of the main crate.

This PR doesn't make any functional changes.
Here is how the workspace structure changed:

- src
- crates/
- - smoketest
- - nasl-c-lib
- - nasl-function-proc-macro

The three crates are separate for the following reasons:
- smoketest has a custom makefile. I don't know why this is necessary, but didn't want to make any functional changes, so I left it as a subcrate
- nasl-c-lib has a custom `build.rs`. This could be ported to be a submodule too, but since I don't know whether it will stay around, I left it as a subcrate.
- `nasl-function-proc-macro` needs to be a subcrate, since proc macros can not be part of the main code base.

The main crate is now called `scannerlib` (happy for any input on how to name it instead). This name is only used inside of the binaries (`scannerctl`/`openvasd`) to import items from the library.

The `src` folder looks like:
- nasl
- - syntax
- - interpreter
- - utils
- - builtin
- - - host
- - - string
- - - misc
- - - ...
- storage
- - file
- - json
- - redis
- scanner
- scheduling
- feed
- feed_verifier
- models
- notus
- openvas
- openvasd
- osp
- plugin_scheduler
- scannerctl

There are some minor cosmetic changes besides the crate structure:
- Added a NASL prelude. This makes it easy to include the necessary items to begin writing nasl functions: `use crate::nasl::prelude::*;`. This automatically imports things such as `NaslValue`, `NaslResult`, `Context`, `Register`, `FromNaslValue`, `ToNaslResult`, `nasl_function` and `function_set!`.
- Added a NASL test prelude. Can be used in tests and automatically imports the `TestBuilder` and utility functions via `use crate::nasl::test_prelude::*`
- Added `allow(unused)` to `plugin_scheduler`. This wasn't previously needed but the new structure makes it apparent that this isn't currently used.
- Added `allow(unused)` to `osp::ScanCommand::Get` and removed `osp::connection::get_scan` which was unused.
- Moved `dep_graph` into `plugin_scheduler` module.
- For now, I moved most `readme.md` into their respective module and included them as a docstring for the module. We can change this in the future if we want.
- Currently, there are two crates that need two different versions simultaneously: `rustls-pemfile` and `generic-array`. I added two separate versions explicitly. We should open a ticket to update the code that requires the old one to a newer version.

**Why**:

- Compile times will improve
- Compiler errors will improve. This is more subtle, but in the current structure, compiler errors come in from one crate after the other. You first have to fix all errros in an upstream crate before you even see the downstream errors. This can be quite annoying when one is trying to make changes to an upstream crate, since one will initially underestimate the amount of impact that a change has on the rest of the codebase, which can result in bad design only becoming apparent once one has fixed the initial batch of errors, possibly resulting in wasted time.
- We will be able to implement traits from the crate on external type. This is more subtle but possibly the most impactful change. As an example that recently came up in a PR: It would be nice to implement `FromNaslValue` for `NaiveDateTime` (which is an external struct). However, we can't do it in the crate that needs it (nasl-builtin-isotime) because of the orphan rule. We also most likely don't want to do it in nasl-builtin-utils because that would require pulling in a dependency. This pattern is especially apparent in error handling (it would be nice to implement `From<DownstreamError>` for `UpstreamError` but we cannot currently do that.
- Better unused code warnings. Currently, all crates that export items need to make them public at crate level. The major problem with this is that public items will never be marked as dead code, even if theyre never used across the code base. Improving this situation could mean that we 1. get to delete unnecessary code or, even better 2. find out unused code that really should be used, indicating a bug.
- Reduction in complexity in dependency management. Previously, a fresh compile built ~470 dependencies, now we need ~370. This is because previously, we rebuilt lots of different versions of libraries unnecessarily. We also used to build experimental dependencies even if the `experimental` feature was not activated. This is not the case anymore.

**Checklist**:

- [x] Check whether CI works with new structure
- [x] Check number of tests
- [x] Check dependencies did not change.
- [x] Open ticket for `rustls_pemfile` / `generic_array`.
- [x] Check if ticket exists for `osp::ScanCommand::Get`
- [x] Add crate overview to `rust/README.md`


nichtsfrei:
I like this proposed workspace structure. 

I'm not sure if I find our toggle construct matching to that structure as it is now somewhere in the middle of a module and it seems a bit random now.

Also I wished that the README.md in `rust` explained the structure a bit more detailed. I know that we ignored explanation within a README.md a while but I'd guess that with that proposal those changes should be rather seldom.

Tehforsch:
> I'm not sure if I find our toggle construct matching to that structure as it is now somewhere in the middle of a module and it seems a bit random now.

Opened a Jira issue for this.

> Also I wished that the README.md in rust explained the structure a bit more detailed. I know that we ignored explanation within a README.md a while but I'd guess that with that proposal those changes should be rather seldom.

Added a brief overview at the end of `rust/README.md`.

Kraemii (approved):

PR #1719, opened 2024-09-12
Fix: scan progress for openvas scanner type.

jjnicola:
**What**:
- Scan progress for openvas scanner type:
  - SC-1161
  - SC-1159

**Why**:
it was always 0% until end of the scans. Openvas doesn't keep the scan progress. Therefore, the Openvasd scan progress must not be overwritten, but updated.

**How**:
Run a scan with openvas type and check status

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/models/src/host_info.rs
+++ b/rust/models/src/host_info.rs
@@ -14,6 +14,7 @@ pub struct HostInfoBuilder {
     pub alive: u64,
     pub queued: u64,
     pub finished: u64,
+    pub scanning: Option<HashMap<String, i32>>,
 }
 
 impl HostInfoBuilder {
@@ -25,7 +26,8 @@ impl HostInfoBuilder {
             alive: self.alive,
             queued: self.queued,
             finished: self.finished,
-            scanning: HashMap::new(),
+            scanning: self.scanning,
+            remaining_vts_per_host: HashMap::new(),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


rust/models/src/host_info.rs
```diff
@@ -43,34 +45,41 @@ pub struct HostInfo {
...
+    // scan progress. Required for Openvas Scanner type
+    #[cfg_attr(
+        feature = "serde_support",
+        serde(skip_serializing_if = "Option::is_none")
+    )]
+    scanning: Option<HashMap<String, i32>>,
     // Hosts that are currently being scanned. The second entry is the number of
     // remaining VTs for this host.
-    scanning: HashMap<String, usize>,
+    remaining_vts: HashMap<String, usize>,
```
nichtsfrei, line 57: The comment indicates that the values holds currently running scans, while the name indicates that those are vts that don't have run yet. Which one is it?

rust/openvasd/src/config.rs
```diff
@@ -597,8 +598,9 @@ impl Config {
         if let Some(path) = cmds.get_one::<PathBuf>("tls-client-certs") {
             config.tls.client_certs = Some(path.clone());
         }
-        if let Some(enable) = cmds.get_one::<bool>("enable-get-scans") {
-            config.endpoints.enable_get_scans = *enable;
+        if let Some(enable) = cmds.get_one::<String>("enable-get-scans") {
+            config.endpoints.enable_get_scans =
+                matches!(enable.to_lowercase().as_str(), "true" | "t" | "1");
```
nichtsfrei, line 603: can you explain why we need to have multiple ways to express 'true|false' within the cmd line?

rust/openvasd/src/config.rs
```diff
@@ -422,7 +422,8 @@ impl Config {
                 clap::Arg::new("enable-get-scans")
                     .env("ENABLE_GET_SCANS")
                     .long("enable-get-scans")
-                    .action(ArgAction::SetTrue)
+                    .num_args(0..=1)
```
nichtsfrei, line 425: why? what was wrong with `--enable-get-scans` enables `\scans` without having to add '1, true, ...' as additional values?

nichtsfrei:
The actual fixes are good, but I don't like that you are doing CLI changes. If you think those are necessary please do them in another PR as the discussion about those changes shadow the fixes.

jjnicola:
> The actual fixes are good, but I don't like that you are doing CLI changes. If you think those are necessary please do them in another PR as the discussion about those argument changes are shadowing the fixes.

I have just split the PR. CLI changes done within #1722

nichtsfrei (approved):

PR #1721, opened 2024-09-16
Emit error for wrong arg order in nasl_functions.

Tehforsch:
Jira: SC-1163.

I also fixed places where the arguments were not properly ordered. While doing so, I noticed that a few `nasl_functions` mention nonexistent args in the attrs of the macro. I turned this into a hard error, since it can lead to problems such as in `scanner_add_port`.

### DIFF
--- a/rust/crates/nasl-function-proc-macro/src/error.rs
+++ b/rust/crates/nasl-function-proc-macro/src/error.rs
@@ -48,7 +48,7 @@ impl Error {
                 "Specific type specified in receiver argument. Currently, only `&self` is supported."
             }
             ErrorKind::WrongArgumentOrder => {
-                "Argument in wrong position. Order of arguments should be: Positionals, Named, Context/Register"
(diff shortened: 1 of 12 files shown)


rust/crates/nasl-function-proc-macro/src/error.rs
```diff
@@ -42,6 +47,9 @@ impl Error {
             ErrorKind::TypedRefReceiverType => {
                 "Specific type specified in receiver argument. Currently, only `&self` is supported."
             }
+            ErrorKind::WrongArgumentOrder => {
+                "Argument in wrong position. Order of arguments should be: Positionals, Named, Context/Register"
```
Kraemii, line 51:
```suggestion
                "Argument in wrong position. Order of arguments should be: Context/Register, Positionals, Named"
```

Kraemii (changes requested):
Only a small change, for printing the error. The rest looks good to me

Kraemii (approved):

PR #1722, opened 2024-09-16
Fix: enable-get-scans

jjnicola:
**What**:
Don't overwrite the setting in the configuration file, if it is not passed via command line. 
Jira: SC-1165

**Why**:
The ArgAction::SetTrue will be set always set to `false` if the option is not given. This cause clap to overwrite the setting in the config file.

**How**:
Run openvasd with the option in the command line and in the configuration file. The configuration file is overwritten with Clap setting

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/src/openvasd/config.rs
+++ b/rust/src/openvasd/config.rs
@@ -422,8 +422,10 @@ impl Config {
                 clap::Arg::new("enable-get-scans")
                     .env("ENABLE_GET_SCANS")
                     .long("enable-get-scans")
-                    .action(ArgAction::SetTrue)
-                    .help("enable get scans endpoint"),
+                    .num_args(0..=1)
+                    .value_parser(clap::builder::BoolValueParser::new())
+                    .default_missing_value("true")
+                    .help("enable get scans endpoint. Default 'true'."),
             )
             .arg(
                 clap::Arg::new("api-key")


rust/openvasd/src/config.rs
```diff
@@ -597,8 +598,8 @@ impl Config {
         if let Some(path) = cmds.get_one::<PathBuf>("tls-client-certs") {
             config.tls.client_certs = Some(path.clone());
         }
-        if let Some(enable) = cmds.get_one::<bool>("enable-get-scans") {
-            config.endpoints.enable_get_scans = *enable;
```
nichtsfrei, line 601: Since we load from the file first, wouldn't it be possible to check if `config.endpoints.enable_get_scans` is false and only when it is false call `cmds.get_one::<bool>("enable-get-scans")`? That way we could still stick to `--enable-get-scans` without having to add an additional value to that.

jjnicola: You don't need to add an additional value. Just running `openvasd --enable-get-scans` will set this automatically to `true`, because of the `.default_missing_value("true")`

rust/openvasd/src/config.rs
```diff
@@ -422,8 +422,9 @@ impl Config {
                 clap::Arg::new("enable-get-scans")
                     .env("ENABLE_GET_SCANS")
                     .long("enable-get-scans")
-                    .action(ArgAction::SetTrue)
-                    .help("enable get scans endpoint"),
+                    .num_args(0..=1)
+                    .default_missing_value("true")
```
nichtsfrei, line 426: This is just a nitpick and you can ignore it: I think if you add ` .value_parser(clap::builder::BoolValueParser::new())` you can set `default_missing_value(true)` and spare the `matches!(...)` check later on.

jjnicola: Hey. That did the trick. Thanks!

nichtsfrei (approved):

PR #1723, opened 2024-09-18
Update doc/manual/nasl/nasl-grammar/index.md

scottvr:
**What**:

Replace the unicode n-dash (U+2013) with two ascii minus (0x2D) characters. 

**Why**:

Because a single Unicode N-dash will not work as the C decrement operator. 
(Probably someone's editor "helped" by converting the typed two dashes to the Unicode character.)

### DIFF
--- a/doc/manual/nasl/nasl-grammar/index.md
+++ b/doc/manual/nasl/nasl-grammar/index.md
@@ -99,7 +99,7 @@ NASL imported some operators from C:
 
 - `++` is the pre-incrementation (`++x`) or post-incrementation (`x++`).
 - `++x` adds 1 to x and returns the result; `x++` adds 1 to x but returns the previous value.
-- `-` is the pre-decrementation (`–x`) or post-decrementation (`x–`).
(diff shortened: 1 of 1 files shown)


scottvr:
FWIW, I don't think these build errors have anything to do with my simple documentation change. 

the log shows a 403 is returned by some server it tries to hit in the build process, and then there appears to be a problem with a missing libssh dependency:

```usr/bin/ld: warning: libssh.so.4, needed by /usr/local/lib/libgvm_util.so, not found (try using -rpath or -rpath-link)
/usr/bin/ld: /usr/local/lib/libgvm_util.so: undefined reference to `ssh_pki_import_privkey_base64@LIBSSH_4_5_0'
/usr/bin/ld: /usr/local/lib/libgvm_util.so: undefined reference to  ... ```

Tehforsch (approved):
Good catch. Thank you!

> FWIW, I don't think these build errors have anything to do with my simple documentation change.

You're right, this is just an unrelated temporary issue with the CI.

PR #1727, opened 2024-09-20
Refactor Sockets

Kraemii:
This PR contains a refactoring of the sockets module, as well, as the implementation of `recv_line` and `ftp_log_in`.

The refactoring divides the monolith socket implementation into the `tcp`, `tls` and `udp` modules.

This PR is part of SC-1120

For testing the FTP function I uploaded an example script, which can be executed by:
```
target/release/scannerctl execute script -t 44.241.66.173 examples/socket/ftp_login.nasl
```
I used this test server: https://dlptest.com/ftp-test/ which for the time being resolves into 44.241.66.173


rust/src/nasl/builtin/network/socket.rs
```diff
@@ -379,62 +205,80 @@ impl NaslSockets {
...
-
-                    result = conn.socket.recv_from(data.as_mut_slice());
-                }
-                if let Some(timeout) = old {
-                    conn.socket.set_read_timeout(Some(timeout))?;
-                }
-                ret
+                    }
+                    None => conn.read_line(&mut data)?,
+                };
```
Tehforsch, line 275: Same structure as above, although it's `read_line` here, so I think this is fine but I would replace it with something like
```rust
let pos = match convert_timeout(timeout) { 
    Some(timeout) => conn.read_with_timeout(&mut data[pos..], timeout), 
    None => conn.read(&mut data[pos..]), 
}?;
```

rust/src/nasl/builtin/network/socket.rs
```diff
@@ -769,11 +561,160 @@ impl NaslSockets {
...
+                        "FTP server did not respond with 220".to_string(),
+                        None,
+                    ));
+                }
+
+                for i in 0..1024 {
+                    let n = conn.read_line(&mut buf)?;
+                    if n == 0 {
+                        return Err(FunctionErrorKind::Diagnostic(
+                            "FTP server did not respond with 220".to_string(),
```
Tehforsch, line 631: Is this a correct error message here? It seems that this was appropriate above but shouldn't this be "FTP server did not respond"? Disclaimer: I don't understand the surrounding code at all since I don't know anything about FTP.

rust/src/nasl/builtin/network/socket.rs
```diff
@@ -769,11 +561,160 @@ impl NaslSockets {
...
+                    ));
+                }
+
+                // Check if buf contains 220
+                if !buf.contains("220") {
+                    return Err(FunctionErrorKind::Diagnostic(
+                        "FTP server did not respond with 220".to_string(),
+                        None,
+                    ));
+                }
```
Tehforsch, line 625: 1. Shouldn't we check whether the newly read buffer parses as 220 instead of just checking if it contains 220 somewhere?
Overall I would prefer if this check was also extracted into another function, maybe like:

```rust
fn get_response(mut conn: impl BufRead, buf: &mut String) -> Result<usize, FunctionErrorKind> {
    let old_len = buf.len();
    let n = conn.read_line(buf)?;
    if n == 0 {
        return Err(FunctionErrorKind::Diagnostic(
            "FTP server did not respond".to_string(),
            None,
        ));
    }
    buf[old_len..].parse::<usize>().map_err(|_| {
        FunctionErrorKind::Diagnostic("FTP server returned invalid response".to_string(), None)
    })
}
```

This would be useful not only for this check but also for all of the ones below.

rust/src/nasl/builtin/network/socket.rs
```diff
@@ -769,11 +561,160 @@ impl NaslSockets {
...
+                    let n = conn.read_line(&mut buf)?;
+                    if n == 0 {
+                        return Err(FunctionErrorKind::Diagnostic(
+                            "FTP server did not respond with 220".to_string(),
+                            None,
+                        ));
+                    }
+                    if buf.chars().nth(3) != Some('-') {
+                        break;
+                    }
```
Tehforsch, line 637: Doesn't this check the same position in the buffer every time?

rust/src/nasl/builtin/network/socket.rs
```diff
@@ -769,11 +561,160 @@ impl NaslSockets {
...
+                for _ in 0..1024 {
+                    let n = conn.read_line(&mut buf)?;
+                    if buf.chars().nth(3) != Some('-') || n == 0 {
+                        break;
+                    }
+                }
+                Err(FunctionErrorKind::Diagnostic(
+                    "FTP Server did not respond with ".to_string(),
+                    None,
+                ))
```
Tehforsch, line 707: missing a number

Tehforsch:
Looks much nicer than before! I have one more general comment:

Since #1731, you can remove the `RwLock` around the `Handles` struct, since the state is already wrapped in one. This will make creating the function set a little more annoying, but it will avoid having nested `RwLock`s. If you want to, check out [`[rust/src/nasl/builtin/ssh/impls.rs](https://github.com/greenbone/openvas-scanner/pull/1688/files#diff-491db9be623d2ff5324b0ed634cf3cdf07c695452d30374fc3b9ac1167793f5b)`](https://github.com/greenbone/openvas-scanner/blob/f4d0f993d07f64a844b0741271e43901a456e91f/rust/src/nasl/builtin/ssh/impls.rs#L30C20-L63)
 to see how to create the function set if it contains both `async fn ...(&mut self)` and `async fn ...(&self)`.

rust/src/nasl/builtin/network/network_utils.rs
```diff
@@ -22,15 +24,22 @@ pub fn ipstr2ipaddr(ip_addr: &str) -> Result<IpAddr, FunctionErrorKind> {
...
 
+/// Convert timeout
+pub fn convert_timeout(timeout: Option<i64>) -> Option<Duration> {
+    timeout.and_then(|timeout| {
+        if timeout < 1 {
+            None
+        } else {
+            Some(Duration::from_secs(timeout as u64))
+        }
+    })
```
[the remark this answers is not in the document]
Kraemii: Thanks, didn't know about filter! :)

rust/src/nasl/builtin/network/socket.rs
```diff
@@ -273,14 +122,14 @@ impl NaslSockets {
...
     ///
     /// On success the number of sent bytes is returned.
-    #[nasl_function(named(socket, data, flags, len))]
+    #[nasl_function(named(socket, data, option, len))]
     fn send(
         &self,
         socket: usize,
         data: &[u8],
-        flags: Option<i64>,
+        option: Option<i64>,
```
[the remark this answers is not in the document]
Kraemii: Yes, in the original implementation this is named option. I guess, I started implementing this function before we actually had the nasl_function macro, in which I could freely choose the variable name for the argument and forgot to rename it, when the macro was introduced.

rust/src/nasl/builtin/network/socket.rs
Kraemii: I rewrite the complete ft stuff using https://github.com/mattnenterprise/rust-ftp as template:
connect: https://github.com/mattnenterprise/rust-ftp/blob/master/src/ftp.rs#L53
login: https://github.com/mattnenterprise/rust-ftp/blob/master/src/ftp.rs#L349
response: https://github.com/mattnenterprise/rust-ftp/blob/master/src/ftp.rs#L694

Personally I'd liked to use a dedicated FTP client for FTP communication instead of just providing the login for NASL, but for now that is out of the question.

rust/src/nasl/builtin/network/mod.rs
```diff
@@ -75,6 +78,24 @@ impl Display for OpenvasEncaps {
...
+        match val {
+            NaslValue::String(val) => val.parse::<u8>().unwrap_or(2),
+            NaslValue::Number(val) => {
+                if !(1..=255).contains(&val) {
+                    2
+                } else {
+                    val as u8
+                }
+            }
+            _ => 2,
```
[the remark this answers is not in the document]
Tehforsch: Oh, right.

rust/src/nasl/builtin/network/socket.rs
```diff
@@ -534,81 +358,125 @@ impl NaslSockets {
...
+            if bufsz < 0 {
+                None
             } else {
-                Duration::from_secs(sec as u64)
+                Some(bufsz as usize)
             }
         } else {
-            Duration::from_secs(10)
+            None
         };
```
Tehforsch, line 375:
```rust
let bufsz = bufsz.filter(|bufsz| bufsz >= 0).map(|bufsz| bufsz as usize);
```

rust/src/nasl/builtin/network/socket.rs
```diff
@@ -534,81 +358,125 @@ impl NaslSockets {
...
+                                format!("unsupported transport layer: {transport} {tls_version}"),
+                                None,
+                            ))
+                        }
+                    },
+                }
+            };
+            let fd = self.add(NaslSocket::Tcp(Box::new(tcp)));
+            fds.push(fd);
         }
```
Tehforsch, line 480: I find this block hard to read.

I suggest extracting into a sub-function and doing something like this to make the logic a little clearer:

```suggestion rust
        let sockets: Vec<Option<NaslSocket>> = vhosts
            .iter()
            .map(|vhost| {
                self.open_sock_tcp_vhost(context, addr, timeout, bufsz, port, vhost, transport)
            })
            .collect::<Result<_, _>>()?;

        Ok(NaslValue::Fork(
            sockets
                .into_iter()
                .filter_map(|socket| socket)
                .map(|socket| {
                    let fd = self.add(socket);
                    NaslValue::Number(fd as i64)
                })
                .collect(),
        ))
    }

    fn make_tls_client_connection(
        &self,
        context: &Context,
        vhost: &str,
    ) -> Option<ClientConnection> {
        Self::get_tls_conf(context).ok().and_then(|conf| {
            create_tls_client(
                vhost,
                &conf.cert_path,
                &conf.key_path,
                &conf.password,
                &conf.cafile_path,
            )
            .ok()
        })
    }

    fn open_sock_tcp_vhost(
        &self,
        context: &Context,
        addr: IpAddr,
        timeout: Duration,
        bufsz: Option<usize>,
        port: u16,
        vhost: &str,
        transport: i64,
    ) -> Result<Option<NaslSocket>, FunctionErrorKind> {
        if transport < 0 {
            // TODO: Get port transport and open connection depending on it
            todo!()
        }
        let tls = match OpenvasEncaps::from_i64(transport) {
            // Auto Detection
            Some(OpenvasEncaps::Auto) => {
                // Try SSL/TLS first
                self.make_tls_client_connection(context, vhost)
            }
            // IP
            Some(OpenvasEncaps::Ip) => None,
            // Unsupported transport layer
            None | Some(OpenvasEncaps::Max) => {
                return Err(FunctionErrorKind::Diagnostic(
                    format!("unsupported transport layer: {transport} (unknown)"),
                    None,
                ))
            }
            // TLS/SSL
            Some(tls_version) => match tls_version {
                OpenvasEncaps::Tls12 | OpenvasEncaps::Tls13 => {
                    self.make_tls_client_connection(context, vhost)
                }
                _ => {
                    return Err(FunctionErrorKind::Diagnostic(
                        format!("unsupported transport layer: {transport} {tls_version}"),
                        None,
                    ))
                }
            },
        };
        Ok(
            TcpConnection::connect(addr, port, tls, timeout, bufsz, get_retry(context))
                .map(|tcp| NaslSocket::Tcp(Box::new(tcp)))
                .ok(),
        )
    }
```

Tehforsch (commented):
One general comment that I just noticed now:

Is the `NaslSocket::Closed` variant really necessary? If we kept track of the current fd in the `NaslSockets` struct and never re-used pre-existing ids we could just delete closed sockets immediately and wouldn't have to check for the closed socket case every time. Of course, the error message would change (from "socket already closed" to "socket not found"), but this could be fixed too by keeping track of the closed sockets

Tehforsch (approved):

PR #1728, opened 2024-09-23
Fix: improves stability on get_available_memory

RuffaloLavoisier:
**What**:
1. Added safety to check if it is null before using the file pointer.
2. Check buf null pointer, change fread() condition check.

**Why**:
1. File pointer may be null due to environmental factors.
2. If the fread fails, the buffer may be NULL.

**How**:
1. Add check condition for file pointer.
2. Add check condition for buf

**Checklist**:

- [ ] Tests

### DIFF
--- a/src/pluginlaunch.c
+++ b/src/pluginlaunch.c
@@ -398,8 +398,9 @@ plugin_timeout (nvti_t *nvti)
 static int
 get_available_memory ()
 {
-  char buf[8192], *hit;
-  FILE *fd;
+  char buf[8192] = NULL;
+  char *hit = NULL;
+  FILE *fd = NULL;
   size_t len;
 
   fd = fopen ("/proc/meminfo", "r");
--- a/src/pluginlaunch.c
+++ b/src/pluginlaunch.c
@@ -403,6 +403,11 @@ get_available_memory ()
   size_t len;
 
   fd = fopen ("/proc/meminfo", "r");
+  if (fd == NULL)
+    {
+      g_warning ("Couldn't open /proc/meminfo");
+      return 0;
+    }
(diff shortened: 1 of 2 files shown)


src/pluginlaunch.c
```diff
@@ -398,14 +398,20 @@ plugin_timeout (nvti_t *nvti)
...
   fd = fopen ("/proc/meminfo", "r");
+  if (fd == NULL)
+    {
+      g_warning ("Couldn't open /proc/meminfo");
+      return 0;
+    }
   len = fread (buf, 1, sizeof (buf) - 1, fd);
   fclose (fd);
-  if (len == 0)
+  if (len < (sizeof (buf) - 1) || buf == NULL)
```
jjnicola, line 414: If the content of the file has fewer chars than the buffer size, should not be an error.
Actually, I have just checked with `wc -c /proc/meminfo` and I got 1503 chars.

jjnicola (changes requested):

jjnicola:
Hi @RuffaloLavoisier ! thanks for your collaboration. Please, take a look at https://github.com/greenbone/openvas-scanner/tree/main/RELICENSE

src/pluginlaunch.c
RuffaloLavoisier: Hi, @jjnicola 
Thank you for review.

Oh I see, i ii revert it.

haha thanks 👍 
> Actually, I have just checked with wc -c /proc/meminfo and I got 1503 chars.

jjnicola (approved):

RuffaloLavoisier:
@jjnicola 

Hi there
When will it be merged into the main branch?

nichtsfrei (approved):

PR #1729, opened 2024-09-23
Fix doc links

ArnoStiefvater:
**What**:

SC-1178

**Why**:

**How**:

**Checklist**:

- [ ] Tests


jjnicola (commented):
I think, we should keep the relative path in case other environments have no `/rust/ ...` root dir

Tehforsch (approved):
Thanks! I checked all links and they all work for me.

PR #1732, opened 2024-09-27
Fix: unit-tests in openvas-scanner

robert-schardt:
**What**:

Fix: unit-tests in openvas-scanner
Ticket: https://jira.greenbone.net/browse/DEVOPS-1251
Fail: https://github.com/greenbone/openvas-scanner/actions/runs/11065187998/job/30744175259

**Why**:

We want to fix the Unit Tests in CI, so the product gets properly built.

**How**:

Look at CI Unit-Test and see if they fail.

**Checklist**:

- [x] Tests

### DIFF
--- a/rust/cross.Dockerfile
+++ b/rust/cross.Dockerfile
@@ -5,7 +5,7 @@ RUN apt-get update && apt-get install -y \
   flex \
   curl \
   zlib1g-dev
-RUN curl -o /tmp/pcap.tar.gz https://www.tcpdump.org/release/libpcap-1.10.3.tar.gz
+RUN curl -L -o /tmp/pcap.tar.gz https://www.tcpdump.org/release/libpcap-1.10.3.tar.gz
 WORKDIR /tmp
 RUN tar xvf pcap.tar.gz
 RUN ls -las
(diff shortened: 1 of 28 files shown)


nichtsfrei:
Please don't use `Change:` within the commit message as it indicates that it is a breaking change.

robert-schardt:
To fix openssl error in functional.yml: https://github.com/greenbone/openvas-scanner/actions/runs/11177195503/job/31072763918
```
Run scannerctl syntax --quiet "$(openvas -s | grep plugins_folder | sed 's/plugins_folder = //')/"
scannerctl: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory
```
Installed pkg-config with libssl-dev as mentioned in crate documentation for openssl-sys: https://docs.rs/openssl/0.10.66/openssl/

nichtsfrei:
Regarding the testing and oldstable bulds we need a versions of gvm-libs for that. Those seem to be currently missing.

rust/cross.Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN ./configure
 RUN make install
 RUN ldconfig
 
-RUN curl -o /tmp/openssl.tar.gz https://www.openssl.org/source/old/1.1.1/openssl-1.1.1.tar.gz
+RUN curl -L -o /tmp/openssl.tar.gz https://www.openssl.org/source/old/1.1.1/openssl-1.1.1.tar.gz
```
nichtsfrei, line 27: wouldn't curl -Lfo be better to indicate early that it failed to load instead of later when tar tries to extract a html page?

y0urself:
@greenbone/scanner-maintainers you may want to review.

nichtsfrei (approved):

PR #1733, opened 2024-10-03
Fix: setting stack buffer to NULL

nichtsfrei:
Since buffer within pluginlaunch.c is within the stack it cannot be NULL. This fixes that oversight.

I'm unsure why we even double check for the buffer to be empty when it is already indicated by len of fread.

### DIFF
--- a/src/pluginlaunch.c
+++ b/src/pluginlaunch.c
@@ -398,10 +398,11 @@ plugin_timeout (nvti_t *nvti)
 static int
 get_available_memory ()
 {
-  char buf[8192] = NULL;
+  char buf[8192];
   char *hit = NULL;
   FILE *fd = NULL;
   size_t len;
+  memset (buf, 0, sizeof (buf));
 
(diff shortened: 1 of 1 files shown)


jjnicola (approved):

src/pluginlaunch.c
```diff
@@ -411,7 +412,11 @@ get_available_memory ()
     }
   len = fread (buf, 1, sizeof (buf) - 1, fd);
```
RuffaloLavoisier, line 413: Hi there, @nichtsfrei 
IMHO - Rather than just reading the file, I think it would be better to get the file size from the offset and then make sure that the fread() reads and check as normally as the file size 😆 

How about you?

RuffaloLavoisier: It may not be too important, but `char buf` is using around 8192 of arbitrary unnecessary memory.
so I think it's good to get the file size and then read it that much.

jjnicola:
Fixes #1728

PR #1741, opened 2024-10-24
Add experimental build to CI

Tehforsch:
**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,6 +35,8 @@ jobs:
         working-directory: rust
       - run: CROSS_CONFIG=Cross.toml cross build --release --target x86_64-unknown-linux-gnu
         working-directory: rust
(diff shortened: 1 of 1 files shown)


nichtsfrei:
Should that be in the release notes?

.github/workflows/build.yml
```diff
@@ -35,6 +35,8 @@ jobs:
         working-directory: rust
       - run: CROSS_CONFIG=Cross.toml cross build --release --target x86_64-unknown-linux-gnu
         working-directory: rust
+      - run: CROSS_CONFIG=Cross.toml cross build --release --target x86_64-unknown-linux-gnu --features experimental
```
nichtsfrei, line 38: if we set it here we will ship the experimental version when we release.

Tehforsch: Ah, thanks. I mostly wanted to see what github would do, so the PR is not yet meant to be taken seriously (yet?)

PR #1752, opened 2024-11-04
Error handling for builtin functions 

Tehforsch:
This is a proposal for revamped error type structure and handling for NASL builtin functions as well as some minor changes to error handling in general.

# Error type structure
The main type this refactoring is concerned with is `FunctionErrorKind` which has been renamed to `FnError`.

In this proposed new error structure, `FnError` is an actual type that contains metadata about the error (such as whether the error can be solved by retrying the operation that caused it and the return value given to the NASL caller), as well as its kind, described by the `FnErrorKind` enum. This enum corresponds to the previous `FunctionErrorKind` and is structured as follows:

```rust
pub enum FnErrorKind {
    Argument(ArgumentError),
    Internal(InternalError),
    Builtin(BuiltinError),
}
```

The variants represent three different kinds of errors: `ArgumentError` refers to errors caused by calling a function with the wrong arguments which typically reflect an error in the usage of the function by the script authors.

`InternalError` refers to any error caused by failure of an internal process of the scanner (such as the storage). This reflects a deeper problem and may mean we cannot simply proceed with execution of the scan.

Finally, `BuiltinError` represents any (more or less expected) error that occurred within a builtin function. This may include timeouts, authentication failures, etc..

# Error handling
I think the exact rules for how we treat each error type are something that is still open to discussion, but I'll write down the rules for the implementation in this PR as a basis for this discussion.

As stated above, the metadata (return value, retryable) on how to handle a specific error are stored in a `FnError` (as opposed to deriving them by matching on the kind of the error). This means that in principle, we always have the option of explicitly overriding any defaults for any specific case. However, I hope that we rarely need this in practice because we can choose sane defaults that reflect what we want most of the time.

## Return behavior
The return behavior of a given error specifies how an error should be handled during execution. This is specified by the following type:
```rust
enum ReturnBehavior {
    ExitScript,
    ReturnValue(NaslValue),
}
```
This type is equivalent to `Option<NaslValue>` but we decided that having a custom type here would make the intent clearer.
When the interpreter encounters an error with `ReturnBehavior::ExitScript` behavior, it will unsurprisingly exit the script. If it encounters an error with `ReturnBehavior::ReturnValue(val)`, it will return `val` and continue execution. This behavior is effectively equivalent to the `Diagnostic` variant of `FunctionErrorKind`.

The `Argument` and `Internal` variants of `FnError` are automatically constructed with `ReturnBehavior::ExitScript`, meaning that they abort execution of the script. The `Builtin` variant is constructed with `ReturnBehavior::ReturnValue(NaslValue::Null)` by default, but this value can easily be overwritten when the error is created, for example:
```rust
HttpError::HandleIdNotFound(handle).with(ReturnValue(-1))
```

## Retry behavior
Certain errors can be flagged as being solvable by retrying the operation that caused them. This is represented by a `retryable` boolean on `FnError`, which is `false` by default for all variants except for a specific internal error in the storage. However, this default behavior can be overwritten at error creation if needed, for example
```rust
HttpError::ConnectionFailed(...).with(Retryable)
```
I also added a small test to make sure that the interpreter does actually retry retryable errors.

# Boilerplate / Cost
Since this was the main point of the discussion on implementation strategies for these errors, I figured I'd write down the amount of additional overhead for creating these custom error types that is currently needed when a new builtin module is created.

1. Add a custom error type for the builtin module. These can be of arbitrary form but a typical error type might look like
```rust
#[derive(Debug, Error)]
enum FooError {
    #[error("Bar occurred.")]
    Bar,
    #[error("Baz occurred. Here is some more data: {0}")]
    Baz(SomeData),
}
```
I personally like this approach anyways since it keeps error messages all in one place and makes it easy to get an overview of all the error messages and to keep them consistent.

2. Add this builtin error as a variant of the `BuiltinError` type described above. This is one line of code:
```rust
enum BuiltinError {
    ...
    Foo(FooError),
    ...
}
```

3. For convenience, some `From` impls and `TryFrom` impls can make the error type easier to use by enabling use of the question mark operator. I added a tiny macro that implements these traits (because the implementations are usually trivial), so this comes down to one line too:
`builtin_error_variant!(Foo, FooError);`

This is all that is needed to make this error usable in NASL functions:
```rust
fn check(a: usize) -> Result<(), FooError> {
    if a == 0 {
        Err(FooError::Bar)
    }
    else { 
        Ok(())
    }
}

#[nasl_function]
fn foo(a: usize) -> Result<usize, FnError> {
    check(a)?;
    Ok(a)
}
```

As a side note, NASL functions can also return any concrete `impl Into<FnError>` directly, so for this case we can also write

```rust
#[nasl_function]
fn foo(a: usize) -> Result<usize, FooError> {
    if a == 0 { 
        Err(FooError::Bar)
    }
    else {
        Ok(a)
    }
}
```

Note that the above `From` impls that are automatically written by the `builtin_error_variant!` macro can also be manually implemented if one wants to specify defaults for a specific error variant. For example

```rust
impl From<FooError> for FnError {
    fn from(e: FooError) -> FnError {
        match e {
            FooError::Foo => BuiltinError::Foo(e).with(ReturnValue(-1)),
            _ => BuiltinError::Foo(e).into(),
        }
    }
}
```

# `PartialEq`, `Eq` and `Clone`
Previously, all our error types derived `PartialEq`, `Eq`, `Clone`. This was basically unused in the codebase (with some minor exceptions where an easy rewrite made these derives unnecessary). In this PR I got rid of all of these derives. This has the advantage of making it easier to wrap certain internal library errors that may not implement any of the three traits (`std::io::Error` implements neither and is used quite a bit, for example, causing us to always use the `io::ErrorKind` instead.). I think there is a case to be made for either choice, but personally prefer not implementing these.
Note that this will also be a problem in a `dyn FnError`-based solution, since the type will have to become `dyn FnError + PartialEq + Eq + Clone` or alternatively `FnError` will have to be defined as `trait FnError: PartialEq + Eq + Clone`, either requiring each concrete type to implement the three traits.

# `WithErrorInfo`
As a last, opinionated part of this PR, I added a trait `WithErrorInfo` that is basically meant to attach some kind of information to an already existing error. This originated in the SSH implementation, where I wanted to add the `SessionId` that caused the error to ~70% of error variants (but not all of them, which would have meant I could simply make it part of the parent error). I also often wanted to add the internal library error that caused the error to occur. This trait made it convenient to simply write `SshError::Foo.with(session_id).with(err)`. I then stumbled upon this concept again when I wanted to attach the retryable/return behavior information discussed above to certain errors, so I simply extended this trait to allow usage like `.with(ReturnValue(-1))` or `.with(Retryable)` where `ReturnValue` and `Retryable` are simply dummy types meant for just this one purpose. This has the nice advantage of allowing autoconversion to `FnError` making the resulting code look really clean, but if this is too fancy I'm fine with getting rid of it. An alternative would be to implement methods like `with_return_value` and `.with_retryable` for every type that needs them or to instead write `FnError::from(...).with_retryable()` which is more verbose.

# Todo
- [ ] Write documentation on how to do error handling now, i.e. the steps to take when writing new builtin functions, general guidelines, as well as the meaning of the individual error variants. Ideally in a `.md` document somewhere in the crate.
- [ ] Open tickets for parts of the code where error handling doesn't follow these guidelines
- [ ] Get rid of two unnecessary constructors `FnError` / `ArgumentError`


rust/src/nasl/builtin/cryptographic/aes_gcm.rs
```diff
@@ -2,12 +2,8 @@
...
 // SPDX-License-Identifier: GPL-2.0-or-later WITH x11vnc-openssl-exception
 
-// FunctionErrorKind::GeneralError
-use crate::nasl::syntax::NaslValue;
-use crate::{
-    function_set,
-    nasl::utils::{Context, FunctionErrorKind, Register},
-};
+// NaslError::GeneralError
+use crate::nasl::prelude::*;
```
nichtsfrei, line 6: personal opinion: while combating boilerplate code this does make lookups more annoying.

rust/src/nasl/interpreter/error.rs
```diff
@@ -18,12 +19,12 @@ pub struct FunctionError {
     pub function: String,
     /// Kind of error
     #[source]
-    pub kind: FunctionErrorKind,
+    pub kind: NaslError,
```
nichtsfrei, line 22: Nit: I am not sure if I like the term  NaslError as it is just for builtin functions.

rust/src/nasl/utils/error.rs
```diff
@@ -3,136 +3,159 @@
...
-    #[error("Authentication failed.")]
-    Authentication,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Error)]
+pub enum InternalError {
+    #[error("{0}")]
+    Storage(#[from] StorageError),
+    #[error("{0}")]
+    Dirty(String),
```
nichtsfrei, line 37: can we get rid of Dirty? Dirty in StorageError is used because it is short version of dirty data which includes but is not limited to corruption or undefined behaviour. I don't think it makes sense as an InternalError. I think we should be more considered and may define extra cases that are not Storage specific.

nichtsfrei:
Keep in mind that the io::ErrorKind is used to identify retry cases, see https://github.com/greenbone/openvas-scanner/blob/main/rust/src/nasl/interpreter/interpreter.rs#L196 for an incomplete example. The possibility to identify cases that can be retried transparently to the user should be considered.

rust/src/nasl/utils/error.rs
Tehforsch: Yes, I just wasnt finished and dirty is used a ton. Got rid of it in some of the newer commits.

rust/src/nasl/interpreter/error.rs
Tehforsch: That makes sense. I guess I'll rename it back to `FunctionErrorKind` for this draft (to make the changes I made more apparent and not flood it with 40 files of rename changes) and then we can discuss the name at some later point.

Tehforsch:
> Keep in mind that the io::ErrorKind is used to identify retry cases, see https://github.com/greenbone/openvas-scanner/blob/main/rust/src/nasl/interpreter/interpreter.rs#L196 for an incomplete example. The possibility to identify cases that can be retried transparently to the user should be considered.

I think we need a more robust set up to identify these cases. Using IO errors to mean retryable might be a good heuristic, but is still not accurate, since there are both false positives (io::ErrorKind::NotADirectory is not something that will be solved by retrying) and false negatives (an ssh connection might fail due to a timeout but won't return an IO error).
I am not sure that we can find a high level structure that represents re-try-ability well and would prefer doing this on a "local" level, i.e. someting like
```foo().map_err(|e| MyCustomError::MyCustomErrorVariant(e).retryable())````

(method naming very much up for debate)

**Comment by nichtsfrei** [comment]
Yes, it’s imperfect—I wasn’t suggesting we use it as-is. However, this concept is entirely missing from the current proposal, even though it’s available, albeit imperfectly, in the current handling.

I’m not in favor of implementing it at a local level because it relies too heavily on the function developer's knowledge and could lead to inconsistent handling, making it harder to reason about, inconvenient, and adding cognitive load—especially when you just want to quickly create a built-in function.

**Comment by Tehforsch** [comment]
> Yes, it’s imperfect—I wasn’t suggesting we use it as-is. However, this concept is entirely missing from the current proposal, even though it’s available, albeit imperfectly, in the current handling.

Yes, youre right.

> I’m not in favor of implementing it at a local level because it relies too heavily on the function developer's knowledge and could lead to inconsistent handling, making it harder to reason about, inconvenient, and adding cognitive load—especially when you just want to quickly create a built-in function.

I was also afraid of that, but if we don't trust the author of the builtin function to make the correct decision on whether an error is solvable by retrying or not, then I don't think we have a chance of solving this problem anyways. I would prefer doing this on a local level because it still gives us the room to discuss these things in code review etc.. I think making the overarching structure of the error type determine whether an error is "retryable" or not carries a larger risk because we will eventually have to fight our own system if we want more precise control over these errors. 

This also doesn't always have to be solved on the most local level, because you could, for example, write an `From` impl for your error type such as

```rust
impl From<io::Error> for MyCustomError {
    fn from(val: io:Error) -> MyCustomError {
        MyCustomError::Foo(val).retryable()
    }
}
```

which effectively reproduces the original behavior of making any `io::Error` something that can be solved by retrying, but does it on a more fine-grained level (for an entire builtin module), where it might always make sense.

nichtsfrei (approved):

PR #1757, opened 2024-11-07
NASL builtin function for open privileged sockets with udp and tcp

Kraemii:
**What**:
Adds the NASL functions for open privileged sockets for TCP and UDP:
- open_priv_sock_tcp
- open_priv_sock_udp

I added two NASL script for testing TCP and UDP. Note that you need privileged 
This PR depends on https://github.com/greenbone/openvas-scanner/pull/1727
Jira: SC-1120

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/src/nasl/builtin/network/network_utils.rs
+++ b/rust/src/nasl/builtin/network/network_utils.rs
@@ -40,16 +40,12 @@ pub fn bind_local_socket(dst: &SocketAddr) -> io::Result<UdpSocket> {
 }
 
 /// Return the source IP address given the destination IP address
-pub fn get_source_ip(dst: IpAddr, port: u16) -> Result<IpAddr, FunctionErrorKind> {
+pub fn get_source_ip(dst: IpAddr, port: u16) -> io::Result<IpAddr> {
     let socket = SocketAddr::new(dst, port);
     let sd = format!("{}:{}", dst, port);
     let local_socket = bind_local_socket(&socket)?;
-    local_socket
-        .connect(sd)
-        .ok()
-        .and_then(|_| local_socket.local_addr().ok())
-        .and_then(|l_addr| IpAddr::from_str(&l_addr.ip().to_string()).ok())
-        .ok_or_else(|| FunctionErrorKind::Diagnostic("No route to destination".to_string(), None))
+    local_socket.connect(sd)?;
+    Ok(local_socket.local_addr()?.ip())
 }
 
 /// Tests whether a packet sent to IP is LIKELY to route through the
--- a/rust/src/nasl/builtin/network/socket.rs
+++ b/rust/src/nasl/builtin/network/socket.rs
@@ -26,12 +26,15 @@ use super::{
     verify_port, OpenvasEncaps,
 };
 
-pub struct Interval {
+/// Interval used for timing tcp requests. Any tcp request has to wait at least
+/// the time defined by this interval after the last request has been sent.
+struct Interval {
     interval: Duration,
     last_tick: SystemTime,
 }
 
 impl Interval {
+    /// Check the time since the last tick and wait if necessary.
     pub fn tick(&mut self) {
         if let Ok(since) = SystemTime::now().duration_since(self.last_tick) {
             if since < self.interval {
@@ -42,6 +45,7 @@ impl Interval {
     }
 }
 
+/// A small configuration Struct to store TLS relevant information
 struct TlsConfig {
     cert_path: String,
     key_path: String,
@@ -49,6 +53,8 @@ struct TlsConfig {
     cafile_path: String,
 }
 
+/// Representation of a NASL socket. A NASL socket can be either TCP (including TLS),
+/// UDP or Closed
 enum NaslSocket {
     // The TCP Connection is boxed, because it uses a lot of space
     // This way the size of the enum is reduced
@@ -57,12 +63,16 @@ enum NaslSocket {
     Closed,
 }
 
+/// Struct storing handles for socket operations as well as a List of Closed sockets
+/// to override next.
 #[derive(Default)]
 struct Handles {
     handles: Vec<NaslSocket>,
     closed_fd: Vec<usize>,
 }
 
+/// The Top level struct storing all NASL sockets as well as the interval for tcp
+/// requests.
 #[derive(Default)]
 pub struct NaslSockets {
     handles: RwLock<Handles>,
@@ -70,6 +80,8 @@ pub struct NaslSockets {
 }
 
 impl NaslSockets {
+    /// Adds a given NASL socket. It returns the position of the socket within the
+    /// list.
     fn add(&self, socket: NaslSocket) -> usize {
         let mut handles = self
             .handles
@@ -109,6 +121,7 @@ impl NaslSockets {
         Ok(NaslValue::Null)
     }
 
+    /// Check if a new tcp request can be sent and waits if necessary.
     fn wait_before_next_probe(&self) {
         if let Some(interval) = &self.interval {
             interval.write().unwrap().tick();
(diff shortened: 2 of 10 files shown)


rust/examples/socket/ftp_login.nasl
```diff
@@ -0,0 +1,16 @@
...
+#
+# SPDX-License-Identifier: GPL-2.0-or-later WITH x11vnc-openssl-exception
+
+display("Start");
+display("is function open_sock_tcp defined: ", defined_func("open_sock_tcp"));
+sock = open_sock_tcp(21, transport: 1);
+display("was socket created: ", !isnull(sock));
+display("fd: ", sock);
+display("is function ftp_log_in defined: ", defined_func("ftp_log_in"));
+# Login data for ftp://ftp.dlptest.com/ provided by https://dlptest.com/ftp-test/
```
nichtsfrei, line 11: q: are we allowed to just use that in our public repository?

rust/src/nasl/builtin/network/network_utils.rs
```diff
@@ -22,29 +24,28 @@ pub fn ipstr2ipaddr(ip_addr: &str) -> Result<IpAddr, FunctionErrorKind> {
...
     let socket = SocketAddr::new(dst, port);
     let sd = format!("{}:{}", dst, port);
     let local_socket = bind_local_socket(&socket)?;
-    local_socket
-        .connect(sd)
-        .ok()
-        .and_then(|_| local_socket.local_addr().ok())
-        .and_then(|l_addr| IpAddr::from_str(&l_addr.ip().to_string()).ok())
-        .ok_or_else(|| FunctionErrorKind::Diagnostic("No route to destination".to_string(), None))
+    local_socket.connect(sd)?;
```
nichtsfrei, line 47: q: don't we need a disconnect/close here?

rust/src/nasl/builtin/network/socket.rs
```diff
@@ -534,81 +397,123 @@ impl NaslSockets {
...
-                            "unsupported transport layer: {transport}{tls_version}"
+                            "unsupported transport layer: {transport}(unknown)"
                         )))
                     }
-                },
-            }
+                    // TLS/SSL
+                    Some(tls_version) => match tls_version {
+                        OpenvasEncaps::Tls12 | OpenvasEncaps::Tls13 => {
+                            if let Ok(tls_conf) = Self::get_tls_conf(context) {
```
nichtsfrei, line 480: nit: can you separate those blocks and may give them a name? This is hard to read and the else branch is always continue.

rust/src/nasl/builtin/network/socket.rs
```diff
@@ -534,81 +397,123 @@ impl NaslSockets {
...
+                                } else {
+                                    continue;
+                                }
+                            } else {
+                                continue;
+                            }
+                        }
+                        _ => {
+                            return Err(FunctionErrorKind::WrongArgument(format!(
+                                "unsupported transport layer: {transport}{tls_version}"
```
nichtsfrei, line 509: q: although it is a wrong argument, that wrong argument could have been valid on a previously installed version (when e.g. a tls_version drops a transport layer, hence I think it should be a Diagnostic instead?

rust/src/nasl/builtin/network/udp.rs
```diff
@@ -0,0 +1,115 @@
...
+            };
+        }
+        unreachable!()
+    }
+}
+
+impl Write for UdpConnection {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        let mtu = mtu(self.socket.peer_addr()?.ip());
+        if buf.len() < mtu {
```
nichtsfrei, line 41: nit: I think guardian instruction are easier to read when the return early, so instead of writing it like this I would propose to switch the order:
```
if buf.len() > mtu {
  return Err(io::Error::new(...);
}
result = ...
Ok(result as usize)
```
That way a reader can immediately skip the guardian check without having to keep the branch in mind and concentrate on the result generation.

rust/src/nasl/builtin/network/udp.rs
```diff
@@ -0,0 +1,115 @@
...
+                    "UDP data of size {} exceeds the maximum length of {}",
+                    buf.len(),
+                    mtu
+                ),
+            ))
+        }
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        Ok(())
```
nichtsfrei, line 68: don't we need flushing?

rust/src/nasl/builtin/network/socket.rs
Tehforsch: I agree. Might be a case for `.and_then`

rust/src/nasl/builtin/network/socket.rs
```diff
@@ -618,161 +523,250 @@ impl NaslSockets {
...
+            )),
+        }?;
+
+        let cafile_path = match get_kb_item(context, "Secret/tls_cafile")? {
+            Some(x) => Ok(x.to_string()),
+            None => Err(FunctionErrorKind::Diagnostic(
+                "KB key 'Secret/tls_cafile' is not set".to_string(),
+                None,
+            )),
+        }?;
```
Tehforsch, line 559:
```suggestion rust
        let get_kb_string = |name| {
            match get_kb_item(context, name)? {
                    Some(x) => Ok(x.to_string()),
                    None => Err(FunctionErrorKind::Diagnostic(
                        format!("KB key {} is not set", name),
                        None,
                    )),
                }
        };

        let cert_path = get_kb_string("Secret/tls_cert")?;
        let key_path = get_kb_string("Secret/tls_key")?;
        let password = get_kb_string("Secret/tls_password")?;
        let cafile_path = get_kb_string("Secret/tls_cafile")?;
```

Tehforsch: or better yet make the closure a function and use it in `open_sock_kdc` as well

rust/src/nasl/builtin/network/socket.rs
```diff
@@ -618,161 +523,250 @@ impl NaslSockets {
...
+
+        let mut sport = 1023;
+
+        while sport > 0 {
+            if let Ok(udp) = UdpConnection::new_priv(addr, sport, dport) {
+                let fd = self.add(NaslSocket::Udp(udp));
+                return Ok(NaslValue::Number(fd as i64));
+            }
+            sport -= 1;
+        }
```
Tehforsch, line 661: I think
```rust
for sport in (1..=1023).rev() {
}
```

is a little cleaner, but its debatable

rust/src/nasl/builtin/network/socket.rs
```diff
@@ -48,177 +43,43 @@ impl Interval {
...
+/// UDP or Closed
 enum NaslSocket {
-    // The TCP Connection is boxed, because it uses allot of space
+    // The TCP Connection is boxed, because it uses a lot of space
     // This way the size of the enum is reduced
-    Tcp(Box<TCPConnection>),
-    Udp(UDPConnection),
+    Tcp(Box<TcpConnection>),
+    Udp(UdpConnection),
     Close,
```
Tehforsch, line 61: Nitpick: Rename this to `Closed`?

rust/src/nasl/builtin/network/network_utils.rs
Kraemii: In general you are correct, but in the rust world this is done automatically as soon, as the connection is out of scope. There is not even a function you could call to close the connection except for the `libc` function itself.

rust/src/nasl/builtin/network/socket.rs
nichtsfrei: or better yet move the closure as a function besides get_kb_item and call it  kb_item_err or similar. That way this case would be available for everybody.

rust/src/nasl/builtin/network/udp.rs
Kraemii: No, as every data is sent out immediately. The internal buffer is for resending that data in case it did not reach its destination on the first try.

rust/examples/socket/ftp_login.nasl
Kraemii: It is a public test server for ftp. As we do not spam this server with automated test I don't see a problem here.

rust/src/nasl/builtin/network/socket.rs
```diff
@@ -512,6 +548,91 @@ impl NaslSockets {
...
+                }
+            } else {
+                if let Ok(udp) = UdpConnection::new_priv(addr, sport, dport) {
+                    self.add(NaslSocket::Udp(udp))
+                } else {
+                    continue;
+                }
+            };
+            return Ok(NaslValue::Number(fd as i64));
+        }
```
Tehforsch, line 593: I would find this a lot easier to read with something like this (couldn't think of a name for the first method):

```rust
    fn open_priv_sock_internal(&self, addr: IpAddr, sport: u16, dport: u16, tcp: bool) -> Result<usize, FunctionErrorKind> {
        if tcp {
            // TODO: set timeout to global recv timeout when available
            let timeout = Duration::from_secs(10);
            self.wait_before_next_probe();
            let tcp = TcpConnection::connect_priv(addr, sport, dport, timeout)?;
            Ok(self.add(NaslSocket::Tcp(Box::new(tcp))))
        } else {
            let udp = UdpConnection::new_priv(addr, sport, dport)?;
            Ok(self.add(NaslSocket::Udp(udp)))
        }
    }

    fn open_priv_sock(
        &self,
        addr: IpAddr,
        dport: i64,
        sport: Option<i64>,
        tcp: bool,
    ) -> Result<usize, FunctionErrorKind> {
        let dport = verify_port(dport)?;

        if let Some(sport) = sport {
            let sport = verify_port(sport)?;
            return Ok(self.open_priv_sock_internal(addr, sport, dport, tcp)?);
        }

        for sport in (1..=1023).rev() {
            let res = self.open_priv_sock_internal(addr, sport, dport, tcp);
            if let Ok(fd) = res {
                return Ok(fd)
            }
        }
        Err(FunctionErrorKind::Diagnostic(
            format!(
                "Unable to open priv socket to {} on any socket from 1-1023",
                addr
            ),
            None,
        ))
    }
```

Tehforsch (approved):
Just one minor comment

Tehforsch (approved):

PR #1758, opened 2024-11-07
Add host nasl built-in functions

jjnicola:
**What**:
Add host nasl built-in functions
Jira: SC-1189

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/src/nasl/builtin/host/mod.rs
+++ b/rust/src/nasl/builtin/host/mod.rs
@@ -14,12 +14,25 @@ use dns_lookup::lookup_addr;
 
 use crate::function_set;
 use crate::nasl::utils::{error::FunctionErrorKind, hosts::resolve, lookup_keys::TARGET};
+use crate::nasl::utils::error::FunctionErrorKind;
 
 use crate::nasl::syntax::NaslValue;
 use crate::nasl::utils::{Context, ContextType, Register};
 
+/// Get a list of found hostnames or a IP of the current target in case no hostnames were found yet.
+fn get_host_names(_register: &Register, context: &Context) -> Result<NaslValue, FunctionErrorKind> {
+    if let Some(hns) = context.target_vhosts() {
+        let hns = hns
+            .into_iter()
+            .map(|(h, _s)| NaslValue::String(h))
+            .collect::<Vec<_>>();
+        return Ok(NaslValue::Array(hns));
+    };
+    Ok(NaslValue::String(context.target().to_string()))
+}
+
 /// Return the target's IP address as IpAddr.
-pub fn get_host_ip(context: &Context) -> Result<IpAddr, FunctionErrorKind> {
+fn get_host_ip(context: &Context) -> Result<IpAddr, FunctionErrorKind> {
     let default_ip = "127.0.0.1";
     let r_sock_addr = match context.target() {
         x if !x.is_empty() => IpAddr::from_str(x),
@@ -35,6 +48,9 @@ pub fn get_host_ip(context: &Context) -> Result<IpAddr, FunctionErrorKind> {
     }
 }
 
+///Expands the vHosts list with the given hostname.
+///The mandatory parameter hostname is of type string. It contains the hostname which should be added to the list of vHosts
+///Additionally a source, how the hostname was detected can be added with the named argument source as a string. If it is not given, the value NASL is set as default.
 pub fn add_host_name(
     register: &Register,
     context: &Context,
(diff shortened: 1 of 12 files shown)


rust/src/nasl/utils/hosts.rs
```diff
@@ -0,0 +1,20 @@
...
+
+use std::net::{SocketAddr, ToSocketAddrs};
+
+use crate::nasl::utils::error::FunctionErrorKind;
+
+pub fn resolve(
+    mut hostname: String,
+) -> Result<Option<Box<dyn Iterator<Item = SocketAddr>>>, FunctionErrorKind> {
+    //std::net to_socket_addrs() requires a port. Therefore, using a dummy port
+    hostname.push_str(":5000");
```
Tehforsch, line 13: Given that we add a dummy port, we shouldn't be returning `SocketAddr` from this function, but rather return an `IpAddr`, since that is the only valuable information a caller can extract.

rust/src/nasl/utils/context.rs
```diff
@@ -329,6 +333,59 @@ impl NaslContext {
...
+    //
+    // A Tokio mutex is mostly intended to be used when locks need to be held
+    // across `.await` yield points. All other cases are **usually** best
+    // served by a std mutex. If the critical section does not include any
+    // async operations but is long (CPU intensive or performing blocking
+    // operations), then the entire operation, including waiting for the mutex,
+    // is considered a "blocking" operation and `tokio::task::spawn_blocking`
+    // should be used.
+    /// vhost list which resolve to the IP address and their sources.
+    vhosts: Mutex<Vec<(String, String)>>,
```
Tehforsch, line 355: Comment: I know we talked about this and didnt find a short-term solution that was any better than this mutex, but I think it is a shame that we have to resort to mutexes all over the place instead of finding a more robust solution that passes the responsibility of handling shared/mutable state to the interpreter/executor.

rust/src/nasl/builtin/host/mod.rs
```diff
@@ -79,14 +133,186 @@ fn nasl_get_host_ip(
...
+        Some(ContextType::Value(NaslValue::String(x))) if !x.is_empty() => x.clone(),
+        _ => {
+            return Err(FunctionErrorKind::diagnostic_ret_null("Missing Hostname"));
+        }
+    };
+
+    match resolve(hostname)? {
+        Some(mut a) => {
+            let address = a.next().map_or_else(String::new, |x| x.to_string());
+            let address = &address[..(address.len() - 5)];
```
Tehforsch, line 151: What is the 5? I think what youre trying to do is strip the port from the ip addr? This should be changed by returning `IpAddr` from `resolve` (which is the only thing that makes sense anyways, see below) and not by doing custom parsing logic on the `Display` result of `SocketAddr`

Tehforsch (changes requested):

rust/src/nasl/builtin/host/mod.rs
```diff
@@ -134,169 +131,76 @@ fn nasl_get_host_ip(
...
-    let cmp_hostname = match register.named("cmp_hostname") {
-        Some(ContextType::Value(NaslValue::Boolean(x))) => *x,
-        _ => false,
-    };
+#[nasl_function(named(cmp_hostname))]
+fn same_host(
+    h1: &str,
+    h2: &str,
+    cmp_hostname: Option<bool>,
+) -> Result<NaslValue, FunctionErrorKind> {
```
Tehforsch, line 173: Nit: You could change the return type to `Result<bool, _>`

rust/src/nasl/builtin/host/mod.rs
```diff
@@ -134,169 +131,76 @@ fn nasl_get_host_ip(
...
+    let addr2: Vec<IpAddr> = h2
+        .into_iter()
+        .map(|x| x.parse().unwrap())
+        .collect::<Vec<_>>();
 
     let hostnames1 = addr1
         .clone()
         .into_iter()
-        .map(|x| lookup_addr(&x))
+        .filter_map(|x| lookup_addr(&x).ok())
```
Tehforsch, line 189: Haven't tested, but you should be able to get rid of the clone by using `.iter()` instead. Something like this:

```suggestion rust 
    let hostnames1 = addr1
        .iter()
        .filter_map(|x| lookup_addr(x).ok())
```

rust/src/nasl/utils/hosts.rs
```diff
@@ -2,18 +2,22 @@
...
+pub fn resolve(mut hostname: String) -> Result<Vec<String>, FunctionErrorKind> {
     //std::net to_socket_addrs() requires a port. Therefore, using a dummy port
     hostname.push_str(":5000");
 
     match hostname.to_socket_addrs() {
-        Ok(addr) => Ok(Some(Box::new(addr))),
+        Ok(addr) => {
+            let ips = addr
+                .into_iter()
+                .map(|x| x.ip().to_string())
```
Tehforsch, line 17: Why the `.to_string()`? It makes more sense to me to be explicit about what this function returns and to simply return the `IpAddr` instead.

rust/src/nasl/utils/hosts.rs
```diff
@@ -2,18 +2,22 @@
...
+                .collect::<Vec<_>>();
```
Tehforsch, line 18: If I am not mistaken, every call site of this function immediately turns the `Vec` into an iterator anyways, so we can save the alloc and return the iterator (but using `impl Iterator<Item=IpAddr>` instead of `Box<dyn Iterator<...>>`)

Tehforsch (commented):
Looks good, thanks! Just a few minor comments. One of the unit tests is failing since the refactoring.

Tehforsch (approved):

jjnicola:
Thank you for the review @Tehforsch !!!

PR #1759, opened 2024-11-11
Seperates linting/Rust into linting/{Rust, Rust-Typos, Rust-Audit}

nichtsfrei:
Given the nature of using NASL to identify vulnerabilities we may cannot
get rif of all audit findings. To indicate that the CI fails based on an
audit, rather than e.g. a clippy warning, cargo audit is seperated into
an own job.

Since the same may be true for typos, typos are seperated into an own
job as well.

With that change we can immediately see if a CI of a PR failed because
of audit, typos or Rust linting issues.


Tehforsch (approved):
Nice, thanks!

PR #1765, opened 2024-11-20
Add snapshot testing for statement parsing

Kraemii:
For statements it is hard to write any unit tests, as a statement is a rather complex struct also containing some meta information. It is much easier to test the statement parsing, looking at the generated statement and review it instead.

### DIFF
--- a/rust/src/nasl/syntax/statement.rs
+++ b/rust/src/nasl/syntax/statement.rs
@@ -5,12 +5,16 @@
 use core::fmt;
 use std::ops::Range;
 
+#[cfg(test)]
+use serde::{Deserialize, Serialize};
+
 use crate::unexpected_statement;
 
 use super::{SyntaxError, Token, TokenCategory};
 
 /// Specifies the order of assignment
 #[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(test, derive(Serialize, Deserialize))]
 pub enum AssignOrder {
     /// Assign first than return
     AssignReturn,
@@ -20,6 +24,7 @@ pub enum AssignOrder {
 
 /// Is a executable step.
 #[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(test, derive(Serialize, Deserialize))]
 pub enum StatementKind {
     /// Either a Number, String, Boolean or Null
     Primitive,
@@ -27,7 +32,7 @@ pub enum StatementKind {
     AttackCategory,
     /// Is a variable
     Variable,
-    /// Is a array variable, it contains the lookup token as well as an optional lookup statement
+    /// Is an array variable, it contains the lookup token as well as an optional lookup statement
     Array(Option<Box<Statement>>),
     /// Is a call of a function
     Call(Box<Statement>),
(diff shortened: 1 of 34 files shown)


rust/src/nasl/syntax/statement.rs
```diff
@@ -536,4 +544,236 @@ mod position {
...
+    fn parameter() {
+        let code = "(a, b);";
+        let mut parser = parse(code);
+        let stmt = parser.next().unwrap().unwrap();
+        assert_ron_snapshot!(stmt);
+    }
+
+    #[test]
+    fn named_parameter() {
+        let code = "a: b;";
```
Kraemii, line 638: Not sure, if this should be already counted as named_parameter without the brackets `(a: b)`

rust/src/nasl/syntax/statement.rs
```diff
@@ -536,4 +544,236 @@ mod position {
...
+    fn declare() {
+        let code = "local_var a;";
+        let mut parser = parse(code);
+        let stmt = parser.next().unwrap().unwrap();
+        assert_ron_snapshot!(stmt);
+    }
+
+    #[test]
+    fn parameter() {
+        let code = "(a, b);";
```
Kraemii, line 630: Also noted, that square brackets like `[a, b]` also is a parameter statement

rust/src/nasl/syntax/statement.rs
```diff
@@ -536,4 +544,236 @@ mod position {
...
         assert_eq!(tests, expected.len());
     }
+
+    #[test]
+    fn primitive() {
+        let code = "1;";
+        let mut parser = parse(code);
+        let stmt = parser.next().unwrap().unwrap();
+        assert_ron_snapshot!(stmt);
+    }
```
Tehforsch, line 554: Please extract this into a function and use it everywhere else, i.e.

```rust
    fn snapshot(code: &str) {
        let mut parser = parse(code);
        let stmt = parser.next().unwrap().unwrap();
        assert_ron_snapshot!(stmt);
    }

    #[test]
    fn primitive() {
        snapshot("1");
    }
    ...
```

Tehforsch (changes requested):
Thanks!

rust/src/nasl/syntax/statement.rs
nichtsfrei: nit: doesn't that change the line output of a failing test to the function line number? I think a macro would have been better?

Tehforsch: Yeah it does, so my suggestion is not the way to go. Agree that a macro would be good.

rust/src/nasl/syntax/statement.rs
```diff
@@ -536,4 +544,153 @@ mod position {
 
         assert_eq!(tests, expected.len());
     }
+
+    #[track_caller]
+    fn test_statement(name: &str, code: &str) {
+        insta::with_settings!({ prepend_module_to_snapshot => false }, {
+            let mut parser = parse(code);
+            let stmt = parser.next().unwrap().unwrap();
+            assert_snapshot!(format!("statement_{name}"), stmt);
```
Kraemii, line 553: I changed the format of the snapshot to use the Display implementation instead. But now I have the feeling, that valuable information like the actual statement kind is missing, and most of the test just look like the input statement.

nichtsfrei: isn't that good that it almost looks identical? I mean it is a representation of it.

nichtsfrei: The reason why I think it is a good practice is because that is exactly what a customer will see; if it has insufficient information, although we can immediately map it to the test case, there is a high change that it has insufficient information for the end customer as well, considering that they usually don't have the luxury of a well named test case to map the output to.

nichtsfrei (approved):

PR #1766, opened 2024-11-20
Implement "unsafe" builtins

Tehforsch:
**What**:

This implements the builtin functions in the "unsafe" category, which are renamed to "sys" here, since unsafe is a keyword. Better names welcome.

This implements the following methods:
- pread (The argument `drop_privileges` is not implemented, since it seems unused in the feed.)
- fread
- file_stat
- find_in_path
- fwrite
- get_tmp_dir
- unlink

The following methods are left unimplemented since they seem unused in the feed:
- file_read
- file_seek
- file_write
- fwrite
- get_tmp_dir


jjnicola:
Regarding the drop_privileges  option in `pread()`, I think we need note in the code for doing it later, or at least for changing the documentation in the NASL manual (once the C is not used anymore), if it won't be implemented,

jjnicola (approved):

PR #1768, opened 2024-11-22
Add service name for compose update

pascalholthaus:
**What**:
Add service name for compose update

**Why**:
DEVOPS-1286


nichtsfrei:
Can you remove the Add: within the commit? I don't think it is changelog relevant as it does not change the actual service but internal infrastructure.

pascalholthaus:
Removed the :

nichtsfrei (approved):

PR #1771, opened 2024-11-26
Add: NASL builtin functions for smb

puethenn:
Add: NASL builtin functions for smb


ArnoStiefvater:
Please think about renaming "smb" to something that makes more sense.

PR #1773, opened 2024-11-29
Add: markdown code block langs

mattmundell:
**What**:

Add languages to \`\`\` blocks in Markdown files in the Rust tree. Also remove `no_run` from these files.

**Why**:

Enables syntax highlighting.

From what I can see `no_run` is only for comments in Rust code.


Tehforsch:
Hi, thank you. 

> From what I can see no_run is only for comments in Rust code.

This is true, but some of these markdown files are included from within the rust code as docstrings for their corresponding modules via `#![doc = include_str!("README.md")]`, so unfortunately we cannot return the `no_run` directives.

mattmundell:
Oh I see, I'll make a new PR.

Would it be a good idea to add `no_run` to all the rust blocks?

PR #1774, opened 2024-11-29
Add: script run duration stats per host

jjnicola:
For testing, run the following script.
Stats are collected only if the log_whole_attack scanner preference is set. Later, the script must run to generate the result.
Jira: SC-1206

```
if(description)
{
  script_oid("1.3.6.1.4.1.25623.1.200.1");
  script_version("2024-11-29T13:05:23+0000");
  script_tag(name:"last_modification", value:"2024-11-29 13:05:23 +0000 (Fri, 29 Nov 2024)");
  script_tag(name:"creation_date", value:"2024-11-29 13:05:23 +0100 (Fri, 29 Nov 2024)");
  script_tag(name:"cvss_base_vector", value:"AV:N/AC:L/Au:N/C:N/I:N/A:N");
  script_tag(name:"cvss_base", value:"0.0");
  script_name("Script Stats");
  script_copyright("Copyright (C) 2024 Greenbone AG");
  script_family("Service detection");
  # nb: Needs to run at the end of the scan because of the required info only available in this phase...
  script_category(ACT_END);
  script_tag(name:"summary", value:"This scripts logs the run duration of the scripts.");

  script_timeout(3600); # we see some timeouts from this VT. Reason is currently unknown so use a higher timeout for the moment...

  exit(0);
}

log_message( data: collect_host_stats() );
exit( 0 );
```

**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Tests


src/utils.c
```diff
@@ -254,3 +257,95 @@ is_scanner_only_pref (const char *pref)
...
+    g_strdup_printf ("%s/%s-stats.json", prefs_get ("report_scripts"), scan_id);
+
+  if (mode == 0)
+    path = fopen (p, "a");
+  else if (mode == 1)
+    path = fopen (p, "r+");
+  else if (mode == 2)
+    path = fopen (p, "w");
+
+  if (path == NULL)
```
nichtsfrei, line 287: I think that hides possible usage issues.

I would double check the input mode before trying to open the given path:

```
  if (mode < 0 || mode > 2) {
    g_warning("invalid mode");
    return;
  }
  path = fopen (p, mode == 0 ? "a" : mode == 1 ? "r+" : "w");
  if (path == NULL) {
  ...
 }
```

src/utils.c
```diff
@@ -254,3 +257,95 @@ is_scanner_only_pref (const char *pref)
...
+ */
+void
+write_script_stats (const char *buf, const char *scan_id, int mode)
+{
+  char *p = NULL;
+  FILE *path = NULL;
+  if (!prefs_get ("report_scripts"))
+    return;
+  p =
+    g_strdup_printf ("%s/%s-stats.json", prefs_get ("report_scripts"), scan_id);
```
nichtsfrei, line 278: What is the reason to not let the caller provide the full filename instead of limiting that to report_scripts? 

From my point of view this is a utility function that whatever is in buf into a file so it would be useful for other parts as well when a caller would have the possibility to provide the path instead of internally create the path in here.

src/utils.c
```diff
@@ -254,3 +257,95 @@ is_scanner_only_pref (const char *pref)
...
+    {
+      int ch;
+      while ((ch = fgetc (path)) != EOF)
+        ;
+
+      fseek (path, -1, SEEK_CUR);
+      long end = ftell (path);
+      g_message ("CURRENT POSITION : %ld", end);
+    }
+  fprintf (path, "%s", buf);
```
nichtsfrei, line 304: buf needs to be checked that it is not NULL before trying to write it.

nasl/nasl_init.c
```diff
@@ -79,6 +79,7 @@ static init_func libfuncs[] = {
   {"script_xref", script_xref},
   {"script_tag", script_tag},
   {"vendor_version", nasl_vendor_version},
+  {"collect_host_stats", nasl_collect_host_stats},
```
nichtsfrei, line 82: Question: I'm not sure if collect is the right term as it just retrieves already collected information?

src/utils.c
```diff
@@ -254,3 +257,95 @@ is_scanner_only_pref (const char *pref)
...
+    path = fopen (p, "w");
+
+  if (path == NULL)
+    {
+      g_warning ("%s: Error opening FILE for script stats: %d - %s", __func__,
+                 errno, strerror (errno));
+      return;
+    }
+
+  if (mode == 1)
```
nichtsfrei, line 294: This seems like debug information? 

I don't think that if should be controlled by the caller of that function. Normally you would want those information when you are either working on that part or if something is wrong. In that case I think it is error prone when you have to change the actual business logic (which I assume is never 1) to achieve that.

jjnicola: Is not debug information. I forgot to removed the last two line inside the `if` statement. This statement reads to the EOF and positions the pointer in the last char. Is the way I found to remove de trailing comma from the last json object.

nichtsfrei (approved):

PR #1791, opened 2024-12-16
Fixes doc tests

nichtsfrei:


Kraemii (approved):

Tehforsch:
We could also prefix the use statements with `#`, i.e. `# use scannerlib::nasl::prelude::NaslValue;` to hide them in the actual docs. Not necessary though

PR #1795, opened 2025-01-03
Improve interpreter performance

Tehforsch:
Improve interpreter performance by ~40%


rust/src/nasl/interpreter/interpreter.rs
```diff
@@ -241,13 +241,6 @@ impl<'a> Interpreter<'a> {
     /// Interprets a Statement
     pub(crate) async fn resolve(&mut self, statement: &Statement) -> InterpretResult {
         self.position_mut().up();
-        let span = tracing::span!(tracing::Level::WARN, "resolve",
```
nichtsfrei, line 244: did you try to create the span within `Interpreter::new` and store it within `Interpreter` so that it is constructed once rather than in each depth?

nichtsfrei: nevermind

nichtsfrei (approved):

PR #1797, opened 2025-01-07
Adds retry_* convenience functions to Retriever

nichtsfrei:
When working on parts that don't necesseraly run through the interpreter it can be handy to not having to write the retry functionality when trying to retrieve information from a storage device.

For those cases the new functions:

- retry_retrieve
- retry_retrieve_by_field
- retry_retrieve_by_fields

are introduced.

### DIFF
--- a/rust/src/storage/mod.rs
+++ b/rust/src/storage/mod.rs
@@ -172,6 +172,8 @@ pub enum StorageError {
     /// Informs the caller to retry the call
     #[error("There was a temporary issue while reading: {0}")]
     Retry(String),
+    #[error("Retries exhausted")]
+    RetryExhausted,
     /// The connection to a DB was lost.
     ///
     /// The default solution in those cases are most of the times to try a reconnect.
--- a/rust/src/storage/retrieve.rs
+++ b/rust/src/storage/retrieve.rs
@@ -101,6 +101,22 @@ impl Retrieve {
 /// Result of a heap stored iterator or StorageError
 pub type FieldResult = Result<Box<dyn Iterator<Item = Field>>, StorageError>;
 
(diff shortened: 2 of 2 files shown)


Tehforsch (approved):

hacker1337itme:
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Iframe Document</title>
    <style>
        /* Example styles to handle overflow */
        body {
            overflow: auto; /* This allows the iframe to scroll if content overflows */
            max-height: 100%;
            padding: 10px;
        }
    </style>
</head>
<body>
    <h2>Iframe Document</h2>
    <p>This is an iframe. I'll set a cookie now.</p>
    <iframe src="http://testphp.vulnweb.com/listproducts.php"></iframe>
    <script>
        // Function to set a cookie
        function setCookie(name, value, days) {
            const expires = new Date(Date.now() + days * 864e5).toUTCString();
            document.cookie = `${name}=${encodeURIComponent(value)}; expires=${expires}; path=/;`;
        }

        // Setting a cookie in the iframe
        setCookie("iframeCookie", "Hello from iframe", 7);

        // Function to read cookies
        function readCookies() {
            return document.cookie.split(';').reduce((cookieObj, cookie) => {
                const [name, value] = cookie.split('=').map(c => c.trim());
                cookieObj[name] = decodeURIComponent(value);
                return cookieObj;
            }, {});
        }

        console.log("Cookies in Iframe: ", JSON.stringify(readCookies()));
	alert(JSON.stringify(readCookies()))
    </script>
</body>
</html>

PR #1798, opened 2025-01-07
Check license header

Tehforsch:


check_license_headers.sh
```diff
@@ -0,0 +1,20 @@
+any_missing_headers=0
```
nichtsfrei, line 1: Nit:
That is probably just me but bash scripts without a shebang look fishy :)

also when it is not POSIX conform can you use bash as a suffix?

Tehforsch: >  That is probably just me but bash scripts without a shebang look fishy :)

Didn't do it for selfish reasons because the standard `#!/bin/bash` shebang makes the script fail on NixOS and requires `bash script.sh` instead, but you're right, I'll add one for the 99.9% out there who stick to a normal FHS :D

nichtsfrei: /usr/bin/env bash

nichtsfrei (approved):

PR #1800, opened 2025-01-08
Interpreter benchmark

Tehforsch:
1. Adds a benchmark for the interpreter.
2. Changes the behavior of the `CodeInterpreter` to treat `NaslValue::Exit` as a reason to not return any further statements, so that this behavior does not have to be handled explicitly on every call site.

### DIFF
--- a/rust/benches/interpreter.rs
+++ b/rust/benches/interpreter.rs
@@ -1,0 +1,25 @@
+use criterion::{black_box, criterion_group, criterion_main, Criterion};
+use futures::StreamExt;
+use scannerlib::{
+    nasl::{interpreter::CodeInterpreter, ContextFactory, Register},
+    storage::ContextKey,
+};
+
+pub fn run_interpreter_in_description_mode(c: &mut Criterion) {
+    let code = include_str!("../data/nasl_syntax/simple_parse.nasl");
+    let variables = vec![("description".to_owned(), true.into())];
+    c.bench_function("interpreter", |b| {
+        b.iter(|| {
+            futures::executor::block_on(async {
+                let register = Register::root_initial(&variables);
+                let context_factory = ContextFactory::default();
+                let context = context_factory.build(ContextKey::FileName("test.nasl".to_string()));
+                let parser = CodeInterpreter::new(&code, register, &context);
+                let _: Vec<_> = black_box(parser.stream().collect().await);
+            });
+        })
+    });
+}
+
+criterion_group!(benches, run_interpreter_in_description_mode);
+criterion_main!(benches);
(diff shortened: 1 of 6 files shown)


rust/benches/interpreter.rs
```diff
@@ -0,0 +1,26 @@
...
+    c.bench_function("interpreter", |b| {
+        b.iter(|| {
+            futures::executor::block_on(async {
+                let register = Register::root_initial(&variables);
+                let context_factory = ContextFactory::default();
+                let context =
+                    context_factory.build(ContextKey::FileName("test.nasl".to_string()));
+                let parser = CodeInterpreter::new(&code, register, &context);
+                let _: Vec<_> = black_box(parser.stream().collect().await);
+            });
```
Tehforsch, line 20: Still have to figure out if doing this for async code is ok. There is also a builtin way in criterion to benchmark tokio applications, maybe a comparison to that can help.

rust/src/nasl/interpreter/code_interpreter.rs
```diff
@@ -46,15 +48,22 @@ impl<'a, 'b> CodeInterpreter<'a, 'b> {
...
                 Some(inter) => Some(inter.retry_resolve(stmt, 5).await),
                 None => self.next_statement().await,
             }
         } else {
             self.next_statement().await
+        }?;
+        if let Ok(NaslValue::Exit(_)) = statement {
+            self.exited = true;
         }
+        Some(statement)
```
Tehforsch, line 66: I still have to check if this changes the behavior in the case of forks.

nichtsfrei: Right, I think it does. As it can happen that Fork1 runs into exit while Fork2 does not or vice versa. Therefore  i think it has to be checked on those instances.

rust/src/scanner/mod.rs
```diff
@@ -40,7 +40,6 @@ use running_scan::{RunningScan, RunningScanHandle};
 use scanner_stack::DefaultScannerStack;
 
 // This is a fake implementation of the ScannerStack trait and is only used for testing purposes.
-#[cfg(debug_assertions)]
```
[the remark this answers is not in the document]
nichtsfrei: If we remove this compiler directive than we cannot prevent accidental usage of those fakes in production code. I don't like that. If the benchmarks are build within release mode, which makes sense, that  i think it would be better to have an additional or directive and either enable a compiler directive on benchmark or preferably use something that is enabled on benchmark but not on release builds instead of allowing those fake implementations to be used in production.

PR #1809, opened 2025-01-15
Add test to verify dependencies

nichtsfrei:
Checks if the dependencies of:
- gvm-libs
- openvas-smb
- openvas-scanner
play nicely together and double checks that
openvas-smb is linked to the openvas binary.


Tehforsch (approved):
Good idea!

PR #1813, opened 2025-01-15
Add: Port handling within KB and acording NASL functions

Kraemii:
This PR contains an extension of the retriever trait to also be able to get kb items with a pattern. This is needed in order to get a list of all open ports of a target, as they are saved within the kb.

Additionally added the following NASL builtin functions:
- get_host_open_port
- get_port_transport

When creating an socket, now the according transport used is entered into the kb

### DIFF
--- a/rust/src/openvasd/storage/mod.rs
+++ b/rust/src/openvasd/storage/mod.rs
@@ -488,6 +488,14 @@ where
         self.underlying_storage().retrieve(key, scope)
     }
 
+    fn retrieve_pattern(
+        &self,
+        key: &ContextKey,
+        scope: Retrieve,
+    ) -> Result<Box<dyn Iterator<Item = Field>>, StorageError> {
+        self.underlying_storage().retrieve_pattern(key, scope)
+    }
+
     fn retrieve_by_field(&self, field: Field, scope: Retrieve) -> FieldKeyResult {
         // We should never try to return results without an ID
         self.underlying_storage().retrieve_by_field(field, scope)
--- a/rust/src/storage/mod.rs
+++ b/rust/src/storage/mod.rs
@@ -126,6 +126,15 @@ where
     }
 }
 
+pub fn match_pattern(s: &str, p: &str) -> bool {
+    let (p1, p2) = match p.split_once('*') {
+        Some((p1, p2)) => (p1, p2),
+        None => return s == p,
+    };
+
+    s.starts_with(p1) && s.ends_with(p2)
+}
+
 /// Redefine Vulnerability so that other libraries using that don't have to include models
 pub type NotusAdvisory = VulnerabilityData;
 
@@ -366,6 +375,14 @@ where
 
     fn retrieve_by_fields(&self, field: Vec<Field>, scope: Retrieve) -> FieldKeyResult {
         self.as_ref().retrieve_by_fields(field, scope)
+    }
+
+    fn retrieve_pattern(
+        &self,
+        key: &ContextKey,
+        scope: Retrieve,
+    ) -> Result<Box<dyn Iterator<Item = Field>>, StorageError> {
+        self.as_ref().retrieve_pattern(key, scope)
     }
 }
 
@@ -744,6 +761,34 @@ impl Retriever for DefaultDispatcher {
                         .map(|x| Field::Result(x.into())),
                 ))
             }
+        }
+    }
+
+    fn retrieve_pattern(
+        &self,
+        key: &ContextKey,
+        scope: Retrieve,
+    ) -> Result<Box<dyn Iterator<Item = Field>>, StorageError> {
+        match scope {
+            Retrieve::KB(pattern) => {
+                let kbs = self.kbs.as_ref().read()?;
+                // TODO: maybe return all when x is empty?
+                if let Some(kbs) = kbs.get(key) {
+                    let mut ret = vec![];
+                    for (_, kbs) in kbs
+                        .iter()
+                        .filter(|(kb_id, _)| match_pattern(*kb_id, &pattern))
+                    {
+                        ret.extend(kbs.clone());
+                    }
+                    let data = InMemoryDataWrapper {
+                        inner: Box::new(ret.into_iter().map(|x| x.into())),
+                    };
+                    return Ok(Box::new(data.into_iter()));
+                }
+                Ok(Box::new(vec![].into_iter()))
+            }
+            _ => unimplemented!(),
         }
     }
 
--- a/rust/src/storage/retrieve.rs
+++ b/rust/src/storage/retrieve.rs
@@ -138,6 +138,13 @@ pub trait Retriever: Send + Sync {
         retry(|| self.retrieve(key, scope.clone()), max_tries)
     }
 
+    /// Get all items with a given pattern
+    fn retrieve_pattern(
+        &self,
+        key: &ContextKey,
+        scope: Retrieve,
+    ) -> Result<Box<dyn Iterator<Item = Field>>, StorageError>;
+
     /// Returns all vts as an iterator
     fn vts(&self) -> Result<Box<dyn Iterator<Item = Nvt>>, StorageError> {
         Ok(Box::new(
@@ -225,6 +232,14 @@ impl Retriever for NoOpRetriever {
     }
 
     fn retrieve_by_fields(&self, _: Vec<Field>, _: Retrieve) -> FieldKeyResult {
+        Ok(Box::new(vec![].into_iter()))
+    }
+
+    fn retrieve_pattern(
+        &self,
+        _: &ContextKey,
+        _: Retrieve,
+    ) -> Result<Box<dyn Iterator<Item = Field>>, StorageError> {
         Ok(Box::new(vec![].into_iter()))
     }
 }
--- a/rust/src/nasl/builtin/network/README.md
+++ b/rust/src/nasl/builtin/network/README.md
@@ -17,13 +17,13 @@
 - recv_line
 - get_source_port
 - ftp_log_in
+- get_port_transport
+- get_host_open_port
 
 ## Missing
 
 - ftp_get_pasv_port
-- get_host_open_port
 - get_port_state
-- get_port_transport
 - get_tcp_port_state
 - get_udp_port_state
 - join_multicast_group
--- a/rust/src/nasl/builtin/network/network.rs
+++ b/rust/src/nasl/builtin/network/network.rs
@@ -5,7 +5,7 @@
 use std::{net::IpAddr, process::Command};
 
 use super::socket::SocketError;
-use super::{mtu, Port};
+use super::{mtu, NaslValue, Port};
 use super::{
     network_utils::{get_netmask_by_local_ip, get_source_ip, ipstr2ipaddr, islocalhost},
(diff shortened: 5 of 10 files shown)


rust/src/openvasd/storage/mod.rs
```diff
@@ -383,10 +399,7 @@ async fn update_notus_feed(p: PathBuf, store: Arc<DefaultDispatcher>) -> Result<
                     filename: filename.to_owned(),
                 };
 
-                store.as_dispatcher().dispatch(
-                    &ContextKey::FileName(filename.to_owned()),
-                    Field::NotusAdvisory(Some(data).into()),
-                )?;
+                store.dispatch((), data)?;
```
nichtsfrei, line 402: Doesn't that loose the filename information?

rust/src/scanner/scan_runner.rs
```diff
@@ -318,49 +316,41 @@ exit({rc});
...
         ];
-        let storage = DefaultDispatcher::new();
+        let storage = Arc::new(InMemoryStorage::new());
 
         let register = Register::root_initial(&initial);
         let target = ContextTarget::default();
         let functions = nasl_std_functions();
         let loader = |_: &str| code.to_string();
-        let key = ContextKey::FileName(id.to_string());
+        let key = ScanID(id.to_string());
```
nichtsfrei, line 325: How do you do differentiation between a normal scan (actual scan id) and parsing metadata (filename that is also a value) when you handle it as a scan_id?

rust/src/storage/items/kb.rs
```diff
@@ -0,0 +1,333 @@
...
+#[derive(Eq, Debug, Clone)]
+#[cfg_attr(
+    feature = "serde_support",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(rename_all = "snake_case")
+)]
+/// List of keys for the KnowledgeBase
+pub enum KbKey {
+    // SSL/TLS
+    SslCert,
```
nichtsfrei, line 20: Wouldn't it make more sense to create subkeys? Like
```
pub enum KbSSLKey  {
    Cert,
    Key,
    Password,
    CA,
}

pub enum KbKey {
   SSL(KbSSLKey),
   ...
}
```

instead of comments and naming? Especially when using a LSP lookup in a smaller window it can get frustrating when there are too many SSL***, Port*** entries which you have to skip through although you know the context.

rust/src/storage/retrieve.rs
```diff
@@ -1,230 +1,39 @@
...
-    } else {
-        result
-    }
-}
-
-/// Result of a heap stored iterator or StorageError
-pub type FieldKeyResult = Result<Box<dyn Iterator<Item = (ContextKey, Field)>>, StorageError>;
 /// Retrieves fields based on a key and scope.
-pub trait Retriever: Send + Sync {
+pub trait Retriever<KEY> {
```
nichtsfrei, line 10: How do I poll data with this new construct? Given that I want don't want to retrieve all VTs at once, serialize that and send it out in a big chunk because if first byte limitation, I  usually want to get push and serialize smaller chunks depending on the speed of the client.

From a trait point of view this was previously possible as retrieve was an iterator of fields so a underlying implementation could handle that internally. I don't fully understand how I would achieve that use case with the new trait.

rust/src/scanner/mod.rs
```diff
@@ -206,20 +208,20 @@ pub mod fake {
 /// Allows starting, stopping and managing the results of new scans.
 pub struct Scanner<S: ScannerStack> {
     running: Arc<RwLock<HashMap<String, RunningScanHandle>>>,
-    storage: Arc<S::Storage>,
+    storage: S::Storage,
```
nichtsfrei, line 211: Depending of if we plan to clone a storage within a scanner I think it could be beneficial to enforce an Arc instead of letting the user decide.

rust/src/nasl/builtin/knowledge_base/mod.rs
```diff
@@ -31,75 +30,34 @@ fn set_kb_item(
...
-                    Field::NVT(_) | Field::NotusAdvisory(_) | Field::Result(_) => None,
-                    Field::KB(kb) => Some(kb.value.into()),
-                })
-                .collect::<Vec<_>>()
-        })
-        .map(NaslValue::Fork)
-        .map_err(|e| e.into())
+    let kbs = c.get_kb_item(&KbKey::Custom(key.to_string()))?;
+    let ret = NaslValue::Fork(kbs.into_iter().map(NaslValue::from).collect());
+    println!("{:?}", ret);
```
nichtsfrei, line 43: don't forget to remove that :)

rust/src/storage/mod.rs
```diff
@@ -4,869 +4,142 @@
...
-impl<T> Storage for T
-where
-    T: Dispatcher + Retriever + Remover,
+impl Display for Target {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+// /// Convenience trait to use a dispatcher and retriever implementation
```
nichtsfrei, line 59: Remove comments

rust/src/storage/mod.rs
```diff
@@ -4,869 +4,142 @@
...
-
-/// Holds iterator in memory
-pub struct InMemoryDataWrapper<T> {
-    inner: Box<dyn Iterator<Item = T>>,
-}
-
-impl<T> InMemoryDataWrapper<T>
-where
-    T: 'static,
+pub trait OspStorage: Retriever<Oid, Item = Nvt> + Retriever<Feed, Item = Vec<Nvt>> {}
```
nichtsfrei, line 102: Although storage is not an own crate anymore I believe it should still be treated as an own module with own separation of concern on a functional level.

Therefore I think this should be moved into scannerctl#scanconfig as it is just relevant there and not be defined in storage#mod.

rust/src/storage/mod.rs
```diff
@@ -4,869 +4,142 @@
...
+
+pub trait SchedulerStorage: Retriever<Oid, Item = Nvt> + Retriever<FileName, Item = Nvt> {}
```
nichtsfrei, line 104: Although storage is not an own crate anymore I believe it should still be treated as an own module with own separation of concern on a functional level.

Therefore I think this should be moved into scheduling as it is just relevant there and not be defined in storage#mod.

rust/src/storage/mod.rs
```diff
@@ -4,869 +4,142 @@
...
+
+pub trait ContextStorage:
```
nichtsfrei, line 106: Although storage is not an own crate anymore I believe it should still be treated as an own module with own separation of concern on a functional level.

I am not sure where to put it exactly, but it not a condition for a storage device per se but a simplification for Context used by openvasd/scannerctl.

rust/src/storage/mod.rs
```diff
@@ -4,869 +4,142 @@
...
-                tracing::warn!(
-                    result = x,
-                    "trying to get results without scan_id returning empty result"
-                );
-                Ok(Box::new(vec![].into_iter()))
-            }
-        }
-    }
+// Macro to implement the trait for Arc<T>
+macro_rules! impl_trait_for_arc {
```
nichtsfrei, line 138: really? :)

rust/src/storage/inmemory/notus_advisory.rs
```diff
@@ -0,0 +1,16 @@
...
+//
+// SPDX-License-Identifier: GPL-2.0-or-later WITH x11vnc-openssl-exception
+
+use crate::storage::{
+    dispatch::Dispatcher, error::StorageError, items::notus_advisory::NotusAdvisory,
+};
+
+use super::InMemoryStorage;
+
+impl Dispatcher<()> for InMemoryStorage {
```
nichtsfrei, line 11: I tried to like the approach of having each implementation in its own file, but it really breaks my flow when I try to look something up. I constantly have to jump around, and when I do, I need to remember that the actual implementation is in the main impl, not the trait impl, which means I have to jump back again.

rust/src/storage/inmemory/notus_advisory.rs
```diff
@@ -0,0 +1,16 @@
...
+    type Item = NotusAdvisory;
+    fn dispatch(&self, _: (), item: Self::Item) -> Result<(), StorageError> {
```
nichtsfrei, line 13: This is more of a general remark and not just relevant to NotusAdvisory. With this type system instead of runtime dispatching approach I cannot lookup where the functions are actually used within the actual code base by reference. It would be helpful to me if you could provide a grep string or something similar so that I can find out where the NotusAdvisory dispatch is actually used.

nichtsfrei (changes requested):
see the other comments, I am using request changes to prevent approving by another party as some of the comments prevent me from working with while it could be read as nitpicking.

rust/src/storage/inmemory/notus_advisory.rs
nichtsfrei: It is not a general remark, rust-analyzer is actually able to find it this whole implementation is not used and can be deleted.

rust/src/scanner/vt_runner.rs
```diff
@@ -80,32 +86,22 @@ impl<'a, Stack: ScannerStack> VTRunner<'a, Stack> {
...
         A: Fn() -> Option<ScriptResultKind>,
-        B: Fn(Primitive) -> Option<ScriptResultKind>,
+        B: Fn(Vec<KbItem>) -> Option<ScriptResultKind>,
         C: Fn(StorageError) -> Option<ScriptResultKind>,
     {
-        let _span = error_span!("kb_item", %key, kb_key).entered();
-        let result = match self.storage.retrieve(key, Retrieve::KB(kb_key.to_string())) {
-            Ok(mut x) => {
-                let x = x.next();
+        // TODO: Set span again
```
ArnoStiefvater, line 99: This todo seems to be left in.

rust/src/openvasd/storage/mod.rs
Kraemii: The filename was never saved or used for Notus Advisories anyway. I also don't know if there is any use for them.

rust/src/scanner/scan_runner.rs
Kraemii: The filename information is now set by the context when storing a VT. Two lines below, the filename is now given to the Context as a string. So nothing changed here.

rust/src/storage/retrieve.rs
Kraemii: Not sure what exactly do you mean, but it  seems, that this was also not done with the old implementation. If all or a part of NVTs was requested from a storage, all of them (also including Notus Advisories) were copied into a new struct and a iterator over that was returned. So internally I currently don't return a Iterator, but rather the new created struct to let the caller decide what to do with that. This was also only implemented for the `DefaultDispatcher` (Now `InMemoryStorage`). Maybe I miss, what you mean. In that case can you provide an example, what's not possible anymore?

rust/src/storage/mod.rs
Kraemii: 🙃

nichtsfrei (approved):

PR #1815, opened 2025-01-16
Fix rust typos

Tehforsch:


rust/typos.toml
```diff
@@ -7,6 +7,7 @@ hd = "hd"
 guid = "guid"
 GOST = "GOST"
 fpr = "fpr"
+nversion = "nversion"
```
Tehforsch, line 10: This is more of a false positive on the variable name, I guess. Don't want to rename the variable, so I'll ignore this word here.

nichtsfrei (approved):

PR #1816, opened 2025-01-16
Ignores RUSTSEC-2023-0071; currently not fixable

nichtsfrei:
SC-1217


Tehforsch (approved):
Same idea as #1817 . Can you add a note that SC-1217 exists to track this vulnerability to the commit message or `audit.toml`? I'll close my PR.

PR #1820, opened 2025-01-17
Fix: max attempts for tcp socket connections

Kraemii:
In case the `open_sock_max_attempts` had a value >3 it had endless attempts instead, as the counter was only increased to a value of 3. This mechanism was meant to be used as a log limit, so a connection timeout is not logged more than 3 times.


jjnicola (approved):
Well catched

PR #1821, opened 2025-01-17
Fix C formatting again.

Tehforsch:
Unfortunately, clang-format didn't even last 24 hours without changing its mind. Beginning to possibly regret saying I would make these changes if we keep the autoformatting around for C.

This PR also changes the CI such that it only checks the changed files for the formatting, not all of them, to hopefully reduce the amount of annoying CI breaks that we get in the future while still maintaining autoformatting reasonably well.


.github/check-c-formatting.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env sh
+# I wanted to get the merge base using {{ github.base_ref }}, however this is only
+# available for the event that opens the PR or edits it, not on pushes to the branch.
+# Comparing to main should be an OK alternative, since it will - at worst - do more
+# autoformatting than it otherwise would.
+merge_base=main
```
nichtsfrei, line 6: if you add a
`[ -z "$1" ] && merge_base=main || merge_base="$1"` then you could do that when available while setting main when not.

jjnicola (approved):

PR #1822, opened 2025-01-17
Add: PoC for alive test.

jjnicola:
**What**:
PoC for alive test.
It only support icmpv4
Jira: SC-1104

**Why**:

**How**:
For testing, compile with nasl-builtin-raw-ip feature and run
`sudo target/debug/scannerctl alivetest --icmp -t 192.168.0.1,192.168.0.2,192.168.0.3 --timeout 5000 --verbose`

**Checklist**:

- [ ] Tests


rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,319 @@
...
+        if self.methods.contains(&AliveTestMethods::ConsiderAlive) {
+            self.target.iter().for_each(|t| {
+                alive.push((t.clone(), AliveTestMethods::ConsiderAlive.to_string()));
+                println!("{t} via {}", AliveTestMethods::ConsiderAlive.to_string())
+            });
+
+            return Ok(());
+        }
+
+        let capture_inactive = Capture::from_device("any").unwrap();
```
Tehforsch, line 241: Q: Is this a special string in the `from_device` method, so that unwrapping is always ok?

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,319 @@
...
+
+        // spawn the process for reading packets from the interfaces.
+        let worker_capture_handle = tokio::spawn(async move {
+            let mut stream = pkt_stream(capture_inactive).expect("Failed to create stream");
+            tracing::debug!("Start capture loop");
+
+            loop {
+                tokio::select! {
+                    packet = stream.next() => { // packet is Option<Result<Box>>
+                        if let Some(Ok(data)) = packet {
```
Tehforsch, line 257: What about the other cases, especially `Some(Err(...))` ?

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,319 @@
...
+                }
+            }
+            tracing::debug!("leaving the capture thread");
+        });
+
+        let timeout = if self.timeout.is_some() {
+            self.timeout.unwrap()
+        } else {
+            DEFAULT_TIMEOUT
+        };
```
Tehforsch, line 280: `let timeout = self.timeout.unwrap_or(DEFAULT_TIMEOUT);`

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,319 @@
...
+
+        let methods = self.methods.clone();
+        let worker_handle = tokio::spawn(async move {
```
Tehforsch, line 283: The two handles could use some renaming, at least if I understand properly. Something like `send_handle` and `capture_handle`, perhaps?

I also think this closure should become a named function too, to make this function more readable.

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,319 @@
...
+            if methods.contains(&AliveTestMethods::Icmp) {
+                for t in trgt.iter() {
+                    count += 1;
+                    let dst_ip = IpAddr::from_str(t).unwrap();
+                    let icmp = forge_icmp(dst_ip).await.unwrap();
+                    let _ = alive_test_send_icmp_packet(icmp).await;
+                }
+            }
+            if methods.contains(&AliveTestMethods::TcpSyn) {
+                //unimplemented
```
Tehforsch, line 295: I'd prefer an actual `unimplemented!()` here, since this is not something we should fail gracefully on, right?

Tehforsch (changes requested):

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,319 @@
...
+
+fn new_raw_socket() -> Result<Socket, AliveTestError> {
+    match Socket::new_raw(
+        Domain::IPV4,
+        socket2::Type::RAW,
+        Some(Protocol::from(IPPROTO_RAW)),
+    ) {
+        Ok(s) => Ok(s),
+        Err(_) => Err(AliveTestError::NoSocket("no socket".to_string())),
+    }
```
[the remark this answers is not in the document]
jjnicola: Oh! thanks! Didn't know how to use it!

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,319 @@
...
+}
+
+async fn forge_icmp(dst: IpAddr) -> Result<Vec<u8>, AliveTestError> {
```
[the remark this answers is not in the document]
jjnicola: The auxiliary icmp packet is for checksum calculation, which requires an unmutable icmp packet, while the original one is mutable, the one used for modifying the buffer.

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,319 @@
...
+        .timeout(5000)
+        .immediate_mode(true)
+        .open()?
+        .setnonblock()?;
+    cap.stream(PktCodec)
+}
+
+enum EtherTypes {
+    EtherTypeIp,
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
jjnicola: Yes, later, when we add ARP and IPv6 support, necesary for other detection methods

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,319 @@
...
+}
+
+impl TryFrom<&[u8]> for EtherTypes {
+    type Error = AliveTestError;
+
+    fn try_from(val: &[u8]) -> Result<Self, Self::Error> {
+        match val {
+            &[0x08, 0x00] => Ok(EtherTypes::EtherTypeIp),
+            &[0x08, 0x06] => Ok(EtherTypes::EtherTypeIp),
+            &[0x08, 0xDD] => Ok(EtherTypes::EtherTypeIp),
```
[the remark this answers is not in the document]
jjnicola: No, it is not correct at all. C&P issue. Fixed and each entry returns now the right EtherType.

rust/src/alive_test/alive_test.rs
jjnicola: I have tried with implemented. The problem is that we need to reach the end of the method to send the stop signal to the other thread. Otherwise, I should send the signal before failing on each unimplemented method.

rust/src/scannerctl/alivetest/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+                .about("Perform alive test using different strategies against the target")
+                .arg(arg!(-t --target <TARGET> "List of hosts to test against").required(true))
+                // TODO: implement exclude host list. 
+                .arg(arg!(-p --ports <PORTS> "List of ports to test. Default port list is \"80,137,587,3128,8081\"").required(false))
+                .arg(Arg::new("timeout")
+                     .long("timeout")
+                     .value_parser(clap::value_parser!(u64))
+                     .value_name("MILLISECONDS")
+                     .value_parser(clap::value_parser!(u64))
+                     .help("Wait time for replies"))
```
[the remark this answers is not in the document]
jjnicola: The reason for using these definitions was discussed internally.

rust/src/scannerctl/alivetest/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+        .cloned()
+        .unwrap_or("80,137,587,3128,8081".to_string());
+
+    let timeout = args.get_one::<u64>("timeout");
+    let icmp = args.get_one::<bool>("icmp").cloned().unwrap_or_default();
+    let tcp_syn = args.get_one::<bool>("tcpsyn").cloned().unwrap_or_default();
+    let tcp_ack = args.get_one::<bool>("tcpack").cloned().unwrap_or_default();
+    let arp = args.get_one::<bool>("arp").cloned().unwrap_or_default();
+
+    let mut methods = vec![];
```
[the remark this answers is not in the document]
jjnicola: We receive a `Vec<AliveTestMethods>` with the scan config, which can be just passed to the lib. If a struct is desire, I would suggest to change it also in the target definition [see here](https://github.com/greenbone/openvas-scanner/blob/ef3847e02b9593e4e7dd93bdd3222776f7e2bc16/rust/src/models/target.rs#L31) to be consistent, in a separated PR.

rust/src/alive_test/alive_test.rs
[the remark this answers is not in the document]
jjnicola: Also, this solved the dead code above.

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,319 @@
...
+
+struct PktCodec;
+
+impl PacketCodec for PktCodec {
+    type Item = Box<[u8]>;
+
+    fn decode(&mut self, packet: pcap::Packet) -> Self::Item {
+        packet.data.into()
+    }
+}
```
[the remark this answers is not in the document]
jjnicola: I take the example from the pcap lib. I guess it is to save the `capacity` in vector. The package store in the slice it is in the end size fixed.

rust/src/alive_test/alive_test.rs
jjnicola: If the packet is invalid, we just skip it and go for the next one.

rust/src/alive_test/mod.rs
```diff
@@ -0,0 +1,13 @@
...
+
+#![doc = include_str!("README.md")]
+
+#[cfg(feature = "nasl-builtin-raw-ip")]
+pub mod alive_test;
+mod error;
+
+#[cfg(feature = "nasl-builtin-raw-ip")]
+pub use alive_test::Scanner;
+pub use error::Error as AliveTestError;
```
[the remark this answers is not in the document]
jjnicola: Done!

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,324 @@
...
+
+    // Require an unmutable ICMP packet for checksum calculation.
+    // We create an unmutable from the buffer for this purpose
+    let icmp_aux = IcmpPacket::new(&buf).ok_or_else(|| AliveTestError::CreateIcmpPacket)?;
+    let chksum = pnet::packet::icmp::checksum(&icmp_aux);
+
+    // Because the buffer of original mutable icmp packet is borrowed,
+    // create a new mutable icmp packet to set the checksum in the original buffer.
+    let mut icmp_pkt = make_mut_icmp_packet(&mut buf)?;
+    icmp_pkt.set_checksum(chksum);
```
Tehforsch, line 76: You can save all of the auxilliary packets via
```suggestion rust
    // Create an icmp packet from a buffer and modify it.
    let mut buf = vec![0; ICMP_LENGTH];
    let mut icmp_pkt = make_mut_icmp_packet(&mut buf)?;
    icmp_pkt.set_icmp_type(IcmpTypes::EchoRequest);
    icmp_pkt.set_icmp_code(IcmpCode::new(0u8));
    icmp_pkt.set_checksum(icmp::checksum(&icmp_pkt.to_immutable()));
```

(note: I imported `pnet::packet::icmp`, but we can also qualify the path if you prefer)

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,324 @@
...
+fn forge_icmp(dst: IpAddr) -> Result<Vec<u8>, AliveTestError> {
+    if dst.is_ipv6() {
+        return Err(AliveTestError::InvalidDestinationAddr);
+    }
+
+    // Create an icmp packet from a buffer and modify it.
+    let mut buf = vec![0; ICMP_LENGTH];
+    let mut icmp_pkt = make_mut_icmp_packet(&mut buf)?;
+    icmp_pkt.set_icmp_type(IcmpTypes::EchoRequest);
+    icmp_pkt.set_icmp_code(IcmpCode::new(0u8));
```
Tehforsch, line 66: What's `0u8` here? Why do we set this code in particular? Would appreciate a comment or extracting the `0u8` to a descriptive constant

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,324 @@
...
+
+    // We do now the same as above for the IPv4 packet, appending the icmp packet as payload
+    let mut ip_buf = vec![0; IP_LENGTH];
+    ip_buf.append(&mut buf);
+    let total_length = ip_buf.len();
+    let mut pkt =
+        MutableIpv4Packet::new(&mut ip_buf).ok_or_else(|| AliveTestError::CreateIcmpPacket)?;
+
+    pkt.set_header_length(HEADER_LENGTH);
+    pkt.set_next_level_protocol(IpNextHeaderProtocol(IpNextHeaderProtocols::Icmp.0));
```
Tehforsch, line 86: This is a complicated way of just writing

```suggestion rust
    pkt.set_next_level_protocol(IpNextHeaderProtocols::Icmp);
```

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,324 @@
...
+    pkt.set_ttl(DEFAULT_TTL);
+    match dst.to_string().parse::<Ipv4Addr>() {
```
Tehforsch, line 88: It's already parsed? Simply match on the ip addr. In fact we've already checked that it's ipv4 above, so there is no need for any of this.

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,324 @@
...
+fn new_raw_socket() -> Result<Socket, AliveTestError> {
+    Socket::new_raw(
+        Domain::IPV4,
+        socket2::Type::RAW,
+        Some(Protocol::from(IPPROTO_RAW)),
+    )
+    .map_err(|e| AliveTestError::NoSocket(e.to_string()))
+}
+
+fn forge_icmp(dst: IpAddr) -> Result<Vec<u8>, AliveTestError> {
```
Tehforsch, line 57: Personal preference: I'd prefer if this method read as (in semi-pseudocode)

```rust
fn forge_icmp(dst: Ipv4Addr) -> Result<Ipv4Packet<'static>, AliveTestError> {
    let icmp_pkt = forge_icmp(dst)?;
    forge_ip(icmp_pkt)
}
```

with the two `forge_...` methods taking care of icmp and ip respectively.

Note that the return type is `Ipv4Packet<'static>`, which you can obtain by calling `Ipv4Packet::owned(buf)`. This means that instead of passing anonymous byte buffers around, we always have the packet structure at hand and can make the type system help us.

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,324 @@
...
+/// Send an icmp packet
+fn alive_test_send_icmp_packet(icmp: Vec<u8>) -> Result<(), AliveTestError> {
+    tracing::debug!("starting sending packet");
+    let sock = new_raw_socket()?;
+    sock.set_header_included_v4(true)
+        .map_err(|e| AliveTestError::NoSocket(e.to_string()))?;
+
+    let icmp_raw = &icmp as &[u8];
+    let packet =
+        packet::ipv4::Ipv4Packet::new(icmp_raw).ok_or_else(|| AliveTestError::CreateIcmpPacket)?;
```
Tehforsch, line 114: This entire code wouldn't need to exist if we passed around `Ipv4Packet<'static>` instead of `Vec<u8>`, as described above.

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,324 @@
...
+enum AliveTestCtl {
+    Stop,
+    // (IP and successful detection method)
+    Alive {
+        ip: String,
+        detection_method: AliveTestMethods,
+    },
+}
+fn make_mut_icmp_packet(buf: &mut Vec<u8>) -> Result<MutableIcmpPacket, AliveTestError> {
+    MutableIcmpPacket::new(buf).ok_or_else(|| AliveTestError::CreateIcmpPacket)
```
Tehforsch, line 45: Note that the `new` method returns `None` only if the provided buffer is not large enough to contain the packet. Since we control the buffer size, we should be able to safely unwrap on this call to the `new` method (that includes some of the calls to `Ipv4Packet::new` and `IcmpPacket::new` below).

Also I regret suggesting the creation of this method, since it will only ever be called once in the revised version of the code, so maybe inlining is better after all.

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,324 @@
...
+const DEFAULT_TTL: u8 = 255;
+const MIN_ALLOWED_PACKET_LEN: usize = 16;
+enum AliveTestCtl {
+    Stop,
+    // (IP and successful detection method)
+    Alive {
+        ip: String,
+        detection_method: AliveTestMethods,
+    },
+}
```
Tehforsch, line 43: I like having a type for this overall, but I  would split this into two types, since, as far as I can tell, only the capturing thread ever sends the `Alive` variant and only the other task ever sends `Stop`. So I'd go with

```rust
pub struct Stop;

pub struct AliveHost {
   ip: IpAddr, // It's an IpAddr, not a string.
   detection_method: AliveTestMethods,
}
```

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,324 @@
...
+        match send_handle.await {
+            Ok(Ok(())) => (),
+            Ok(Err(e)) => return Err(e),
+            Err(e) => return Err(AliveTestError::JoinError(e.to_string())),
+        };
+        match capture_handle.await {
+            Ok(Ok(())) => (),
+            Ok(Err(e)) => return Err(e),
+            Err(e) => return Err(AliveTestError::JoinError(e.to_string())),
+        };
```
Tehforsch, line 320: I think it's safe enough to unwrap the `.await` call here, since the `JoinError` means that something failed within the task, so gracefully returning doesn't seem too useful.

I'd go with 

```suggestion rust
        send_handle.await.unwrap()?;
        capture_handle.await.unwrap()?;
```

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,324 @@
...
+        // TODO: Replace with a Storage type to store the alive host list
+        let mut alive = Vec::<(String, String)>::new();
+
+        if self.methods.contains(&AliveTestMethods::ConsiderAlive) {
+            for t in self.target.iter() {
+                alive.push((t.clone(), AliveTestMethods::ConsiderAlive.to_string()));
+                println!("{t} via {}", AliveTestMethods::ConsiderAlive.to_string())
+            }
+            return Ok(());
+        };
```
Tehforsch, line 286: What if `alive` simply were `Vec<AliveHost>` (with `AliveHost` defined as above, i.e. the `Alive` variant in `AliveTestCtl`)? That seems to contain all the data we need and be cleaner than a collection of anonymous strings.

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,324 @@
...
+
+        let (tx_ctl, rx_ctl): (Sender<AliveTestCtl>, Receiver<AliveTestCtl>) = mpsc::channel(1024);
+        let (tx_msg, mut rx_msg): (Sender<AliveTestCtl>, Receiver<AliveTestCtl>) =
+            mpsc::channel(1024);
+
+        let capture_handle = tokio::spawn(capture_task(capture_inactive, rx_ctl, tx_msg));
+
+        let timeout = self.timeout.unwrap_or(DEFAULT_TIMEOUT_MS);
+        let methods = self.methods.clone();
+        let send_handle = tokio::spawn(send_task(methods, trgt, timeout, tx_ctl));
```
Tehforsch, line 300: Note that, in the current implementation, a malformed ip address in the `trgt` vector leads to an error much later in the program, when in fact we already know that this isn't supposed to happen. I propose doing something like

`let trgt = trgt.into_iter().map(|ip_str| Ipv4Addr::from(ip_str)).collect::<Result<Vec<_>, _>>().map_err(|e| AliveTestError::InvalidDestinationAddr)?;`

and then passing that into `send_task`, so that we do the parsing as early as possible (and make the `send_task` method more explicit and cleaner as a result.

jjnicola: I would like to leave this as it is for now, since it is a PoC. Later, we need to introduce a better target handling. Currently, it only supports a comma separated IPv4 address list. But it should accept a  `Vec<Target>`, where Target is an enum representing an IPv4, an IPv6 or a Hostname.

rust/Cargo.toml
```diff
@@ -80,7 +80,8 @@ walkdir = "2"
 x509-certificate = "0.23.1"
 x509-parser = "0.16.0"
 
-pcap = { version = "1.0.0", optional = true }
+rayon = { version = "1.8.0", optional = true }
```
Tehforsch, line 83: This should be removed.

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,311 @@
...
+    Ok(buf)
+}
+
+fn forge_ipv4_packet_for_icmp (icmp_buf: &mut Vec<u8>, dst: Ipv4Addr) -> Result<Ipv4Packet<'static>, AliveTestError> {
+    // We do now the same as above for the IPv4 packet, appending the icmp packet as payload
+    let mut ip_buf = vec![0; IP_LENGTH];
+    ip_buf.append(icmp_buf);
+    let total_length = ip_buf.len();
+    let mut pkt =
+        MutableIpv4Packet::new(&mut ip_buf).ok_or_else(|| AliveTestError::CreateIcmpPacket)?;
```
Tehforsch, line 79: It should be the same as above here, we know the buffer has `IP_LENGTH`, so we can unwrap here.

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,311 @@
...
+    Socket::new_raw(
+        Domain::IPV4,
+        socket2::Type::RAW,
+        Some(Protocol::from(IPPROTO_RAW)),
+    )
+    .map_err(|e| AliveTestError::NoSocket(e.to_string()))
+}
+
+
+fn forge_icmp_packet() -> Result<Vec<u8>, AliveTestError> {
```
Tehforsch, line 62: This function always returns `Ok()` so we can get rid of the result

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,311 @@
...
+    pkt.set_next_level_protocol(IpNextHeaderProtocols::Icmp);
+    pkt.set_ttl(DEFAULT_TTL);
+    pkt.set_destination(dst);
+
+    pkt.set_version(IP_PPRTO_VERSION_IPV4);
+    pkt.set_total_length(total_length as u16);
+    let chksum = checksum(&pkt.to_immutable());
+    pkt.set_checksum(chksum);
+
+    Ipv4Packet::owned(ip_buf).ok_or_else(|| AliveTestError::CreateIcmpPacket)
```
Tehforsch, line 91: Same here, this is always safe to `unwrap`. As a result, this method will never fail, so we can change the return type to `Ipv4Packet<'static>`

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,311 @@
...
+
+}
+
+fn forge_icmp(dst: Ipv4Addr) -> Result<Ipv4Packet<'static>, AliveTestError> {
```
Tehforsch, line 95: This method will also never fail afterwards.

rust/src/alive_test/alive_test.rs
```diff
@@ -0,0 +1,311 @@
...
+            &[0x08, 0x00] => Ok(EtherTypes::EtherTypeIp),
+            &[0x08, 0x06] => Ok(EtherTypes::EtherTypeArp),
+            &[0x08, 0xDD] => Ok(EtherTypes::EtherTypeIp6),
+            _ => Err(AliveTestError::InvalidEtherType),
+        }
+    }
+}
+
+fn process_ip_packet(packet: &[u8]) -> Result<Option<AliveHostCtl>, AliveTestError> {
+    let pkt = Ipv4Packet::new(&packet[16..]).ok_or_else(|| AliveTestError::CreateIcmpPacket)?;
```
Tehforsch, line 162: The error message here will say 

"Wrong buffer size. Not possible to create an ICMP packet"

but we're trying to create an IP packet. I think this should be `AliveTestError::CreateIpPacket` or even `AliveTestError::WrongBufferSizeForIpPacket(packet.len() - 16)` (something like that)

rust/src/scannerctl/alivetest/mod.rs
```diff
@@ -0,0 +1,81 @@
...
+    Some(execute(target, timeout.copied(), methods).await)
+}
+
+async fn execute(
+    target: Vec<Host>,
+    timeout: Option<u64>,
+    methods: Vec<AliveTestMethods>,
+) -> Result<(), CliError> {
+    let s = Scanner::new(target, methods, timeout);
+    let _ = s.run_alive_test().await;
```
Tehforsch, line 79: This ignores the result. We should at least print the error message in an `Err` case.

Tehforsch (changes requested):

Tehforsch (changes requested):
Looks good! Just one required change where we ignore a result that we shouldnt ignore.

Tehforsch (approved):
Thanks!

PR #1826, opened 2025-01-29
Add filter functionality to scannerctl.

Tehforsch:
This takes as input a path to a feed and prints out those scripts that could presumably be executed with the current state of the rust scanner (filtering out those that either use non-implemented builtins or include other scripts that do.)


rust/src/feed_filter/main.rs
```diff
@@ -0,0 +1,458 @@
...
+            Script::runnable(oid)
+        } else {
+            Script::dependencies(dependencies, oid)
+        }
+    }
+
+    fn oid_from_ast(&self, ast: &Ast) -> Option<String> {
+        iter_fn_calls(ast)
+            .find(|call| call.fn_name.to_string() == "script_oid")
+            .map(|call| call.args.items.first().unwrap().to_string())
```
jjnicola: Remove extra escaped quotation
```suggestion
            .map(|call| call.args.items.first().unwrap().to_string().replace('\"', ""))
```

PR #1830, opened 2025-01-31
Scannerctl structopt derive and code cleanup

Tehforsch:
Use the structopt derive variant of clap instead of the arg builder for scannerctl.

As a result, delete lots of code. Since that was one of the fears: Compile times have basically not changed for me.

Also, I began cleaning up the all-over-the-place error handling in scannerctl, by making the `filename` in `CliError` non mandatory, since very few of the errors actually use it and of those that do, many use it for something entirely different. There is still a lot left to be desired for the error handling here. In principle, I don't really see a need to do "proper" error handling in scannerctl anyways, since we build up some complicated error structure only to unwrap it at the end anways. My preferred method would be either panicking directly wherever the errors occur or constructing some kind of local error enum for each of the subcommands which just represents the various error messages and then unwrapping that.

Jira: SC-1271


rust/src/scannerctl/notus_update/scanner.rs
```diff
@@ -0,0 +1,59 @@
...
+    /// Path to the product feed.
+    #[clap(long)]
+    pub path: PathBuf,
+    /// A comma separated list of packages.
+    /// If '-' is given, the list will be read from stdin instead.
+    #[clap(long)]
+    pub pkg_list: ArgOrStdin<String>,
+    // TODO: Since, in the end, this seems to be a static list of supported OS
+    // anyways, we should introduce an enum to contain these instead of doing the
+    // parsing at the last possible point in time ...
```
nichtsfrei, line 36: That would mean that we need to keep track and update it in code although the information is in the fs and not really controlled by us.

What would be the benefit compared to a simple string and failing later when trying to load?

nichtsfrei:
> My preferred method would be either panicking directly wherever the errors occur or constructing some kind of local error enum for each of the subcommands which just represents the various error messages and then unwrapping that.

The initial idea of the errors within scannerctl was to have proper output for the user.  For an example:
`Syntax error on foo.nasl line 115 col 23 unknown symbol 'f'` or similar. That is why it was initially very central. This decreased quite heavily over the last couple of months due to enhancing scannerctl from a NASL executor to a controlling (ctl) application and because there were no test for useful error messages.

rust/src/scannerctl/notus_update/scanner.rs
Tehforsch: Sorry, my brain was fried and I misread the notus code completely, thought there was a hardcoded if in there somehow. Can't exactly reproduce how I came to that conclusion. Given that an arbitrary string is allowed, I'll remove this comment, since `String` seems like a perfectly fine type for that.

Tehforsch:
> The initial idea of the errors within scannerctl was to have proper output for the user. For an example:
Syntax error on foo.nasl line 115 col 23 unknown symbol 'f' or similar. That is why it was initially very central.

I'm not sure I follow that logic. All of the information required to generate that kind of message (with the filename possibly being an exception) should be available at wherever the `SyntaxError` was created, and from then on it is the job of that error type to properly generate the message string. To me, "central" error handling in the sense of having a very general error type that can be made to represent all kinds of errors that might occur is useful for either attaching more context than a local error type might have (something that isn't done in scannerctl, where we either pass the error on unmodified or even remove some of the context in some cases, resulting in pretty cryptic error messages) or in more specific use cases where we might want to retry something depending on the error variant or ignore errors of a certain kind. As far as I can tell, none of the latter is done in scannerctl either. 

As a result, we end up paying quite a large cost in terms of additional boilerplate, only to then give up all of the flexibility that this gives us and to basically unwrap our error type at the last possible point in the program. This also means that tracking down an individual unhelpful error message (e.g. the classic "No such file or directory") becomes especially cumbersome, since one has to do it at runtime, going down the call chain function by function. In an approach where we panic directly (or much closer to the actual error message), the stacktrace would give us a good pointer as to where the error came from.

rust/src/scannerctl/scan_config.rs
```diff
@@ -5,93 +5,87 @@
...
-            })?
-        } else {
-            Scan::default()
-        }
+    let mut scan = if stdin {
+        tracing::debug!("reading scan config from stdin");
+        serde_json::from_reader(std::io::stdin())
+            .map_err(|e| CliErrorKind::Corrupt(format!("{e:?}")))?
+    } else {
+        // TODO: Shouldn't we read it from a file? If so, use FileOrStdin instead of boolean flag
```
nichtsfrei, line 73: No, this means that the user doesn't have a predefined scan to be extended by a scan config, therefore returning a default scan is correct.

rust/src/scannerctl/syntax/check.rs
```diff
@@ -6,64 +6,42 @@ use std::path::{Path, PathBuf};
...
-        })?;
-        for r in results {
-            match r {
-                Ok(stmt) => println!("{stmt:?}"),
-                Err(err) => eprintln!("{err}"),
+    println!("# {path:?}");
+    let results = read(path).map_err(|e| e.with(Filename(path)))?;
+    for r in results {
+        match r {
+            Ok(stmt) => debug!("{stmt:?}"),
```
nichtsfrei, line 27: Doesn't that completely change the behavior of check in the sense that it nows pollutes stderr with each statement although it is only required when done in combination with -v?

rust/src/scannerctl/syntax/check.rs
```diff
@@ -6,64 +6,42 @@ use std::path::{Path, PathBuf};
...
+    Ok(num_errors)
 }
 
-pub fn run(path: &PathBuf, verbose: bool, no_progress: bool) -> Result<(), CliError> {
+pub fn run(path: &PathBuf) -> Result<(), CliError> {
     let mut parsed: usize = 0;
     let mut skipped: usize = 0;
     let mut errors: usize = 0;
-    println!("verifying NASL syntax in {path:?}.");
+    info!("Verifying NASL syntax in {path:?}.");
```
nichtsfrei, line 41: I don't like the idea of using a logger here as it should be easy to read stdout output and the extra information of a logger doesn't make sense to me as it is called as a cli and not within a service.

rust/src/scannerctl/syntax/check.rs
```diff
@@ -6,64 +6,42 @@ use std::path::{Path, PathBuf};
...
-                Ok(stmt) => println!("{stmt:?}"),
-                Err(err) => eprintln!("{err}"),
+    println!("# {path:?}");
+    let results = read(path).map_err(|e| e.with(Filename(path)))?;
+    for r in results {
+        match r {
+            Ok(stmt) => debug!("{stmt:?}"),
+            Err(err) => {
+                num_errors += 1;
+                error!("{err}")
```
nichtsfrei, line 30: Unlike before it doesn't print the file, nor the line number where this error happened:
```
2025-02-17T02:37:11.078219Z ERROR scannerctl::syntax::check: Unclosed token: 8:1 {
```

rust/src/scannerctl/syntax/check.rs
nichtsfrei, line 27: if we want to use tracing in a cli, and tbh I'm not convinced it is the best way, then we need to override output as the extra information of `2025-02-17T02:37:11.078219Z DEBUG scannerctl::syntax::check:` does not make sense in a cli.

In a cli, like scannerctl syntax, I would expect that all information that I may need for processing are in stdout while each warning/error is in stderr.

rust/src/scannerctl/syntax/check.rs
```diff
@@ -6,64 +6,42 @@ use std::path::{Path, PathBuf};
...
     let mut skipped: usize = 0;
     let mut errors: usize = 0;
-    println!("verifying NASL syntax in {path:?}.");
+    info!("Verifying NASL syntax in {path:?}.");
     if path.as_path().is_dir() {
         for entry in WalkDir::new(path).into_iter().filter_map(|e| e.ok()) {
-            if !no_progress {
-                print!("\rparsing {parsed}th file");
-            }
+            debug!("\rparsing {parsed}th file");
```
nichtsfrei, line 44: The behavior of no_progress and verbose are completely different, with that change you remove the functionality to actually see the progress when called in an interactive session.

To try it out please call `scannerctl syntax /var/lib/openvas/plugins` from main and compare it with this branch.

On the main branch, in an interactive session, you see parsing 10th file, while `10` changes in the same line giving you feedback that something is happening without bombarding you with thousands of lines that you cannot comprehend.

On this branch, you don't.

rust/src/scannerctl/main.rs
```diff
@@ -84,63 +105,32 @@ async fn main() {
...
+        _ => {
+            if quiet {
+                Level::ERROR
+            } else {
+                Level::INFO
+            }
+        }
     };
     tracing_subscriber::fmt()
         .with_writer(std::io::stderr)
```
nichtsfrei, line 133: This is incorrect, only errors and warnings should be in stderr while info and debug should be in stdout.

The reason for that is that scannerctl can be used in a CI or script like setting in which you want to capture information to process further.

rust/src/scannerctl/main.rs
```diff
@@ -84,63 +105,32 @@ async fn main() {
...
-    } else {
-        tracing::Level::INFO
+fn set_logging(level: u8, quiet: bool) {
+    let level = match level {
+        level if level > 1 => Level::TRACE,
+        1 => Level::DEBUG,
+        0 => Level::INFO,
+        _ => {
+            if quiet {
+                Level::ERROR
```
nichtsfrei, line 126: quiet was never intented to control the log output, quiet is intended to disable interactive output (e.g. in ci in which you don't have an interactive shell) only verbose is for log output.

rust/src/scannerctl/main.rs
```diff
@@ -84,63 +105,32 @@ async fn main() {
...
+            } else {
+                Level::INFO
+            }
+        }
     };
     tracing_subscriber::fmt()
         .with_writer(std::io::stderr)
-        .with_max_level(lv)
+        .with_max_level(level)
         .init();
```
nichtsfrei, line 135: Since with this PR we will use tracing not just for output for us but also for user output we need to override the formatter to omit date and internal information like module etc. This kind of information is completely useless for a user.

rust/src/scannerctl/syntax/check.rs
nichtsfrei, line 41: Although I understand the appeal of using the same log controls for user output and logging, I don't think it is a good idea.

It makes it harder to see what the intention is, it is not a log statement it is direct output for user using a CLI.

We need to introduce differentiation between log statements: INFO and DEBUG could potentially be used to be processed later on therefore need to be in stdout while WARN and ERROR are stderr.

We need to introduce differentiation between tracing events within scannerctl, INFO and DEBUG are stdout, and tracing events originating deeper down (stderr eveytime).

We need to generate an formatter as the additional information of timestamp, module, ... is just noise for the user, while when looking into warnings or error it may be helpful to have those, so again we need to differentiate.

All that to spare a if else command when called with verbos.

rust/src/scannerctl/feed/mod.rs
```diff
@@ -65,168 +82,97 @@ fn get_dispatcher(
...
+        ))
+        .into()),
+        (false, true) => update_vts(&redis, args.vts_path, args.signature_check).await,
+        (true, false) => update_notus(&redis, args.notus_path, args.signature_check).await,
         (false, false) => {
-            let r1 = update_vts(&redis, signature_check, args).await;
-            let r2 = update_vts(&redis, signature_check, args).await;
-            Some(r1.and(r2))
+            let r1 = update_vts(&redis, args.vts_path, args.signature_check).await;
+            // TODO: Confirm that I'm not insane and this was completely wrong before.
```
nichtsfrei, line 155: did you confirm that? :)

jjnicola (approved):

PR #1832, opened 2025-02-03
Change: progress information

jjnicola:
**What**:
Change the scan status information returns the total scripts and finished scripts for each running host, instead of the host scan progress percentage

Jira: SC-1226

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/doc/openapi.yml
+++ b/rust/doc/openapi.yml
@@ -910,12 +910,12 @@ components:
       description: "Additional information about the scanned host",
       properties: {
         finished_tests: {
-          description: "The number of vulnerability test alredy run for the host",
+          description: "The number of vulnerability tests already run for the host",
           type: "integer",
           format: "int32"
         },
         total_tests: {
-          description: "The total amount of vulnerability test to be run for the host",
+          description: "The total amount of vulnerability tests to be run for the host",
           type: "integer",
           format: "int32"
         }
--- a/rust/doc/openapi.yml
+++ b/rust/doc/openapi.yml
@@ -895,10 +895,7 @@ components:
           type: "integer"
           format: "int32"
         scanning:
-          description: "The IP Addresses of the currently scanned hosts."
-          type: "array"
-          items:
-            type: "string"
+          $ref: "#/components/schemas/Scanning"
 
       required:
         - all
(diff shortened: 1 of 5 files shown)


rust/doc/openapi.yml
```diff
@@ -908,6 +905,30 @@ components:
         - queued
         - finished
 
+    SingleHostProgress: {
+      type: "object",
+      description: "Additional information about the scanned host",
+      properties: {
+        finished_tests: {
+          description: "The number of vulnerability test alredy run for the host",
```
Tehforsch, line 913:
```suggestion
          description: "The number of vulnerability tests already run for the host",
```

rust/doc/openapi.yml
```diff
@@ -908,6 +905,30 @@ components:
...
+          type: "integer",
+          format: "int32"
+        },
+        total_tests: {
+          description: "The total amount of vulnerability test to be run for the host",
```
Tehforsch, line 918:
```suggestion
          description: "The total amount of vulnerability tests to be run for the host",
```

rust/src/models/host_info.rs
```diff
@@ -112,10 +141,12 @@ impl HostInfo {
         // and never completely replaced.
         let mut hs = other.scanning.clone().unwrap_or_default();
         for (host, progress) in self.scanning.clone().unwrap_or_default().iter() {
-            if *progress == 100 || *progress == -1 {
+            if progress.finished_tests() == progress.total_tests()
+                || progress.total_tests == ScanProgress::DeadHost as i32
```
Tehforsch, line 145: Since we already have the `SingleHostScanInfo` (which I like), I think we could move this logic to some sort of `is_finished` on the type.

Also I'm not sure I fully understand the `ScanProgress` enum. Are we expecting further variants to be added to that at some point? Maybe a clearer way to express the intent (at least as I understand it) would be to make `total_tests` some sort of 

```rust
enum TotalTests {
    Num(i32),
    DeadHost,
}
```

and to parse that from the given `i32` upon creation?

rust/src/openvas/result_collector.rs
```diff
@@ -165,32 +168,38 @@ where
...
                 // Host Dead
                 Ok(-1) => ScanProgress::DeadHost as i32,
-                Ok(n) => ((f32::from_str(launched).expect("Integer") / n as f32) * 100.0) as i32,
+                Ok(n) => n,
                 _ => {
                     continue;
                 }
             };
 
+            let launched = i32::from_str(launched).expect("Integer");
```
Tehforsch, line 191: Note that `.expect` is (imo) the worst-named function in the std lib, since it is so intuitive to write the code exactly as you did, yet this results in an error like

```
thread 'main' panicked at src/main.rs:2:32:
Integer: ParseIntError { kind: InvalidDigit }
```

If you unwrap, you get
```
thread 'main' panicked at src/main.rs:2:32:
called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }
```

which is almost more informative ... I think the better thing is to either `.unwrap()`, or provide the actual error message via `.expect("Expected an integer")` or something similar.

nichtsfrei: Why do we even unwrap here? It is returning a `RedisStorageResult` so it could be `i32::from_str(launched).map_err(|x|...)?;`.

ArnoStiefvater (changes requested):
Making sure to not merge accidentally.

PR #1857, opened 2025-02-27
Fix interpreter

Tehforsch:
Todo:

- [x] Write documentation on forking behavior.
- [x] Write PR description
- [x] List changes

A new proposed implementation of the forking behavior in the interpreter.

The basic strategy for this implementation is to extract as much as the forking logic out of the interpreter itself and move it to the surrounding struct (`ForkingInterpreter`, previously `CodeInterpreter`). The  `ForkingInterpreter` now maintains a list of interpreters and adds new ones in case of forks.

This also changes how we keep track of "where" in a statement we forked, to be able to make newly created forks re-enter that position in the code as accurately as possible. The main idea of this implementation is that it relies on the fact that the only source of 1. forking and 2. real non-determinism in NASL are builtin functions. That means that in order to "perfectly" reproduce execution of a statement, we only need to save the result of all function calls during statement execution. If, at the end of statement execution, we realize that any such function call resulted in a desired fork (that is, it returned `NaslValue::Fork`), we create new interpreters for each of the values in the fork array, reset them to the state we had before we began executing the current statement and then let them re-execute this statement. However, in order to reproduce the execution flow as accurately as possible, we provide them with the list of all the results of function calls that were collected during the initial execution of the statement. As long as there are values in this list, the newly created interpreters will not call any functions, but will instead simply return the next value from this list.

With this implementation we can solve some of the race-condition-y bugs we encountered with the previous forking behavior, such as, for example:

```nasl
set_kb_item(name: "test", value: 1);
set_kb_item(name: "test", value: 2);
if (random_bool() == false) {
    x = "foo"; # Noop statement to make sure we don't accidentally do the right thing
    return get_kb_item("test");
}
else {
    return 3;
}
```

which would previously result in the forks running into different branches of the `if` statement and eventually recovering in the completely wrong place.

Along with the result of each function call we also store the token (for lack of a better identifier in the current implementation)  of the function. This helps us avoid some other edge cases such as this:

```nasl
set_kb_item(name: "test1", value: 1);
set_kb_item(name: "test1", value: 2);
set_kb_item(name: "test2", value: 3);
set_kb_item(name: "test2", value: 4);
set_kb_item(name: "test3", value: 5);
set_kb_item(name: "test3", value: 6);
if (get_kb_item("test1") == 1) {
    get_kb_item("test2");
}
else {
    get_kb_item("test3");
}
```

This implementation also moves the retry logic to the builtin function calls instead of having it on top of the interpreter call chain, since the only source of retryable errors is within builtin function calls anyways.

This also solves 
Jira: SC-1175


rust/src/nasl/interpreter/forking_interpreter.rs
```diff
@@ -0,0 +1,357 @@
...
+                (self.interpreter_index + num_forks - 1) % self.interpreters.len();
+            true
+        } else {
+            false
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
nichtsfrei, line 103: would it make sense to write tests that triggers the forking error? That way we could work on reducing those cases more easily.

Tehforsch: There is already such a test: `forks_with_different_branches`.

Tehforsch: Am definitely in favor of adding more though.

PR #1860, opened 2025-03-05
Fix: get vts endpoint

nichtsfrei:
`GET /vts` is too slow for bombardiers default timeout.

    This fixes the first byte timeout issue and can be tested via:
    `bombardier localhost:3000/vts` which shou just return 2xx and not crash
    openvasd.

Depends on: https://github.com/greenbone/openvas-scanner/pull/1813

SC-1219


rust/src/openvasd/controller/entry.rs
```diff
@@ -447,14 +447,9 @@ where
...
                         },
-                        None if meta => Ok(ctx
-                            .response
-                            .ok_json_stream(ctx.scheduler.vts().await?)
-                            .await),
-                        None => Ok(ctx
-                            .response
-                            .ok_json_stream(ctx.scheduler.oids().await?)
-                            .await),
+                        // TODO: try to return slices instead of Vec
```
Tehforsch, line 450: Can we either remove or do the TODO? Otherwise this will be here for eternity.

I might be misreading the intention, but I don't think changing the return type of `.vts()` can be done, since, for example, the in memory storage will clone the retrieved `Vec` so you won't be able to return a slice.

rust/src/openvasd/storage/redis.rs
```diff
@@ -271,7 +271,8 @@ where
         let noids = noids.await.unwrap()?;
         let foids = foids.await.unwrap()?;
         let results = noids.chain(foids);
-        Ok(Box::new(results))
+        // FIXME: moep that is probably slower
```
Tehforsch, line 274: fixme :D

rust/src/storage/inmemory/mod.rs
```diff
@@ -48,47 +58,69 @@ impl InMemoryStorage {
...
+        oid_lookup.insert(
+            Self::to_notus_advisory_key(&item.oid),
+            item.filename.clone(),
+        );
+        drop(oid_lookup);
+        let mut vts = self.vts.write()?;
+        vts.insert(Self::to_notus_advisory_key(&item.filename), item);
+
+        // let mut data = self.advisories.write()?;
+        // data.push(adv);
```
Tehforsch, line 100: remove?

rust/src/storage/inmemory/mod.rs
```diff
@@ -48,47 +58,69 @@ impl InMemoryStorage {
...
-        data.push(adv);
+        let item: Nvt = adv.into();
+
+        // TODO: here we can warn about duplicates
+        let mut oid_lookup = self.oid_lookup.write()?;
+        oid_lookup.insert(
+            Self::to_notus_advisory_key(&item.oid),
+            item.filename.clone(),
+        );
+        drop(oid_lookup);
```
Tehforsch, line 95: Why the drop here? `oid_lookup` and `vts` are different locks, so they dont contend

rust/src/storage/inmemory/mod.rs
```diff
@@ -48,47 +58,69 @@ impl InMemoryStorage {
...
+        // data.push(adv);
         Ok(())
     }
 
     fn all_vts(&self) -> Result<Vec<Nvt>, StorageError> {
         let vts = self.vts.read()?;
         let vts = vts.values().cloned();
-        let notus = self.advisories.read()?.clone().into_iter().map(Nvt::from);
-        Ok(vts.chain(notus).collect())
+        // let notus = self.advisories.read()?.clone().into_iter().map(Nvt::from);
```
Tehforsch, line 107: remove?

rust/src/storage/inmemory/mod.rs
```diff
@@ -48,47 +58,69 @@ impl InMemoryStorage {
...
+        Ok(vts.collect())
```
Tehforsch, line 108: Due to the smarter drop rules in 2024 you we now get to write

`Ok(self.vts.read()?.values().cloned().collect())`

rust/src/storage/inmemory/mod.rs
```diff
@@ -170,21 +202,50 @@ impl Retriever<FileName> for InMemoryStorage {
     type Item = Nvt;
     fn retrieve(&self, key: &FileName) -> Result<Option<Self::Item>, StorageError> {
         let vts = self.vts.read()?;
-        Ok(vts.get(&key.0).cloned())
+        // Duplicate Notus Nasl prevention, when notus available return that otherwise NASL
```
Tehforsch, line 205: I could not figure out what this comment was trying to tell me without reading the code. Could you please reword it?

rust/src/storage/inmemory/mod.rs
```diff
@@ -170,21 +202,50 @@ impl Retriever<FileName> for InMemoryStorage {
...
 impl Retriever<Oid> for InMemoryStorage {
     type Item = Nvt;
     fn retrieve(&self, key: &Oid) -> Result<Option<Self::Item>, StorageError> {
-        let vts = self.vts.read()?;
         let oid_lookup = self.oid_lookup.read()?;
-        Ok(oid_lookup
-            .get(&key.0)
-            .and_then(|filename| vts.get(filename).cloned()))
+        // is it really better to use the filename as a key identifier?
+        // I think in the most cases we would lookup oids?
```
Tehforsch, line 221: That means changing the `oid_lookup` to sth like `filename_lookup`? Agree that that sounds like a good idea if we do look up OIDs more often

rust/src/storage/inmemory/mod.rs
```diff
@@ -48,47 +58,69 @@ impl InMemoryStorage {
...
 
     /// Stores an already existing Vts structure.
-    pub fn set_vts(&self, vts: Vts) -> Result<(), StorageError> {
+    pub fn set_vts(&self, vts: HashMap<String, Nvt>) -> Result<(), StorageError> {
         let mut vts_internal = self.vts.write()?;
         let mut oid_lookup = self.oid_lookup.write()?;
 
         for (filename, nvt) in &vts {
-            oid_lookup.insert(nvt.oid.clone(), filename.clone());
+            oid_lookup.insert(Self::to_nasl_key(&nvt.oid), filename.clone());
```
Tehforsch, line 71: If I understand correctly, we currently never clear the `oid_lookup` map and it only ever grows. Is this intentional?

Tehforsch (changes requested):

rust/src/openvasd/controller/entry.rs
nichtsfrei: right, that was an remand when I was considering Cow.

rust/src/storage/inmemory/mod.rs
nichtsfrei: Without the drop it stops working on aarch64, I'm not sure exactly why.

Tehforsch (approved):

PR #1864, opened 2025-03-14
Fix >>>= operator

Tehforsch:


nichtsfrei (approved):

rust/src/nasl/interpreter/assign.rs
```diff
@@ -223,7 +223,7 @@ impl Interpreter<'_, '_> {
                     // get rid of minus sign
                     let left = i64::from(left) as u32;
                     let right = i64::from(right) as u32;
-                    NaslValue::Number((left << right) as i64)
+                    NaslValue::Number((left >> right) as i64)
```
Kraemii, line 226: Nice catch :)
I think it would also better to use u64 instead

Tehforsch: While I agree that using `u64` seems more reasonable, `u32` is closer to the C implementation since that uses `unsigned`. I propose leaving it as it is.

Kraemii (approved):

PR #1867, opened 2025-03-17
Fix scanner preferences 

jjnicola:
**What**:
This PR fixes some scanner preferences related issues which produces a difference between results when openvas is directly run and via ospd

Jira: SC-1255

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/src/storage/items/nvt.rs
+++ b/rust/src/storage/items/nvt.rs
@@ -198,15 +198,62 @@ make_str_lookup_enum! {
     }
 }
 
-make_str_lookup_enum! {
-    PreferenceType: "Allowed types for preferences" => {
-       checkbox => CheckBox,
-       entry => Entry,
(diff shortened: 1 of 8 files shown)


rust/src/storage/items/nvt.rs
```diff
@@ -198,15 +198,63 @@ make_str_lookup_enum! {
...
-make_str_lookup_enum! {
-    PreferenceType: "Allowed types for preferences" => {
-       checkbox => CheckBox,
-       entry => Entry,
-       file => File,
-       password => Password,
-       radio => Radio,
-       sshlogin => SshLogin,
-       integer => Integer
+#[doc = "Allowed types for preferences"]
```
nichtsfrei, line 201: Can we please stick to `///`?

PR #1871, opened 2025-03-25
Move Sockets to `Context`.

Tehforsch:
This moves the state that was previously contained in the `NaslSockets` struct (the state struct on which the socket-related builtins are implemented) to the `Context` so we have access to it in other places as well. We will eventually need to use custom sockets in the HTTP and SSH builtins (among other things), so having a central managing instance is necessary.

In order to make the access to the sockets reasonably convenient (and enabling future changes to how they are accessed from the `Context`, I also added the ability to take the `NaslSockets` struct as an argument (either via ref or mut ref) to a builtin function.)

The changes in this PR probably look a lot scarier than they are, since, in order to use the new structure in the socket builtins, I had to move all of the builtins out of the `NaslSockets` into top level functions that do the exact same thing but take `&NaslSockets` instead of `&self`.

Jira: SC-1210

### DIFF
--- a/rust/src/nasl/builtin/network/socket.rs
+++ b/rust/src/nasl/builtin/network/socket.rs
@@ -11,10 +11,7 @@ use std::{
 };
 
 use crate::nasl::{prelude::*, utils::function::Seconds};
-use crate::{
-    nasl::prelude::*,
-    storage::items::kb::{self, KbKey},
-};
+use crate::storage::items::kb::{self, KbKey};
 use dns_lookup::lookup_host;
 use lazy_regex::{Lazy, lazy_regex};
 use regex::Regex;
--- a/rust/src/nasl/builtin/network/socket.rs
+++ b/rust/src/nasl/builtin/network/socket.rs
@@ -217,7 +217,7 @@ impl NaslSockets {
 
 /// Close a given file descriptor taken as an unnamed argument.
 #[nasl_function]
-fn close(sockets: &mut NaslSockets, socket_fd: usize) -> Result<(), FnError> {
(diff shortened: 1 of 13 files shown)


rust/src/nasl/utils/context.rs
```diff
@@ -714,6 +739,16 @@ impl<'a> Context<'a> {
         };
         Ok(ret)
     }
+
+    pub async fn read_sockets(&self) -> tokio::sync::RwLockReadGuard<'_, NaslSockets> {
+        // TODO do not unwrap?
```
nichtsfrei, line 744: I don't see a unwrap?

rust/src/nasl/builtin/network/socket.rs
```diff
@@ -561,167 +227,497 @@ impl NaslSockets {
...
+#[nasl_function(named(socket, data, option, len))]
+async fn send(
+    sockets: &mut NaslSockets,
+    socket: usize,
+    data: &[u8],
+    option: Option<i64>,
+    len: Option<usize>,
+) -> Result<usize, SocketError> {
+    let len = if let Some(len) = len {
+        if len < 1 || len > data.len() {
```
nichtsfrei, line 263: Isn't that a usage error? At least a debug message would be nice?

Tehforsch: None of this code is new or really part of this PR, it just had to be moved out of `NaslSockets`

rust/src/nasl/utils/context.rs
Tehforsch: Oh, oops, this was from back when the lock wasn't a tokio lock. My bad, got rid of the comment.

Kraemii (approved):

PR #1873, opened 2025-03-26
Notus regex

Kraemii:
Reformatting the Regular Expressions for notus to make them more readable.


Tehforsch (approved):
Looks a lot better. Also thanks for the pointer to the specification

PR #1874, opened 2025-03-26
Add: Support for Arch Linux / ALPM in Notus

Kraemii:
SC-1240

### DIFF
--- a/rust/src/notus/packages/alpm.rs
+++ b/rust/src/notus/packages/alpm.rs
@@ -1,0 +1,302 @@
+// SPDX-FileCopyrightText: 2025 Greenbone AG
+//
+// SPDX-License-Identifier: GPL-2.0-or-later WITH x11vnc-openssl-exception
+
+use std::cmp::Ordering;
+
+use lazy_regex::{Lazy, lazy_regex};
+use regex::Regex;
+
+use super::{Package, PackageVersion};
+
+static RE: Lazy<Regex> = lazy_regex!(
+    r"^(?x)
+    (?:
+        (?P<epoch>
+            \d+
+        )
+        :
+    )?
+    (?P<pkgver>
+        [^:\/\s-]+
+    )
+    (?:
+        -
+        (?P<pkgrel>
+            \d+
+            (?:
+                .\d+
+            )?
+        )
+    )?
+    $"
+);
+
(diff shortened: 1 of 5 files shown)


rust/src/notus/packages/alpm.rs
```diff
@@ -0,0 +1,302 @@
...
+            };
+        };
+
+        if let Some(comp) = self.pkgver.partial_cmp(&other.pkgver) {
+            if comp.is_ne() {
+                return Some(comp);
+            }
+        }
+
+        self.pkgrel.partial_cmp(&other.pkgrel)
```
Tehforsch, line 63: Nit:

Could be written as 
```rust
        self.epoch
            .partial_cmp(&other.epoch)
            .filter(|comp| comp.is_ne())
            .or(self
                .pkgver
                .partial_cmp(&other.pkgver)
                .filter(|comp| comp.is_ne()))
            .or(self.pkgrel.partial_cmp(&other.pkgrel))
```

rust/src/notus/packages/alpm.rs
```diff
@@ -0,0 +1,302 @@
...
+            .map(|m| m.as_str().to_string())
+            .map(PackageVersion)
+            // Cannot fail, because the regex ensures that there is a pkgver
+            .unwrap();
+
+        let pkgrel = captures
+            .name("pkgrel")
+            .map(|m| m.as_str().to_string())
+            .map(PackageVersion)
+            .unwrap_or_else(|| PackageVersion("1".to_string()));
```
Tehforsch, line 101: hyper-nit (does not need to be done in this pr at all): if `PackageVersion` had a `From<&str>`, the conversions could be a lot cleaner via

```rust
        let pkgrel = captures
            .name("pkgrel")
            .map(|m| m.as_str())
            .unwrap_or("1")
            .into();
```

or even

```rust
        let pkgrel = captures
            .name("pkgrel")
            .map_or("1", |m| m.as_str())
            .into();
```

Tehforsch (dismissed):
Again, #1872 and #1875 should fix the CI, maybe wait for them

Kraemii:
Thanks for the review! I will implement the suggestions in an additional PR, as it is also applicable to the other packages.

Tehforsch (approved):

PR #1879, opened 2025-04-01
Notus improvements

Kraemii:
Tidy up notus package modules and fix ebuild parsing and comparison:
    
Many things were not considered regarding ebuild files.
- The path to the file could potentially contain more than one '/'
- The suffixes must be compaired iteratively and follow different rules than 'normal' versions

SC-1280


rust/src/notus/packages/ebuild.rs
```diff
@@ -2,15 +2,220 @@
...
+                        if ordering.is_ne() {
+                            return Some(ordering);
+                        }
+                    }
+                }
+            }
+        }
+        Some(Ordering::Equal)
+    }
+}
```
Tehforsch, line 208: Just because it's fun:

```rust
fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
    let len = self.0.len().max(other.0.len());
    let it1 = self.0.iter().chain(iter::repeat(&Suffix::Normal)).take(len);
    let it2 = other.0.iter().chain(iter::repeat(&Suffix::Normal)).take(len);
    it1.partial_cmp(it2)
}
```

rust/src/notus/packages/ebuild.rs
```diff
@@ -29,27 +241,35 @@ impl Package for EBuild {
...
+            })?;
 
         Some(EBuild {
+            path,
             name,
-            version: PackageVersion(full_version),
+            version,
+            suffix_part,
+            revision,
         })
```
Tehforsch, line 273: Nit:

```suggestion rust
        RE_FULL.captures(full_name).and_then(|caps| {
            let path = caps.name("path").map(|m| m.as_str()).unwrap_or_default();
            let name = caps.name("name")?.as_str();
            let version = caps.name("version")?.as_str();
            let suffix = caps.name("suffix")?.as_str();
            let revision = caps
                .name("revision")
                .map(|r| r.as_str().parse::<u64>().unwrap())
                .unwrap_or_default();

            Some(EBuild {
                path: path.into(),
                name: name.into(),
                version: version.into(),
                suffix: suffix.into(),
                revision,
            })
        })
    }
```

rust/src/notus/packages/ebuild.rs
```diff
@@ -112,15 +387,36 @@ mod ebuild_tests {
...
 
         let apache3 =
             EBuild::from_name_and_full_version("www-servers/apache", "2.4.51-r3").unwrap();
 
-        assert!(apache2 <= apache3);
-        assert!(apache2 >= apache3);
+        assert!(apache2 == apache3);
 
-        let apache4 = EBuild::from_name_and_full_version("apache", "2.4.51-r3").unwrap();
+        let apache4 = EBuild::from_name_and_full_version("apachee", "2.4.51-r3").unwrap();
```
Tehforsch, line 396: Is this typo intentional?

Tehforsch (commented):
Nice, thank you!

A few general comments that came to mind while reviewing:

The types that implement `Package`, such as `EBuild` contain more information than just the version information (`name` and `path` for example).
The `PartialOrd` implementation for those types only compares the version fields. This intuitively makes sense, but will lead to a situation where the `PartialEq` and `PartialOrd` traits are incompatible. 
For example on https://doc.rust-lang.org/std/cmp/trait.PartialOrd.html#examples-of-incorrect-partialord-implementations, the first condition is
`a == b` iff `partial_cmp(a, b) == Some(Equal)`.
This won't be the case for packages with different names but equal versions/revisions.
The standard library strongly discourages `PartialOrd` impls like this, since it can lead to inconsistent and unexpected behavior of the `==` and `<` operators. In the current code this isn't a problem at all, so I'd be fine with merging it as it is, but if we want to do this without violating the conditions, we could change the `Package` trait to something like

```rust
pub trait Package<Rhs = Self> {
    fn version(&self) -> impl PartialOrd;
    ...
}
```

with an example impl like:

```rust
impl Package for Alpm {
    fn version(&self) -> impl PartialOrd {
        (self.epoch, &self.pkgver, &self.pkgrel)
    }

    ...
}
```

and then replacing the instances of `<=` and `>=` directly on packages with `p1.version() <= p2.version()` of course.
I personally find this a bit easier to read in any case.

Technically all of our types also form a total order, so we could also implement `Ord` (and derive `PartialOrd` from it) but that feels more like a mathematical excercise than one that is actually relevant to the code.

rust/src/notus/packages/deb.rs
```diff
@@ -103,20 +108,14 @@ impl PartialOrd for Deb {
...
-
-        self.debian_revision.partial_cmp(&other.debian_revision)
+        self.epoch
+            .partial_cmp(&other.epoch)
+            .filter(|comp| comp.is_ne())
+            .or(self
+                .upstream_version
+                .partial_cmp(&other.upstream_version)
+                .filter(|comp| comp.is_ne()))
+            .or(self.debian_revision.partial_cmp(&other.debian_revision))
```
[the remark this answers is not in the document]
Kraemii: Never thought about this. This is really nice :D
The only note: I need references everywhere in this case.

rust/src/notus/packages/ebuild.rs
Kraemii: Yes, it tests, for two ebuild with different names. They should be not Comparable

rust/src/notus/packages/deb.rs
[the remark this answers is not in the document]
Tehforsch: (You can technically leave out the reference for the epoch, but whatever :D)

Tehforsch (approved):
Thanks!

PR #1880, opened 2025-04-02
Add: implement in rust pem_to_rsa() and pem_to_dsa() nasl builtin functions

jjnicola:
**What**:
Add: implement in rust pem_to_rsa() and pem_to_dsa() nasl builtin functions
Also, add nasl example and update Readme file
Jira:: SC-1286

**Why**:

**How**:

**Checklist**:

- [ ] Tests


rust/src/nasl/builtin/cryptographic/pem_to.rs
```diff
@@ -0,0 +1,46 @@
...
+            Ok(x.iter().map(|x| *x as char).collect::<String>())
+        }
+        _ => Err(CryptographicError::Rsa("invalid key".to_string())),
+    }
+}
+
+#[nasl_function(named(passphrase))]
+fn pem_to_rsa(register: &Register, passphrase: String) -> Result<NaslValue, CryptographicError> {
+    let ori_pem = get_priv(register)?;
+    let decrypted_key = RsaPrivateKey::from_pkcs8_encrypted_pem(&ori_pem, passphrase).unwrap();
```
Tehforsch, line 27: We should handle the result properly here, right? Or is there any reason we can expect to be able to `.unwrap` the encryption result?

Tehforsch (changes requested):

rust/src/nasl/builtin/cryptographic/pem_to.rs
jjnicola: done with [56c427c](https://github.com/greenbone/openvas-scanner/pull/1880/commits/56c427cdc1f521e8c6dc22f2004da0bb3eb5e1be)

Tehforsch (approved):
Thanks!

PR #1883, opened 2025-04-08
Add: nasl builtin function script_get_preference()

jjnicola:
**What**:
Add: nasl builtin function script_get_preference()
SC-1102

**Why**:

**How**:

**Checklist**:

- [ ] Tests


rust/src/scanner/vt_runner.rs
```diff
@@ -66,20 +67,19 @@ where
...
-
     fn set_parameters(&mut self, register: &mut Register) -> Result<(), ExecuteError> {
         if let Some(params) = &self.param {
+            let mut parameters = Vec::new();
+            for _ in params.iter() {
+                parameters.push("".into());
+            }
             for p in params.iter() {
-                self.parameter(p, register)?;
+                parameters.insert(p.id.into(), NaslValue::String(p.value.clone()));
```
Tehforsch, line 77: This seems very confusing to me. If I understand correctly, the `id` of the parameter is obtained from deserialization? So it could really be any integer and there is no reason it has to be in the range `0..params.len()`, meaning that this insertion could fail.
Even if we do have a reason to believe that the id is in the correct range (and that there is exactly one id for every integer in `0..params.len()`) then I think this should be checked (by checking for index out of bounds before the insertion) or at the very least be written down in a comment.

If we don't have reason to believe that the IDs are that well behaved, then we need to do this differently.
Unfortunately the available data structures aren't great - we don't have tuples, so we'd have to do a two dimensional array, or alternatively use `NaslValue::Dict`, with the stringified id as a key (which also seems ugly).

Tehforsch (changes requested):

rust/src/nasl/utils/context.rs
```diff
@@ -252,6 +255,22 @@ impl Register {
         }
     }
 
+    /// Retrieves all script parameters
+    pub fn script_params(&self) -> &[NaslValue] {
+        match self.named(SCRIPT_PARAMS) {
+            Some(ContextType::Value(NaslValue::Array(arr))) => arr,
+            _ => &[],
+        }
+    }
```
Tehforsch, line 264: I didn't think of this before, but this is not going to work with the current structure, since this will always return the empty array.

I wonder if storing these variables in the register is even the right approach. Adding a hash map with the appropriate types to context could really simplify the code and allow the type checker to do its job. Iterating over all script parameters would then become a simple `.iter()` call while its impossible in the current structure.

Tehforsch (changes requested):

Tehforsch (approved):
Thanks!

PR #1884, opened 2025-04-09
Use target in builtins

Tehforsch:
Use the `Target` field of `Context` in all builtins to ensure consistency and to actually use the hostname resolution.

Jira: SC-1289

### DIFF
--- a/rust/src/nasl/builtin/report_functions/mod.rs
+++ b/rust/src/nasl/builtin/report_functions/mod.rs
@@ -49,7 +49,7 @@ impl Reporting {
             Some("udp") => Protocol::UDP,
             _ => Protocol::TCP,
         };
-        let target = context.target_orig();
+        let target = context.target();
         let hostname = target.hostname();
         let ip_address = target.ip_addr();
         let result = models::Result {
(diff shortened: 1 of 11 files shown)


rust/src/nasl/builtin/report_functions/mod.rs
```diff
@@ -49,12 +49,14 @@ impl Reporting {
...
+        let hostname = target.hostname();
+        let ip_address = target.ip_addr();
         let result = models::Result {
             id: self.id(),
             r_type: typus,
-            ip_address: Some(context.target().to_string()),
-            // TODO: where to get hostname? is it only vhost relevant?
-            hostname: None,
+            ip_address: Some(ip_address.to_string()),
+            hostname,
```
Tehforsch, line 59: This is a outward-facing change and we need to make sure its ok. This new behavior returns the hostname if the given target was a hostname or `None` if the given target was an ip address, which sounds like sane behavior to me, but I am not totally sure.

jjnicola (changes requested):

rust/src/nasl/builtin/report_functions/mod.rs
Tehforsch: Valid points, but do you think that the new implementation is an improvement over the status quo (which never sets a hostname, even if one was given in the original target)? If so, I'd create an issue for your suggestions but do them at a later point.

jjnicola (approved):

PR #1885, opened 2025-04-09
Fix: nvt xrefs param name and value exchanged

jjnicola:
**What**:
Fix: nvt xrefs param name and value exchanged
SC-1287

**Why**:

**How**:

**Checklist**:

- [ ] Tests


Tehforsch (approved):
thanks!

PR #1886, opened 2025-04-10
Fix crash during feed transform .

Tehforsch:
After #1813, the feed transform into json would crash due to a `unimplemented!()` in the dispatching of `FeedVersion` to the storage.

This reverts back to the state before #1813, where the feed version information would simply not be exported to json.

Jira: SC-1287

### DIFF
--- a/rust/src/storage/infisto/json/mod.rs
+++ b/rust/src/storage/infisto/json/mod.rs
@@ -135,7 +135,8 @@ impl<S: Write> Dispatcher<FileName> for JsonStorage<S> {
 impl<S: Write> Dispatcher<FeedVersion> for JsonStorage<S> {
     type Item = String;
     fn dispatch(&self, _: FeedVersion, _: Self::Item) -> Result<(), StorageError> {
-        unimplemented!()
+        // TODO: Feed information is currently not written to the output json
+        Ok(())
     }
 }
 


jjnicola (approved):

rust/src/storage/infisto/json/mod.rs
```diff
@@ -135,7 +135,8 @@ impl<S: Write> Dispatcher<FileName> for JsonStorage<S> {
 impl<S: Write> Dispatcher<FeedVersion> for JsonStorage<S> {
     type Item = String;
     fn dispatch(&self, _: FeedVersion, _: Self::Item) -> Result<(), StorageError> {
-        unimplemented!()
+        // Feed information is currently not written to the output json
+        Ok(())
```
jjnicola, line 139: Is this the final implementation or is it a workaround ? I mean. Will the feed version never be dispached for JsonStorage? Otherwise, I would add at least a TODO comment, to not forget to implement this treat for JsonStorage.

jjnicola (dismissed):
I mark this as approved. I think we currently do not need to store the version into the json file. Please consider my suggestion, if the Dispacher<FeedVersion> trait implementation for JsonStorage still need to be implemented for other cases.

rust/src/storage/infisto/json/mod.rs
Tehforsch: I don't really know, I think that's something we still have to decide. I will open a ticket for it.

Tehforsch:
Opened SC-1297 for the issue of whether to implement json storage for the feed information which I think should be decided separately.

jjnicola (dismissed):

jjnicola (approved):

PR #1892, opened 2025-04-22
Fix: Improve config parsing

Kraemii:
The parser know checks if there is a file in one of the default paths and only returns None if there isn't. In all other cases he tries to parse the config file and panics, if he fails to do so.

One additional thing to discuss here:
Should we ignore unknown fields in the config file or throw an error.
Pro of throwing an error would be, that typos are recognized.
Pro of ignoring them when we change the config structure, we don't throw an error, when openvasd is updated. So just updating without adjusting the config file could lead to users having errors.

### DIFF
--- a/rust/src/openvasd/config.rs
+++ b/rust/src/openvasd/config.rs
@@ -317,17 +317,32 @@ impl Display for Config {
 }
 
 impl Config {
+    fn read_to_string<P>(path: P) -> String
+    where
+        P: AsRef<std::path::Path> + std::fmt::Display + std::fmt::Debug,
+    {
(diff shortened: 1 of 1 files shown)


jjnicola (approved):

jjnicola:
> One additional thing to discuss here: Should we ignore unknown fields in the config file or throw an error. Pro of throwing an error would be, that typos are recognized. Pro of ignoring them when we change the config structure, we don't throw an error, when openvasd is updated. So just updating without adjusting the config file could lead to users having errors.

I think that a warning message would be an option for unknown fields.

PR #1893, opened 2025-04-23
Add: store ports of scan config in context

Kraemii:
This makes the access of the requested ports available within the Context of NASL.

SC-1101


rust/src/scannerctl/execute/mod.rs
```diff
@@ -41,6 +41,12 @@ struct ScriptArgs {
     /// Target to scan.
     #[clap(short, long)]
     target: Option<String>,
+    /// TCP Ports to scan.
+    #[clap(short, long)]
+    port: Vec<u16>,
+    /// UDP Ports to scan.
+    #[clap(short, long)]
+    udp_port: Vec<u16>,
```
Tehforsch, line 49: Shouldn't this be plural?

rust/src/nasl/utils/context.rs
```diff
@@ -406,6 +406,36 @@ pub struct Target {
     kind: TargetKind,
 }
 
+#[derive(Clone, Debug, Default)]
+pub struct Ports {
+    /// The TCP ports to test against.
+    pub tcp: HashSet<u16>,
+    /// The UDP ports to test against.
+    pub udp: HashSet<u16>,
```
Tehforsch, line 414: I assume the `HashSet` is used because eventually, we'll want to quickly check if a port is within the given range(s), right? Have you considered using a type like `Vec<Range<u16>>` instead? That still allows really quick member checks (if the `Vec` is small, could even be faster than `HashSet`) and it would use a lot less memory.

rust/src/nasl/test_utils.rs
```diff
@@ -124,6 +125,8 @@ pub struct TestBuilder<L: Loader, S: ContextStorage> {
     scan_id: ScanID,
     filename: PathBuf,
     target: String,
+    ports_tcp: HashSet<u16>,
+    ports_udp: HashSet<u16>,
```
Tehforsch, line 129: This is currently unused and the ports will always be empty, right? Since there is no setter for these fields right now, I'd prefer getting rid of them and simply initializing with empty ports instead, unless you expect to write a test that utilizes and tests against these port ranges (I'm not sure how that would work).

rust/src/nasl/utils/context.rs
```diff
@@ -406,6 +406,36 @@ pub struct Target {
     kind: TargetKind,
 }
 
+#[derive(Clone, Debug, Default)]
```
Tehforsch, line 409: I would prefer if the `Default` derive were removed and replaced with a `empty()` or `none()` constructor on the `Ports` type. Possibly even with a `#[cfg(test)]` gate. I think the happy path of the code should have a non-empty set of ports to scan and an empty port range is really only useful in tests (unless I am misunderstanding something).

rust/src/nasl/utils/context.rs
Kraemii: There are several applications, that ports are used to. I tried to use the best Data type for all worlds, as we do not know, in which format the Ports are given (in Ranges or only Single ports). Yes you are right, when we get a few large port ranges, but if something like the list of iana assigned tcp and udp (https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers) is used, we have thousands of small ranges, many only containing only a single Port. So with a `HashSet` it is easy to look up, pick a Random one, iterate through etc.

rust/src/scannerctl/execute/mod.rs
Kraemii: The reason I did it in singular is for better understanding for the command line, but I changed it, so it is now plural within the code, but the command for adding a port is in singular.

rust/src/nasl/utils/context.rs
Kraemii: Not sure, what exactly you mean here. It is correct, that we do not allow empty port list for scans, but there are many other cases, we do allow them for example `scannerctl` or for testing. So in general it is not possible to force a non empty port list.

Tehforsch: My point is that the existence of a `Default` impl implies that it is totally fine to construct the struct by using this impl. Here, the `Default` is more of an edge case rather than the default, so I'd prefer if we had an explicit `empty()` constructor instead. Not super important though, if you prefer the `Default`, thats fine

Tehforsch (approved):

PR #1894, opened 2025-04-24
Add: Handle scan preferences

jjnicola:
**What**:
Loads default preferences and replace them with the ones sent by the client in the scan config. Also, make the preferences available to the openvasd-scanner.
TODO: load config (defaults and from config file) for scannerctl.

Jira: SC-1103

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/src/models/scanner_preference.rs
+++ b/rust/src/models/scanner_preference.rs
@@ -34,6 +34,27 @@ impl Default for PreferenceValue {
     }
 }
 
+impl PreferenceValue {
+    pub fn get_bool(&self) -> bool {
+        match self {
+            PreferenceValue::Bool(v) => *v,
+            _ => false,
+        }
+    }
+    pub fn get_string(&self) -> String {
+        match self {
+            PreferenceValue::String(v) => v.to_string(),
+            _ => "".to_string(),
+        }
+    }
+    pub fn get_int(&self) -> i64 {
+        match self {
+            PreferenceValue::Int(v) => *v,
+            _ => 0,
+        }
+    }
+}
+
 /// Configuration preference information for a scan. The type can be derived from the default value.
 #[derive(Default, Debug, Clone, PartialEq, Eq)]
 #[cfg_attr(
--- a/rust/src/models/scanner_preference.rs
+++ b/rust/src/models/scanner_preference.rs
@@ -34,29 +34,8 @@ impl Default for PreferenceValue {
     }
 }
 
-impl PreferenceValue {
-    pub fn get_bool(&self) -> bool {
-        match self {
-            PreferenceValue::Bool(v) => *v,
-            _ => false,
-        }
-    }
-    pub fn get_string(&self) -> String {
-        match self {
-            PreferenceValue::String(v) => v.to_string(),
-            _ => "".to_string(),
-        }
-    }
-    pub fn get_int(&self) -> i64 {
-        match self {
-            PreferenceValue::Int(v) => *v,
-            _ => 0,
-        }
-    }
-}
-
 /// Configuration preference information for a scan. The type can be derived from the default value.
-#[derive(Default, Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 #[cfg_attr(
     feature = "serde_support",
     derive(serde::Serialize, serde::Deserialize)
--- a/rust/src/openvasd/config.rs
+++ b/rust/src/openvasd/config.rs
@@ -576,7 +576,32 @@ impl Config {
             );
 
         for pref in PREFERENCES {
-            cmds = cmds.arg(clap::Arg::new(pref.id).long(pref.id).help(pref.description));
+            match pref.default {
+                PreferenceValue::Bool(_) => {
+                    cmds = cmds.arg(
+                        clap::Arg::new(pref.id)
+                            .long(pref.id)
+                            .value_parser(clap::builder::BoolValueParser::new())
+                            .help(pref.description),
+                    );
+                }
+                PreferenceValue::String(_) => {
+                    cmds = cmds.arg(
+                        clap::Arg::new(pref.id)
+                            .long(pref.id)
+                            .value_parser(clap::value_parser!(String))
+                            .help(pref.description),
+                    );
+                }
+                PreferenceValue::Int(_) => {
+                    cmds = cmds.arg(
+                        clap::Arg::new(pref.id)
+                            .long(pref.id)
+                            .value_parser(clap::value_parser!(i64))
+                            .help(pref.description),
+                    );
+                }
+            };
         }
 
         let cmds = cmds.get_matches();
--- a/rust/src/openvasd/config.rs
+++ b/rust/src/openvasd/config.rs
@@ -362,6 +362,19 @@ impl Config {
         toml::from_str(&config).unwrap()
     }
 
+    fn ret_pref_value_or_panic_with_wrong_type(
+        p: &ScanPrefValue,
+        expected: &str,
+        name: &str,
+    ) -> ScanPrefValue {
+        match (p, expected) {
+            (ScanPrefValue::Bool(_), "bool") => p.clone(),
+            (ScanPrefValue::Int(_), "int") => p.clone(),
+            (ScanPrefValue::String(_), "string") => p.clone(),
+            _ => panic!("Wrong value type. Expected {} for {}", expected, name),
+        }
+    }
+
     pub fn load() -> Self {
         let mut cmds = clap::Command::new("openvasd")
             .arg(
@@ -680,10 +693,8 @@ impl Config {
                     PreferenceValue::Bool(_) => {
                         if let Some(p) = cmds.get_one::<bool>(pref.id) {
                             ScanPrefValue::Bool(*p)
-                        } else if let Some(ScanPrefValue::Bool(p)) =
-                            config.scanner.preferences.get(pref.id)
-                        {
-                            ScanPrefValue::Bool(*p)
+                        } else if let Some(p) = config.scanner.preferences.get(pref.id) {
+                            Self::ret_pref_value_or_panic_with_wrong_type(p, "bool", pref.id)
                         } else {
                             pref.default.clone().into()
                         }
@@ -691,10 +702,8 @@ impl Config {
                     PreferenceValue::String(_) => {
                         if let Some(p) = cmds.get_one::<String>(pref.id) {
                             ScanPrefValue::String(p.to_string())
-                        } else if let Some(ScanPrefValue::String(p)) =
-                            config.scanner.preferences.get(pref.id)
-                        {
-                            ScanPrefValue::String(p.to_string())
+                        } else if let Some(p) = config.scanner.preferences.get(pref.id) {
+                            Self::ret_pref_value_or_panic_with_wrong_type(p, "string", pref.id)
                         } else {
                             pref.default.clone().into()
                         }
@@ -702,10 +711,8 @@ impl Config {
                     PreferenceValue::Int(_) => {
                         if let Some(p) = cmds.get_one::<i64>(pref.id) {
                             ScanPrefValue::Int(*p)
-                        } else if let Some(ScanPrefValue::Int(p)) =
-                            config.scanner.preferences.get(pref.id)
-                        {
-                            ScanPrefValue::Int(*p)
+                        } else if let Some(p) = config.scanner.preferences.get(pref.id) {
+                            Self::ret_pref_value_or_panic_with_wrong_type(p, "int", pref.id)
                         } else {
                             pref.default.clone().into()
                         }
--- a/rust/src/openvasd/config.rs
+++ b/rust/src/openvasd/config.rs
@@ -576,32 +576,13 @@ impl Config {
             );
 
         for pref in PREFERENCES {
-            match pref.default {
-                PreferenceValue::Bool(_) => {
-                    cmds = cmds.arg(
-                        clap::Arg::new(pref.id)
-                            .long(pref.id)
-                            .value_parser(clap::builder::BoolValueParser::new())
-                            .help(pref.description),
-                    );
-                }
-                PreferenceValue::String(_) => {
-                    cmds = cmds.arg(
-                        clap::Arg::new(pref.id)
-                            .long(pref.id)
-                            .value_parser(clap::value_parser!(String))
-                            .help(pref.description),
-                    );
-                }
-                PreferenceValue::Int(_) => {
-                    cmds = cmds.arg(
-                        clap::Arg::new(pref.id)
-                            .long(pref.id)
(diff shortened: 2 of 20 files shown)


rust/src/models/scanner_preference.rs
```diff
@@ -34,6 +34,27 @@ impl Default for PreferenceValue {
...
+    pub fn get_int(&self) -> i64 {
+        match self {
+            PreferenceValue::Int(v) => *v,
+            _ => 0,
+        }
+    }
+}
+
 /// Configuration preference information for a scan. The type can be derived from the default value.
 #[derive(Default, Debug, Clone, PartialEq, Eq)]
```
Tehforsch, line 59: This isn't in this PR, but why the `Default` impl, at all? I don't think this struct or `PreferenceValue` should have a `Default`.

rust/src/models/scanner_preference.rs
```diff
@@ -34,6 +34,27 @@ impl Default for PreferenceValue {
...
+        }
+    }
+    pub fn get_int(&self) -> i64 {
+        match self {
+            PreferenceValue::Int(v) => *v,
+            _ => 0,
+        }
+    }
+}
+
```
Tehforsch, line 57: These are all unused, so I suggest removing and adding them if we ever need them.

rust/src/openvasd/config.rs
```diff
@@ -568,8 +573,38 @@ impl Config {
...
+                PreferenceValue::Int(_) => {
+                    cmds = cmds.arg(
+                        clap::Arg::new(pref.id)
+                            .long(pref.id)
+                            .value_parser(clap::value_parser!(i64))
+                            .help(pref.description),
+                    );
+                }
+            };
+        }
```
Tehforsch, line 605: neat :)

Could make this shorter by removing the boiler plate:

```rust
        for pref in PREFERENCES {
            let mut arg = clap::Arg::new(pref.id).long(pref.id).help(pref.description);
            arg = match pref.default {
                PreferenceValue::Bool(_) => {
                    arg.value_parser(clap::builder::BoolValueParser::new())
                }
                PreferenceValue::String(_) => {
                    arg.value_parser(clap::value_parser!(String))
                }
                PreferenceValue::Int(_) => {
                    arg.value_parser(clap::value_parser!(i64))
                }
            };
            cmds = cmds.arg(arg);
        }
```

rust/src/openvasd/config.rs
```diff
@@ -655,19 +691,71 @@ impl Config {
...
+                    } else if let Some(ScanPrefValue::Int(p)) =
+                        config.scanner.preferences.get(pref.id)
+                    {
+                        scan_prefs.insert(pref.id.to_string(), ScanPrefValue::Int(*p));
+                    } else {
+                        scan_prefs.insert(pref.id.to_string(), pref.default.clone().into());
+                    }
+                }
+            }
+        }
```
Tehforsch, line 734: This can also be cleaned up by extracting the boilerplate:

```rust
        let scan_prefs: HashMap<String, ScanPrefValue> = PREFERENCES
            .iter()
            .map(|pref| {
                let val = match pref.default {
                    PreferenceValue::Bool(_) => {
                        if let Some(p) = cmds.get_one::<bool>(pref.id) {
                            ScanPrefValue::Bool(*p)
                        } else if let Some(ScanPrefValue::Bool(p)) =
                            config.scanner.preferences.get(pref.id)
                        {
                            ScanPrefValue::Bool(*p)
                        } else {
                            pref.default.clone().into()
                        }
                    }
                    PreferenceValue::String(_) => {
                        if let Some(p) = cmds.get_one::<String>(pref.id) {
                            ScanPrefValue::String(p.to_string())
                        } else if let Some(ScanPrefValue::String(p)) =
                            config.scanner.preferences.get(pref.id)
                        {
                            ScanPrefValue::String(p.to_string())
                        } else {
                            pref.default.clone().into()
                        }
                    }
                    PreferenceValue::Int(_) => {
                        if let Some(p) = cmds.get_one::<i64>(pref.id) {
                            ScanPrefValue::Int(*p)
                        } else if let Some(ScanPrefValue::Int(p)) =
                            config.scanner.preferences.get(pref.id)
                        {
                            ScanPrefValue::Int(*p)
                        } else {
                            pref.default.clone().into()
                        }
                    }
                };
                (pref.id.to_string(), val)
            })
            .collect();
```

rust/src/openvasd/config.rs
```diff
@@ -655,19 +691,71 @@ impl Config {
...
+            match pref.default {
+                PreferenceValue::Bool(_) => {
+                    if let Some(p) = cmds.get_one::<bool>(pref.id) {
+                        scan_prefs.insert(pref.id.to_string(), ScanPrefValue::Bool(*p));
+                    } else if let Some(ScanPrefValue::Bool(p)) =
+                        config.scanner.preferences.get(pref.id)
+                    {
+                        scan_prefs.insert(pref.id.to_string(), ScanPrefValue::Bool(*p));
+                    } else {
+                        scan_prefs.insert(pref.id.to_string(), pref.default.clone().into());
```
Tehforsch, line 706: If I understand this logic correctly, this means we silently fail when a wrong type is given and use the default instead. This seems like a UX nightmare. At the very least we should log an `error!`. Maybe even panic?

rust/src/openvasd/config.rs
```diff
@@ -655,19 +691,71 @@ impl Config {
...
+                        scan_prefs.insert(pref.id.to_string(), ScanPrefValue::Int(*p));
+                    } else {
+                        scan_prefs.insert(pref.id.to_string(), pref.default.clone().into());
+                    }
+                }
+            }
+        }
+
+        config.scanner.preferences = scan_prefs;
+        println!("Config: {config}");
```
Tehforsch, line 737: Remove this or replace with a `info!` call if we really want it.

rust/src/openvasd/config.rs
```diff
@@ -695,7 +783,7 @@ mod tests {
         assert_eq!(config.log.level, "INFO".to_string());
         // this is used to verify the default config manually.
         // se to true to write the default configuration to `tmp`
-        if false {
+        if true {
```
Tehforsch, line 786: This should be reverted to the previous state, but in any case I'm not a fan of this redundant if as a toggle. Maybe simply comment out the code if its really something thats needed.

rust/src/scanner/preferences/preference.rs
```diff
@@ -217,3 +219,121 @@ pub const PREFERENCES: [ScanPreferenceInformation; 22] = [
...
+    fn default() -> Self {
+        Self::Int(0)
+    }
+}
+
+#[derive(Default, Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(
+    feature = "serde_support",
+    derive(serde::Serialize, serde::Deserialize)
+)]
```
Tehforsch, line 251: I'd get rid of the `Clone` derive. This is potentially a very large struct and in the current code the only `.clone` calls are completely redundant.

rust/src/scanner/preferences/preference.rs
```diff
@@ -217,3 +219,121 @@ pub const PREFERENCES: [ScanPreferenceInformation; 22] = [
...
+        let mut scan_preferences = Vec::new();
+        for pref in PREFERENCES.iter() {
+            scan_preferences.push(FullScanPreference::from(pref));
+        }
+        Self { scan_preferences }
+    }
+
+    /// Override the default scan parameters with the one from the config file or command line.
+    pub fn load_params(&mut self, params: HashMap<String, ScanPrefValue>) -> &Self {
+        for (param_id, param_val) in params.clone().iter() {
```
Tehforsch, line 312: `params` is moved into this method, so you have ownership. No need to clone when you can just call `params.into_iter()` instead.

rust/src/scanner/preferences/preference.rs
```diff
@@ -217,3 +219,121 @@ pub const PREFERENCES: [ScanPreferenceInformation; 22] = [
...
+    pub fn new() -> Self {
+        let mut scan_preferences = Vec::new();
+        for pref in PREFERENCES.iter() {
+            scan_preferences.push(FullScanPreference::from(pref));
+        }
+        Self { scan_preferences }
+    }
+
+    /// Override the default scan parameters with the one from the config file or command line.
+    pub fn load_params(&mut self, params: HashMap<String, ScanPrefValue>) -> &Self {
```
Tehforsch, line 311: This method signature is a little strange, I suggest either not returning anything, or taking `self` instead of `&mut self` and then returning `Self` instead of `&Self`.

rust/src/openvasd/controller/context.rs
```diff
@@ -88,6 +93,14 @@ impl<S, DB, T> ContextBuilder<S, DB, T> {
         self
     }
 
+    /// Set scanner preferences
+    pub fn scan_preferences(mut self, prefs: HashMap<String, preference::ScanPrefValue>) -> Self {
+        let mut prefs_bind = preference::ScanPreferences::new();
+        let prefs = prefs_bind.load_params(prefs);
+        self.preferences = Some(prefs.clone());
+        self
```
Tehforsch, line 101: You don't have to clone here, since you can simply use `prefs_bind`:

```rust
        let mut prefs_bind = preference::ScanPreferences::new();
        prefs_bind.load_params(prefs);
        self.preferences = Some(prefs_bind);
        self
```

rust/src/openvasd/controller/entry.rs
```diff
@@ -454,9 +460,17 @@ where
                         Ok(ctx.response.not_found("scans", "all"))
                     }
                 }
-                (&Method::GET, ScanPreferences) => Ok(ctx
-                    .response
-                    .ok_static(crate::preference::PREFERENCES_JSON.as_bytes())),
+                (&Method::GET, ScanPreferences) => {
+                    if let Some(p) = ctx.scan_preferences.clone() {
```
Tehforsch, line 464: This clone is also redundant and can simply be removed if you take a reference to `ctx.scan_preferences`.

rust/src/openvasd/main.rs
```diff
@@ -211,6 +211,7 @@ async fn run(config: &Config) -> Result<()> {
 #[tokio::main]
 async fn main() -> Result<()> {
     let config = Config::load();
+    //    dbg!(&config);
```
Tehforsch, line 214: remove

rust/src/scanner/preferences/preference.rs
```diff
@@ -217,3 +219,121 @@ pub const PREFERENCES: [ScanPreferenceInformation; 22] = [
...
     pub static ref PREFERENCES_JSON: String = serde_json::to_string(&PREFERENCES).unwrap();
 }
+
+#[derive(Deserialize, Serialize, Debug, Clone, Eq, PartialEq)]
+#[serde(untagged)]
+pub enum ScanPrefValue {
+    Bool(bool),
+    Int(i64),
+    String(String),
+}
```
Tehforsch, line 229: Comment: A little unfortunate that this has to exist just because we decided to go with `&'static str` for `PreferenceValue`. Not really something that has to be changed (since `PREFERENCES` is a constant, it'd be harder than just changing `&'static str` to `String` in `PreferenceValue`).

Tehforsch (changes requested):
Ci currently fails

rust/src/scanner/preferences/preference.rs
```diff
@@ -217,3 +219,121 @@ pub const PREFERENCES: [ScanPreferenceInformation; 22] = [
...
+    pub fn new() -> Self {
+        let mut scan_preferences = Vec::new();
+        for pref in PREFERENCES.iter() {
+            scan_preferences.push(FullScanPreference::from(pref));
+        }
+        Self { scan_preferences }
+    }
+
+    /// Override the default scan parameters with the one from the config file or command line.
+    pub fn load_params(mut self, params: HashMap<String, ScanPrefValue>) -> Self {
```
Tehforsch, line 311: This is still called param when it really is a preference. Would prefer consistent naming here

rust/src/scanner/tests.rs
```diff
@@ -217,14 +218,15 @@ fn parse_meta_data(filename: &str, code: &str) -> Option<Nvt> {
     let executor = nasl_std_functions();
     let loader = |_: &str| code.to_string();
     let scan_id = ScanID(filename.to_string());
-
+    let scan_params = Vec::default();
```
Tehforsch, line 221: same here

rust/src/scanner/vt_runner.rs
```diff
@@ -37,6 +37,7 @@ pub struct VTRunner<'a, S: ScannerStack> {
     stage: Stage,
     param: Option<&'a Vec<Parameter>>,
     scan_id: ScanID,
+    scan_params: &'a Vec<ScanPreference>,
```
Tehforsch, line 40: same here

rust/src/scanner/preferences/preference.rs
```diff
@@ -217,3 +219,121 @@ pub const PREFERENCES: [ScanPreferenceInformation; 22] = [
...
+        for (param_id, param_val) in params.into_iter() {
+            if let Some(a) = self.scan_preferences.iter_mut().find(|p| p.id == *param_id) {
+                a.default = param_val.clone();
+            }
+        }
+        self
+    }
```
Tehforsch, line 318: This method is currently O(n^2) for no real reason. For example we could do
```rust
    /// Override the default scan parameters with the one from the config file or command line.
    pub fn load_params(&mut self, mut params: HashMap<String, ScanPrefValue>) {
        for p in self.scan_preferences.iter_mut() {
            if let Some(v) = params.remove(&p.id) {
                p.default = v;
            }
        }
    }
```

rust/src/openvasd/controller/context.rs
```diff
@@ -88,6 +93,14 @@ impl<S, DB, T> ContextBuilder<S, DB, T> {
         self
     }
 
+    /// Set scanner preferences
+    pub fn scan_preferences(mut self, prefs: HashMap<String, preference::ScanPrefValue>) -> Self {
+        let prefs_bind = preference::ScanPreferences::new();
+        let prefs = prefs_bind.load_params(prefs);
+        self.preferences = Some(prefs);
+        self
+    }
```
Tehforsch, line 102: If we remove the self-return on the `load_params` method, this simply becomes

```rust
    pub fn scan_preferences(mut self, prefs: HashMap<String, preference::ScanPrefValue>) -> Self {
        let mut prefs_bind = preference::ScanPreferences::new();
        prefs_bind.load_params(prefs);
        self.preferences = Some(prefs_bind);
        self
    }
```

rust/src/scanner/preferences/preference.rs
Tehforsch, line 311: Since this method overrides defaults, can we call it `override_defaults`?

Tehforsch (changes requested):

Tehforsch (approved):
Thanks!

PR #1895, opened 2025-04-24
Fix: solve issue when using cmake>4.0

### LINKED ISSUE #1889: CMake 4.0 removed compatibility with CMake < 3.5
### Expected behavior

Successful compilation with `CMake 4.0`

### Actual behavior

`CMake` removed with their newest release (`4.0.0`) compatibility with `CMake < 3.5`.
It can be overriden, but best it would be addressed upstream. Especially if the compatibility was/is necessary.

### Steps to reproduce

Compile the source with CMake 4.0

```
CMake Error at CMakeLists.txt:5 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```

### Environment

**Operating system:**

Linux

jjnicola:
**What**:
Solve issue when using cmake>4.0
Closes #1889 
Jira: SC-1296

**Why**:
Cmake v4.0 or later don't work anymore with deprecated stuff from cmake version < 3.5. Also, it triggers a warning if the minimum version is set to somethink <3.10.

Therefore, setting the minimum required version to 3.10.

I have tested with version 4.0.1 and cmake works as expected, so nothing else to be changed beside the version. Also, we need to support at least cmake 3.18, the one in Debian Bullseye, one of our supported base systems.

**How**:
Install cmake 4.0 or later and try to compile from sources.

**Checklist**:

- [ ] Tests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: GPL-2.0-or-later
 
-cmake_minimum_required (VERSION 3.0)
+cmake_minimum_required (VERSION 3.10)
 
 message ("-- Configuring the Scanner...")
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -71,7 +71,7 @@ if (NOT CMAKE_BUILD_TYPE MATCHES "Release")
     add_custom_target (revisiontag ALL)
 
     # creates gitversion.h using cmake script
-    add_custom_command (TARGET revisiontag COMMAND ${CMAKE_COMMAND}
+    add_custom_command (TARGET revisiontag POST_BUILD COMMAND ${CMAKE_COMMAND}
                         -DSOURCE_DIR=${CMAKE_SOURCE_DIR}
                         -P ${CMAKE_SOURCE_DIR}/cmake/GetGit.cmake)
 


Kraemii:
I now get the following warning:
```
CMake Warning (dev) at src/CMakeLists.txt:74 (add_custom_command):
  Exactly one of PRE_BUILD, PRE_LINK, or POST_BUILD must be given.  Assuming
  POST_BUILD to preserve backward compatibility.

  Policy CMP0175 is not set: add_custom_command() rejects invalid arguments.
  Run "cmake --help-policy CMP0175" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
This warning is for project developers.  Use -Wno-dev to suppress it.
```
`cmake --help-policy CMP0175`:
```
CMP0175
-------

.. versionadded:: 3.31

``add_custom_command()`` rejects invalid arguments.

CMake 3.30 and earlier silently ignored unsupported keywords and missing or
invalid arguments for the different forms of the ``add_custom_command()``
command. CMake 3.31 implements more rigorous argument checking and will flag
invalid or missing arguments as errors.

The ``OLD`` behavior of this policy will accept the same invalid keywords or
arguments as CMake 3.30 and earlier. The ``NEW`` behavior will flag the
following as errors that previously went unreported:

* The ``OUTPUT`` form does not accept ``PRE_BUILD``, ``PRE_LINK``, or
  ``POST_BUILD`` keywords.
* When the ``APPEND`` keyword is given, the ``OUTPUT`` form also does not
  accept ``BYPRODUCTS``, ``COMMAND_EXPAND_LISTS``, ``DEPENDS_EXPLICIT_ONLY``,
  ``DEPFILE``, ``JOB_POOL``, ``JOB_SERVER_AWARE``, ``USES_TERMINAL``, or
  ``VERBATIM`` keywords.
* The ``TARGET`` form requires exactly one of ``PRE_BUILD``, ``PRE_LINK``, or
  ``POST_BUILD`` to be given.  Previously, if none were given, ``POST_BUILD``
  was assumed, or if multiple keywords were given, the last one was used.
* The ``TARGET`` form does not accept ``DEPENDS``, ``DEPENDS_EXPLICIT_ONLY``,
  ``DEPFILE``, ``IMPLICIT_DEPENDS``, ``MAIN_DEPENDENCY``, ``JOB_POOL``, or
  ``JOB_SERVER_AWARE`` keywords.
* The ``TARGET`` form now requires at least one ``COMMAND`` to be given.
* If a keyword expects a value to be given after it, but no value is provided,
  that was previously treated as though the keyword was not given at all.
* The ``COMMENT`` keyword expects exactly one value after it.  If multiple
  values are given, or if the ``COMMENT`` keyword is given more than once,
  this is an error.

This policy was introduced in CMake version 3.31.
It may be set by ``cmake_policy()`` or ``cmake_minimum_required()``.
If it is not set, CMake warns, and uses ``OLD`` behavior.

.. note::
  The ``OLD`` behavior of a policy is
  ``deprecated by definition``
  and may be removed in a future version of CMake.
```

According to the help message, this was always inside the `src/CMakeLists.txt` but was silently ignored before.

PR #1899, opened 2025-04-29
Transport value for sockets

Kraemii:
This commit corrects the KB Item used for Port transport and also reading the transport from the KB, in case no transport was given. For Auto Transport the used transport is now set within the KB, after a connection is successfully established.

### DIFF
--- a/rust/src/nasl/builtin/network/socket.rs
+++ b/rust/src/nasl/builtin/network/socket.rs
@@ -427,36 +427,49 @@ fn open_sock_tcp_vhost(
     port: u16,
     vhost: &str,
     transport: i64,
-) -> Result<Option<NaslSocket>, SocketError> {
-    if transport < 0 {
-        // TODO: Get port transport and open connection depending on it
-        todo!()
-    }
+) -> Result<Option<NaslSocket>, FnError> {
+    let mut transport = if transport.is_negative() {
+        context.get_port_transport(port)?.unwrap_or(1)
+    } else {
+        transport
+    };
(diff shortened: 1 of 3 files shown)


rust/src/nasl/builtin/network/socket.rs
```diff
@@ -427,36 +427,49 @@ fn open_sock_tcp_vhost(
...
-            .map(|tcp| NaslSocket::Tcp(Box::new(tcp)))
-            .ok(),
-    )
+
+    let conn = TcpConnection::connect(addr, port, tls, timeout, bufsz, get_retry(context))
+        .map(|tcp| NaslSocket::Tcp(Box::new(tcp)))
+        .ok();
+    if set_transport {
+        context.set_port_transport(port, transport as usize)?;
+    }
```
Tehforsch, line 471: Does this have to be done after we initialize the connection? If not we could just move this into the `OpenvasEncaps::Auto)` match arm and get rid of `set_transport` altogether

Kraemii: This is really hard to answer, as this might have some side effects. At least in the C implementation it is done, after the connection is established. It somehow also feels wrong to set this key, before a connection is established. But yes the code would be cleaner and easier to read.

Tehforsch: Ok, thanks - let's leave it as it is

Tehforsch (approved):
Let's wait for CI before merging

Issue #1488, opened 2023-09-12
Memory leak on scan


y0d4a:
When OpenVAS start to scanning

> https://developer.hashicorp.com/boundary/docs 
> https://developer.hashicorp.com/vault

we are getting memory leak by process:

> gb_loghj_ CVE-2021-44228_http_web_dirs_active.nasl

This seems is some kind of bug.
If you can, replicate it, all software are free

ArnoStiefvater:
Hey @y0d4a

Thanks for creating the issue.

Just to make sure. You are scanning the Products "boundary" and "Vault" from haschicorp?

How do you know that the posted nasl script is the culprit? Do you have additional logs or reasons to believe so?

y0d4a:
Hi, yes those two products.
I saw with monitoring the process, when he come to IPs with that services he scanning for very long time and on the end, he used all of memory and interrupt the scan (sometimes crash openvas service - when is run from docker)

wdoekes:
In https://github.com/greenbone/ospd-openvas/issues/974 I've commented my findings. Summarizing:

## Example of "leak"

I dropped the db that seemed to be transient:

```diff
-server_time_usec:1709405329080351
-uptime_in_seconds:588
+server_time_usec:1709406156005681
+uptime_in_seconds:1415
...
 # Memory
-used_memory:3344099096
-used_memory_human:3.11G
-used_memory_rss:3001507840
-used_memory_rss_human:2.80G
+used_memory:171441592
+used_memory_human:163.50M
+used_memory_rss:194514944
+used_memory_rss_human:185.50M
...

 # Keyspace
 db0:keys=1,expires=0,avg_ttl=0
 db1:keys=177456,expires=0,avg_ttl=0
-db6:keys=3468,expires=0,avg_ttl=0
```
Dropping those 3400 entries in db6 freed a whopping 2GB.

The keys I freed look like:
```console
$ sudo docker exec -it greenbone-community-container_redis-server_1 redis-cli -s /run/redis/redis.sock -n 6
redis /run/redis/redis.sock[6]> keys *
   1) "Cache/node.example.com/8200/excluding_404_body/URL_/ui/vcav-bootstrap/rest/vcav-providers/config.neon"
   2) "Cache/node.example.com/8200/excluding_404_body/URL_/ui/vcav-bootstrap/rest/WEB-INF/local.properties"
[...]

PR #1904, opened 2025-05-06
C nasl kb usage

Kraemii:
Currently it can happen, that script try to allocate huge amount of memory in the KB, which leads
to flooding the RAM and crashing redis and openvas. The introduced config `max_mem_kb` limits
the maximum KB traffic (in MB) to the specified value for a single script. Only written strings
are tracked and only an estimation, as we do not know, if an item is replaced or appended. Also
the key size is not counted as well. If the value is not set or <= 0, the maximum amount is
unlimited (Default).

Jira: SC-919
Issue: https://github.com/greenbone/openvas-scanner/issues/1488
osp-openvas: https://github.com/greenbone/ospd-openvas/pull/1039

To test this, I created a custom feed with two script, each setting KB items with a size of 1MB in a loop and see, if the log message is shown, that the maximum amount of KB traffic for the script is reached. At the same time look at the memory used by redis and see, that redis' memory consumption does not grow anymore.

### DIFF
--- a/misc/plugutils.c
+++ b/misc/plugutils.c
@@ -37,6 +37,51 @@
 /* Used to allow debugging for openvas-nasl */
 int global_nasl_debug = 0;
 
+/* Track amount of KB memory used */
+static size_t kb_usage = 0;
+
+static size_t max_kb_usage;
+
+void
+init_kb_usage (void)
+{
+  const char *usage_char;
+  int usage_int;
+  usage_char = prefs_get ("max_mem_kb");
+  if (usage_char)
+    {
+      usage_int = atoi (usage_char);
+      if (usage_int < 0)
+        {
+          max_kb_usage = 0;
+        }
+      else
+        {
+          max_kb_usage = (size_t) usage_int * 1024 * 1024;
+        }
+    }
+  else
+    max_kb_usage = 0;
+}
+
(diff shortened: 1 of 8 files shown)


misc/plugutils.c
```diff
@@ -37,6 +37,51 @@
...
+add_kb_usage (struct script_infos *args, size_t size)
+{
+  if (max_kb_usage == 0)
+    return 0;
+  if (kb_usage > max_kb_usage)
+    return -1;
+  kb_usage += size;
+  if (kb_usage > max_kb_usage)
+    {
+      g_warning ("KB usage exceeded %lu MB. Unable to store any further KB "
```
jjnicola, line 77: I would add a flag to show this message once per script per host. Otherwise it will flood the log file.

jjnicola (approved):

jjnicola (changes requested):
Just the comment about the log

misc/plugutils.c
Kraemii: It is only shown once per script per host, as it is only shown the first time the limit is reached. After that the function returns immediately

jjnicola (approved):

jjnicola (approved):

PR #1907, opened 2025-05-13
Add: cancellation of running CI jobs on new commit

henning:
## What

Add cancellation of running CI jobs on new commit in same ref
to all workflows.

## Why

We usually don't need to continue running old jobs when a new commit comes in that starts the same jobs again.

This saves us some running time and resources.

## References
see DEVOPS-1538

## Checklist

- [NA] Tests

### DIFF
--- a/.github/workflows/push-helm-chart.yml
+++ b/.github/workflows/push-helm-chart.yml
@@ -12,7 +12,6 @@ on:
       token:
         required: true
 
-
 jobs:
   helm:
     runs-on: self-hosted-generic
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,4 +1,5 @@
 name: "release"
+
 on: 
   workflow_call:
     inputs:
@@ -31,7 +32,6 @@ on:
         required: true
       gpg_pass:
         required: true
-
 
 # This job first determines the target branch of the closed pull request. If the target branch is "main",
 # then the latest release tag is used. If no release tag exists, it is set to 0.1.0. If it is a release
--- a/.github/workflows/auto_label.yml
+++ b/.github/workflows/auto_label.yml
@@ -3,6 +3,10 @@ name: Labeler
 on:
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 permissions:
   pull-requests: write
   contents: read
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -8,6 +8,10 @@ on:
   schedule:
     - cron: '30 5 * * 0' # 5:30h on Sundays
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   analyze:
     name: Analyze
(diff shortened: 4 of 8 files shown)


.github/workflows/build.yml
```diff
@@ -2,6 +2,10 @@ name: "Build"
 
 on: [workflow_call]
 
+concurrency:
```
nichtsfrei, line 5: The build yml is integrated into `control.yml` therefore I don't think this is necessary here as it should be controlled by control.yml. See the `on workflow_call`.

.github/workflows/functional.yaml
```diff
@@ -3,6 +3,10 @@ name: functional
 on:
   workflow_call:
 
+concurrency:
```
nichtsfrei, line 6: The functional yaml is integrated into control.yml therefore I don't think this is necessary here as it should be controlled by control.yml. See the on workflow_call.

.github/workflows/init.yaml
```diff
@@ -43,6 +43,10 @@ on:
         description: "The name of the project to be released"
         value: ${{ jobs.init.outputs.project }}
 
+concurrency:
```
nichtsfrei, line 46: The init yaml is integrated into control.yml therefore I don't think this is necessary here as it should be controlled by control.yml. See the on workflow_call.

.github/workflows/linting.yml
```diff
@@ -2,6 +2,10 @@ name: "Linting"
 
 on: [workflow_call]
 
+concurrency:
```
nichtsfrei, line 5: Same as above.

.github/workflows/push-container-oldstable.yml
```diff
@@ -15,6 +15,10 @@ on:
       dockerhub_token:
         required: true
 
+concurrency:
```
nichtsfrei, line 18: Same as above.

.github/workflows/push-container-testing.yml
```diff
@@ -15,6 +15,10 @@ on:
       dockerhub_token:
         required: true
 
+concurrency:
```
nichtsfrei, line 18: Same as above.

.github/workflows/push-container.yml
```diff
@@ -33,6 +33,10 @@ on:
       mattermost_webhook_url:
         required: true
 
+concurrency:
```
nichtsfrei, line 36: Same as above.

.github/workflows/push-helm-chart.yml
```diff
@@ -12,6 +12,9 @@ on:
       token:
         required: true
 
+concurrency:
```
nichtsfrei, line 15: Same as above.

.github/workflows/release.yml
```diff
@@ -33,6 +34,10 @@ on:
         required: true
 
 
+concurrency:
```
nichtsfrei, line 37: Same as above.

.github/workflows/tests.yml
```diff
@@ -2,6 +2,10 @@ name: "Unit-Tests"
 
 on: [workflow_call]
 
+concurrency:
```
nichtsfrei, line 5: Same as above.

nichtsfrei (changes requested):

Tehforsch (approved):
Thank you!

PR #1911, opened 2025-05-20
Add: NASL builtin functions for port states in rust

Kraemii:
Jira: SC-1306, SC-1120


rust/src/nasl/utils/context.rs
```diff
@@ -889,6 +889,53 @@ impl<'a> Context<'a> {
...
+            .find(|x| x.id == key)
+            .and_then(|x| x.value.parse::<i64>().ok())
+    }
+
+    pub fn get_preference_string(&self, key: &str) -> Option<String> {
+        self.scan_preferences
+            .iter()
+            .find(|x| x.id == key)
+            .map(|x| x.value.clone())
+    }
```
Tehforsch, line 914: Nit: Could write this as a single fn with a signature like `get_preference<T: FromPreference>(&self, key: &str) -> Option<T>`.  Debatable if that is really worth it for three cases.

Tehforsch (approved):
Minor comment that can you can ignore if you want

rust/src/nasl/utils/context.rs
Kraemii: As there are only three cases and will most probably not extended I will leave it, as it is.

rust/src/nasl/utils/context.rs
```diff
@@ -904,6 +904,50 @@ impl<'a> Context<'a> {
...
+
+    pub fn get_port_state(&self, port: u16, protocol: Protocol) -> Result<bool, FnError> {
+        match protocol {
+            Protocol::TCP => {
+                if !self.target.ports_tcp.contains(&port)
+                    || self.get_kb_item(&KbKey::Host(kb::Host::Tcp))?.is_empty()
+                {
+                    return Ok(!self.get_preference_bool("unscanned_ports").unwrap_or(true));
```
jjnicola: Do you mean `unscanned_closed` preference?
```suggestion
                    return Ok(!self.get_preference_bool("unscanned_closed").unwrap_or(true));
```

rust/src/nasl/utils/context.rs
```diff
@@ -904,6 +904,50 @@ impl<'a> Context<'a> {
...
+                }
+                self.get_single_kb_item(&KbKey::Port(kb::Port::Tcp(port.to_string())))
+            }
+            Protocol::UDP => {
+                if !self.target.ports_udp.contains(&port)
+                    || self.get_kb_item(&KbKey::Host(kb::Host::Udp))?.is_empty()
+                {
+                    return Ok(!self
+                        .get_preference_bool("unscanned_ports_udp")
```
jjnicola: same as above.

Tehforsch (approved):

jjnicola (approved):

PR #1915, opened 2025-05-20
Fix: scan id which is already in use

jjnicola:
**What**:
This is about given that you post a scan with fixed scan id which is already in use we should return not acceptable instead of creating double.

SC-1291

**Why**:

**How**:
Post a scan config with a fixed scanid many times

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/doc/openapi.yml
+++ b/rust/doc/openapi.yml
@@ -232,7 +232,7 @@ paths:
                   $ref: "#/components/examples/scan_id"
         "400":
           description: "Bad Request body"
-        "406":
+        "403":
           description: "ScanID already in use"
 
   /scans/preferences:
--- a/rust/doc/openapi.yml
+++ b/rust/doc/openapi.yml
@@ -232,6 +232,8 @@ paths:
                   $ref: "#/components/examples/scan_id"
         "400":
           description: "Bad Request body"
+        "406":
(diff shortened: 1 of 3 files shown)


rust/doc/openapi.yml
```diff
@@ -232,6 +232,8 @@ paths:
                   $ref: "#/components/examples/scan_id"
         "400":
           description: "Bad Request body"
+        "406":
```
Kraemii, line 235: I guess status code `403` would be more fitting. According to https://en.wikipedia.org/wiki/List_of_HTTP_status_codes `406` is used, if the generated contend does not conform to the accept header, whereas `403` can be used if something is only allowed to exist once.

rust/src/openvasd/response.rs
```diff
@@ -368,4 +368,17 @@ impl Response {
         };
         self.create(hyper::StatusCode::NOT_ACCEPTABLE, &value)
     }
+
+    pub fn not_accepted_with_reason<T>(&self, got: &T, reason: String) -> Result
```
Kraemii, line 372: If using the status code `403` this function has to be adjusted as well

Kraemii (commented):
I just checked the HTTP status code and think the code `403` is more fitting
Except of that it looks good

PR #1920, opened 2025-05-27
Implement nasl builtin start_denial() and end_denial()

jjnicola:
**What**:
Implement nasl builtin start_denial() and end_denial()
SC-1305

This functions require to store internal data. Denial port and if the service was alive before starting the test. Therefore, this PR includes a change that makes the Register mutable.

Also, extends the Context with the alive_test_methods member, so an alive test can be performed as ultimate resource.

**Why**:

**How**:

**Checklist**:

- [ ] Tests


rust/Cargo.toml
```diff
@@ -96,6 +96,7 @@ openssl = { version = "0.10.72", features = ["vendored"] }
 blowfish = "0.9.1"
 rc4 = "0.1.0"
 dsa = "0.6.3"
+cargo-expand = "1.0.106"
```
Tehforsch: Should be removed. You can also remove the changes to Cargo.lock then

rust/src/nasl/builtin/raw_ip/denial.rs
```diff
@@ -0,0 +1,102 @@
...
+
+#[nasl_function]
+fn start_denial(context: &Context, script_info: &mut ScriptInfo) -> Result<NaslValue, FnError> {
+    let retry = if let Some(p) = context
+        .scan_params()
+        .find(|p| p.id == "checks_read_timeout")
+    {
+        p.value.parse::<u8>().unwrap_or(5)
+    } else {
+        5
```
Tehforsch, line 22: I'd prefer if 5 was a named const, i.e. DEFAULT_TIMEOUT or similar

rust/src/nasl/builtin/raw_ip/denial.rs
```diff
@@ -0,0 +1,102 @@
...
+#[nasl_function]
+async fn end_denial(context: &Context<'_>, register: &Register, script_info: &ScriptInfo) -> Result<NaslValue, FnError> {
+    let retry = if let Some(p) = context
+        .scan_params()
+        .find(|p| p.id == "checks_read_timeout")
+    {
+        p.value.parse::<u8>().unwrap_or(5)
+    } else {
+        5
+    };
```
Tehforsch, line 48: This is the same as above, so maybe extract this to a method?

Tehforsch (changes requested):
Very nice, thanks! Only had a few very minor comments.

I would personally prefer if we took the time to rename `Context` and `ScriptInfo` to something more descriptive (`ScanCtx` and `ScriptCtx`, for example), but this PR is already quite big, so it's fine if we delay this discussion for now.

PR #1921, opened 2025-05-27
Fix: feed transform. Strip feed root path prefix 

jjnicola:
**What**:
Fix: remove the trailing slash from the root path if necessary
SC-1317
Related to SC-1147

**Why**:

**How**:
run with trailing slash
`scannerctl feed transform -p /var/plugins/ `

and without trailing slash in the root path
`scannerctl feed transform -p /var/plugins`

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2988,9 +2988,9 @@ checksum = "925383efa346730478fb4838dbe9137d2a47675ad789c546d150a6e1dd4ab31c"
 
 [[package]]
 name = "prettyplease"
-version = "0.2.31"
+version = "0.2.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5316f57387668042f561aae71480de936257848f9c43ce528e311d89a07cadeb"
+checksum = "664ec5419c51e34154eec046ebcba56312d5a2fc3b09a06da188e1ad21afadf6"
 dependencies = [
  "proc-macro2",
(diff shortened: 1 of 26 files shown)


rust/src/feed/verify/mod.rs
```diff
@@ -456,7 +456,11 @@ fn get_all_plugins(loader: &FSPluginLoader) -> Vec<PathBuf> {
...
                 if let Some(p) = e.path().to_str() {
-                    let path_str = p[rp.to_string().len() + 1..].to_string();
+                    let path_str = if rp.ends_with('/') {
+                        p[rp.to_string().len()..].to_string()
+                    } else {
+                        p[rp.to_string().len() + 1..].to_string()
+                    };
                     let path = Path::new(&path_str).to_owned();
                     files.push(path);
                 }
```
Tehforsch, line 466: This felt a little too low-level to really be the right way to do this, so I discovered 
https://doc.rust-lang.org/std/path/struct.Path.html#method.strip_prefix

With this, we should be able to reduce this to

```rust
            if e.path().extension().is_some_and(|ext| ext == "nasl") {
                let relative_path = e.path().strip_prefix(Path::new(&rp)).unwrap();
                files.push(relative_path.to_owned());
            }
```

Tehforsch (changes requested):

jjnicola:
Thanks @Tehforsch !

Issue #1922, opened 2025-05-28
Enable Link-Time Optimization (LTO) and codegen-units = 1


zamazan4ik:
Hi!

I wasn't sure about where I should create this issue. Since it's not a question and related directly to the compiler flags - I created the issue here. Excuse me, if I am wrong - in this case, feel free to move it to a proper place.

I noticed that in the `Cargo.toml` file Link-Time Optimization (LTO) for the Rust part of the project is not enabled. I suggest switching it on since it will reduce the binary size (always a good thing to have) and will likely improve the application's performance (due to more aggressive compiler optimizations). If you want to read more about LTO and its possible modes, I recommend starting from [this](https://doc.rust-lang.org/rustc/codegen-options/index.html#lto) Rustc documentation.

I think you can enable LTO only for the Release builds so as not to sacrifice the developers' experience while working on the project, since LTO consumes an additional amount of time to finish the compilation routine. In this case, we can create a dedicated `[profile.optimized-dev]` profile where LTO will be disabled (so developers experience will not be affected). If we enable it on the Cargo profile level for the Release profile, in all places where Release profile is used, LTO will be applied "automatically". E.g., check `cargo-outdated` Release [profile](https://github.com/kbknapp/cargo-outdated/blob/master/Cargo.toml#L48). You also could be interested in other optimization options like `codegen-units = 1` - it also brings improvements over the current defaults.

Basically, it can be enabled with the following lines to the root Cargo.toml file:

```
[profile.release]
codegen-units = 1
lto = true
```

I have made quick tests (AMD Ryzen 9 5900x, Fedora 42, Rust 1.87, the latest version of the project at the moment, `CC=clang CXX=clang++ cargo build --release --workspace` command) - here are the results:

| Build config \ Binary name | feed-verifier | openvasd | scannerctl |
| --- | --- | --- | --- |
| Release | 1.2 Mib | 34 Mib | 27 Mib |
| Release + CG1 + FatLTO | 958 Kib | 26 Mib | 21 Mib |

Clean build time:

* Release: 1m 21s
* Release + `codegen-units = 1` + Fat LTO:  2m 57s

Also, it's possible to enable LTO for the C implementation as well, but I didn't test it on the project.

Thank you.

PR #1924, opened 2025-05-30
Add `release-lto` profile with LTO enabled.

Tehforsch:
Use this new profile in releases built via github actions.

Solves #1922 

TODO: 

- [x] Test artifacts are generated properly


rust/Cargo.toml
```diff
@@ -135,6 +135,10 @@ version = "0.1.0"
 edition = "2024"
 license = "GPL-2.0-or-later"
 
+[profile.release-lto]
+inherits = "release"
+lto = true
```
zamazan4ik, line 140: Additionally to LTO, I suggest enabling `codegen-units = 1` for `release-lto`. It will enable even more aggressive optimizations.

Tehforsch: Thanks - I had originally tested `codegen-units = 1` with a (admittedly not very representative) benchmark and found no difference and wanted to save the slight increase in compile times (we use release artifacts for our CIs so compile time is not entirely irrelevant), but I supposed that this could eventually make a difference for real scans, so I added it in after all.

PR #1925, opened 2025-05-30
Add: rust implementation for telnet_init() nasl function

jjnicola:
**What**:
Add: rust implementation for telnet_init() nasl function

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/src/nasl/builtin/network/socket.rs
+++ b/rust/src/nasl/builtin/network/socket.rs
@@ -239,17 +239,7 @@ async fn close(sockets: &mut NaslSockets, socket_fd: usize) -> Result<(), FnErro
     Ok(())
 }
 
-/// Send data on a socket.
-/// Args:
-/// takes the following named arguments:
-/// - socket: the socket, of course.
-/// - data: the data block. A string is expected here (pure or impure, this does not matter).
-/// - length: is optional and will be the full data length if not set
-/// - option: is the flags for the send() system call. You should not use a raw numeric value here.
-///
-/// On success the number of sent bytes is returned.
-#[nasl_function(named(socket, data, option, len))]
-async fn send(
+fn send_shared(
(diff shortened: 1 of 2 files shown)


rust/src/nasl/builtin/network/socket.rs
```diff
@@ -712,6 +734,68 @@ async fn ftp_get_pasv_port(sockets: &mut NaslSockets, socket: usize) -> Result<u
...
+        // and take either the one set for the socket or the default from preferences
+        buffer = recv_shared(sockets, socket, 3, Some(3), Some(Seconds(2)))?;
+        if buffer[0] != 255 || buffer.is_empty() || buffer.len() != 3 {
+            break;
+        }
+        if buffer[1] == 251 || buffer[1] == 252 {
+            buffer[1] = 254; /* WILL , WONT -> DON'T */
+        } else if buffer[1] == 253 || buffer[1] == 254 {
+            buffer[1] = 252 /* DO,DONT -> WONT */
+        };
```
Tehforsch: Could you add a small comment explaining why this is happening? Would be nice for people like me who know nothing about telnet :)

Tehforsch (approved):
Thank you! I had one minor comment (asking for a comment).

Another general thing you could do if you wanted to be fancy about this but that is absolutely not necessary is wrapping the `buffer` in some sort of struct that contains the internal byte buffer and contains getters/setters for the iac/code/option fields, making what's happening slightly more explicit.

rust/src/nasl/builtin/network/socket.rs
jjnicola: Done! :)
Also, added the struct for wrapping the first three bytes of the buffer

Tehforsch: Thanks!

PR #1926, opened 2025-06-02
Parser/Interpreter refactoring

Tehforsch:
This PR makes the following major changes:

- Use `codespan_reporting` to generate prettier error messages, especially useful for usage with `scannerctl`.
```
error: Not a valid assignment target.
  ┌─ examples/4_place_expr.nasl:1:1
  │
1 │ a(1) = 3;
  │ ^^^^ Not a valid assignment target.
```
- Improve generated error messages (and spans), especially for tokenization/parsing.
- Explicitly write out the NASL grammar, moving a large class of syntax errors from runtime to parsing time.
- Add a large number of tests for tokenization, parsing and interpreting. Move all existing tests in those categories to snapshot tests. Also adds a large number of tests to check for proper syntax error messages.
- Fix some cases of NASL behavior:
    - `if (!x = 3)` was previously allowed syntactically but would error at runtime. Since this construct is used quite a lot in the feed, allow it explicitly in the grammar by fixing the operator precedence.
    - Fix `a && (x = 3)` not short-circuiting, leading to unintended behavior.
- Unify loading/creating of NASL code via a `Code` interface that is used to handle loading/tokenization/parsing and emitting syntax errors.
- Update the nasl/grammar docs to more accurately reflect the current state
- (minor, unfinished): begin a list of features that we want to implement/drop in a future NASL version.


rust/src/nasl/syntax/metadata.rs
```diff
@@ -0,0 +1,53 @@
...
+    pub(crate) fn to_ast(self) -> Ast {
+        Ast::new(self.stmts)
+    }
+}
+
+impl Parse for DescriptionBlock {
+    fn parse(parser: &mut super::Parser) -> Result<Self> {
+        let mut if_: If = parser.parse()?;
+        if if_.else_branch.is_some() {
+            return Err(ErrorKind::InvalidDescriptionBlock(
```
nichtsfrei: question: why is this considered an error? From a script perspective it just means that it is not an description run, so if an script author wants to have a else branch and work in there why should we not allow that?

doc/manual/nasl/nasl-changes.md
```diff
@@ -0,0 +1,17 @@
...
+* Simplify the autoconversion rules (i.e. strings magically being turned into integers, arrays into bools and so on, ...). Prefer throwing errors over implicit conversions.
+
+## Obsolete operators
+Get rid of barely used operators such as the `x` operator and `>>>=`.
+
+## Control flow
+Disallow using `break` to escape from functions/the script.
+
+## Forking
+Remove the forking concept altogether.
```
nichtsfrei, line 14: did we? in this version?

doc/manual/nasl/nasl-grammar.md
```diff
@@ -0,0 +1,247 @@
...
+- Unary: A unary operation (e.g., `-a` or `!b`).
+- Assignment: An assignment expression (e.g., `x = y`)
+- Increments: Similarly, increments are also allowed (e.g. `++x` or `y++`)
+
+### Atoms
+An Atom is a fundamental expression type with no further decomposition (other than subexpressions). It includes:
+
+- Literals: Primitive values (e.g., integer, string, etc.).
+- Variables: A variable name.
+- Arrays: An array literal (e.g. `[1, 2, 3]`)
```
nichtsfrei, line 35: do we actually have arrays? I think value lookup `[1, "zwei", "polizei"]` is allowed?

```bash
user@be9782232c7a /w/g/o/rust (replace-storage-with-sqlite)> cat /tmp/aha.nasl
a = [1, "zwei", "polizei"];
display(a[0]);
display(a[1]);
display(a["zwei"]);

user@be9782232c7a /w/g/o/rust (replace-storage-with-sqlite)> target/release/scannerctl execute script /tmp/aha.nasl
1
zwei
zwei
```

doc/manual/nasl/nasl-grammar.md
```diff
@@ -0,0 +1,247 @@
...
+
+Note, in particular, that assignment has relatively high precedence, allowing expressions like
+```nasl
+if (!x = foo() ) { ... }
+```
+which will first assign the result of the `foo()` call to `x`, and then negate the value.
+
+## Loops and Control Flow
+### If
+If statements are written in the usual way as
```
nichtsfrei: I would remove `in the usual way` as it depends on the experience of the reader if it is usual or not and it doesn't bring value into the sentence.

doc/manual/nasl/nasl-grammar.md
```diff
@@ -0,0 +1,247 @@
...
+- **Null**
+
+# Current behavior
+## Type autoconversion
+**Booleans**
+Every type is automatically converted into booleans when a boolean is needed (for example in `if (x) { ... }`)
+- The undefined or null value is false
+- All integers except 0 are true
+- Strings are true if not empty and not "0".
+- Arrays are always true whether they are empty or not
```
nichtsfrei, line 209: at least in the rust version this is not true:
```bash
user@be9782232c7a /w/g/o/rust (replace-storage-with-sqlite)> cat /tmp/random.nasl
a = [];
if (a) display("hooray");
b = [1];
if (b) display("booray");
display("end");
user@be9782232c7a /w/g/o/rust (replace-storage-with-sqlite)> target/release/scannerctl execute script /tmp/random.nasl
booray
end
```

and for the C version: 

```
user@be9782232c7a /w/g/o/rust (replace-storage-with-sqlite) [1]> cat /tmp/random.nasl
a = [];
if (a) display("hooray");
b = [1];
if (b) display("booray");
display("end");
user@be9782232c7a /w/g/o/rust (replace-storage-with-sqlite)> openvas-nasl -X /tmp/random.nasl
** WARNING : packet forgery will not work
** as NASL is not running as root
lib  nasl-Message: 15:22:48.504: Parse error at or near line 1:
lib  nasl-Message: 15:22:48.504:     syntax error
lib  nasl-Message: 15:22:48.504: /tmp/random.nasl. There were 1 parse errors.
```

🗡️

nichtsfrei: Ok it is true for the C version: 
```
user@be9782232c7a /w/g/o/rust (replace-storage-with-sqlite)> cat /tmp/random.nasl
a = make_array();
if (a) display("hooray");
b = [1];
if (b) display("booray");
display("end");
user@be9782232c7a /w/g/o/rust (replace-storage-with-sqlite)> openvas-nasl -X /tmp/random.nasl
** WARNING : packet forgery will not work
** as NASL is not running as root
lib  nasl-Message: 15:25:30.541: hooray
lib  nasl-Message: 15:25:30.541: booray
lib  nasl-Message: 15:25:30.541: end
```

so that's a bug in the rust version then?

doc/manual/nasl/nasl-grammar.md
Tehforsch: Hm, I'm not sure I understand the requirements here. `[1, "zwei", "polizei"]["zwei"]` is allowed on `main`, that is correct. If we want to allow indexing via strings, I can introduce that back in, but it is not value lookup. The string simply gets converted into an integer and that happens to be `1` due to the `From<NaslValue> for i64` impl. See:

```nasl
a = [1, "zwei", "polizei"];
display(a["polizei"]);

a = [1, "polizei", "zwei"];
display(a["zwei"]);
```

Which yields:
```
zwei
polizei
```

doc/manual/nasl/nasl-grammar.md
Tehforsch: Good catch. I copied this from the doc without ever checking whether the implementation actually fulfills it. Fixed the impl and added a test.

PR #1927, opened 2025-06-02
Fix: get_port_transport

jjnicola:
**What**:
Fix: get_port_transport
 SC-1318

**Why**:
Currently, if the transport is not set for the port in the Kb, it returns an error. Now, it returns None, and and open_sock_tcp() will be able to open a socket with IP encapsulation (no tls).

**How**:

run scannerctl execute script test_open_socket.nasl -t localhost

test_open_socket.nasl:
```
open_sock_tcp(23);
```

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/src/nasl/utils/context.rs
+++ b/rust/src/nasl/utils/context.rs
@@ -858,9 +858,10 @@ impl<'a> Context<'a> {
         )
     }
 
-    pub fn get_port_transport(&self, port: u16) -> Result<Option<i64>, FnError> {
+    pub fn get_port_transport(&self, port: u16) -> Option<i64> {
         self.get_single_kb_item_inner(&KbKey::Transport(kb::Transport::Tcp(port.to_string())))
-            .map(|x| match x {
+            .ok()
+            .and_then(|item| match item {
                 KbItem::Number(n) => Some(n),
                 _ => None,
             })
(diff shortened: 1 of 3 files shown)


rust/src/nasl/utils/context.rs
```diff
@@ -846,9 +846,9 @@ impl<'a> Context<'a> {
 
     pub fn get_port_transport(&self, port: u16) -> Result<Option<i64>, FnError> {
         self.get_single_kb_item_inner(&KbKey::Transport(kb::Transport::Tcp(port.to_string())))
-            .map(|x| match x {
-                KbItem::Number(n) => Some(n),
-                _ => None,
+            .map_or(Ok(None), |x| match x {
+                KbItem::Number(n) => Ok(Some(n)),
+                _ => Ok(None),
```
Tehforsch: Note that at this point the function return type might as well be `Option<i64>`. Giving the caller the impression that we might return an error, even if we don't seems like a bad idea. After that, we might just write:

```rust
        self.get_single_kb_item_inner(&KbKey::Transport(kb::Transport::Tcp(port.to_string())))
            .ok()
            .and_then(|item| match item {
                KbItem::Number(n) => Some(n),
                _ => None,
            })
```

jjnicola: Thanks!

PR #1932, opened 2025-06-10
Fix: openapi specification for Notus API endpoint

Kraemii:
SC-1320


rust/doc/openapi.yml
```diff
@@ -758,48 +755,59 @@ components:
...
+        end:
+          description: "End version of the fixed range"
+          type: "string"
     Result:
       description: "A result of a Scan"
       type: "object"
       properties:
         id:
-          description: "An ID, which is unique for the scan. It is auto incremental and starts at 0 for the first result."
+          description: "An ID, which is unique for https://petstore.swagger.io/v2/swagger.jsonthe scan. It is auto incremental and starts at 0 for the first result."
```
jjnicola:
```suggestion
          description: "An ID, which is unique for https://petstore.swagger.io/v2/swagger.json scan. It is auto incremental and starts at 0 for the first result."
```

jjnicola (changes requested):
Just a small typo

rust/doc/openapi.yml
Kraemii: I wonder how that happened :laughing:

PR #1936, opened 2025-06-17
Fix: Calculate MTU dinamically.

jjnicola:
**What**:
Fix: Calculate MTU dinamically.
SC-1324

**Why**:
This functionality is under the feature `nasl-builtin-raw-ip` since it depends on libc. Currently, there is no rust native equivalent implementation of the C  struct

**How**:
Compile with and without the feature `nasl-builtin-raw-ip` and run the following nasl script:
Compare the output running the same script with `openvas-nasl`
```
display(get_mtu());
```

**Checklist**:

- [ ] Tests


Tehforsch (approved):
Thanks!

PR #1937, opened 2025-06-17
Fix: Solves TODOs

jjnicola:
**What**:
- Fix: Solves TODO. Use get_open_port()
- Fix: TODO related to vhosts
- Fix: solve TODO related to checks_read_timeout scan preference
SC-1325

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/src/nasl/builtin/network/socket.rs
+++ b/rust/src/nasl/builtin/network/socket.rs
@@ -552,8 +552,16 @@ async fn open_sock_tcp(
 
     // TODO: set timeout to global recv timeout * 2 when available
     let timeout = convert_timeout(timeout).unwrap_or(Duration::from_secs(10));
-    // TODO: for every vhost
-    let vhosts = ["localhost"];
+    let mut vhosts = context
+        .target()
+        .vhosts()
+        .clone()
+        .into_iter()
+        .map(|v| v.hostname().to_string())
+        .collect::<Vec<String>>();
+    if vhosts.is_empty() {
+        vhosts.push(context.target().ip_addr().to_string());
+    };
     let sockets: Vec<Option<NaslSocket>> = vhosts
         .iter()
         .map(|vhost| open_sock_tcp_vhost(context, addr, timeout, bufsz, port.0, vhost, transport))
--- a/rust/src/nasl/builtin/network/socket.rs
+++ b/rust/src/nasl/builtin/network/socket.rs
@@ -4,7 +4,7 @@
 
 use crate::nasl::{
     prelude::*,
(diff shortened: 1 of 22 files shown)


jjnicola:
depends on #1911

rust/src/nasl/builtin/network/socket.rs
```diff
@@ -195,8 +197,8 @@ impl NaslSockets {
         tcp: bool,
     ) -> Result<NaslValue, SocketError> {
         if tcp {
-            // TODO: set timeout to global recv timeout when available
-            let timeout = Duration::from_secs(10);
+            // Default timeout is from C implementation in nasl/exec.c:1681
+            let timeout = self.recv_timeout.unwrap_or(Duration::from_secs(5));
```
Tehforsch: If I remember correctly, the 5 second timeout is not just the default here, but in a lot of other places of the code too. If that's true, I think we should move it into a global constant.

Tehforsch: Also: Is the comment really necessary? Maybe it could go into the commit message instead.

rust/src/nasl/builtin/network/socket.rs
```diff
@@ -550,10 +557,25 @@ async fn open_sock_tcp(
...
+            .unwrap_or(5) as u64
+            * 2,
+    ));
+    let mut vhosts = context
+        .target()
+        .vhosts()
+        .clone()
+        .into_iter()
+        .map(|v| v.hostname().to_string())
+        .collect::<Vec<String>>();
```
Tehforsch: No need for cloning the whole vec:

```rust
    let mut vhosts = context
        .target()
        .vhosts()
        .iter()
        .map(|v| v.hostname().to_string())
        .collect::<Vec<String>>();
```

rust/src/scanner/preferences/preference.rs
```diff
@@ -349,3 +349,30 @@ impl ScanPreferences {
...
+            });
+        };
+    }
+
+    fn get_preference_int(&self, key: &str) -> Option<i64> {
+        self.iter()
+            .find(|x| x.id == key)
+            .and_then(|x| x.value.parse::<i64>().ok())
+    }
+}
```
Tehforsch: Before we begin defining extension traits, I'd rather go with the simpler solution of making the `Vec<ScanPreference>` a `ScanPreferences` struct that provides the helper methods directly. This also encapsulates the scan preferences nicely, so that we can change internal implementation details in the future.

Tehforsch (changes requested):

PR #1938, opened 2025-06-23
Add: kb scannerctl command line option.

jjnicola:
**What**:
Add: kb scannerctl command line option.
 Allow to set KbItem key value to scannerctl execute script command line

SC-1331

**Why**:

**How**:
run the following nasl script:
```
display(get_kb_item("TCP/PORTS"));
```
with:
`scannerctl execute script --kb="TCP/PORTS=443" examples/get_kb_item.nasl `

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/src/scannerctl/interpret/mod.rs
+++ b/rust/src/scannerctl/interpret/mod.rs
@@ -20,7 +20,10 @@ use scannerlib::{
             scan_ctx::{Ports, Target},
         },
     },
-    storage::{ScanID, items::nvt::Oid},
+    storage::{
+        ScanID,
+        items::{kb::KbContextKey, nvt::Oid},
+    },
 };
 use scannerlib::{nasl::utils::scan_ctx::ContextStorage, storage::inmemory::InMemoryStorage};
 use scannerlib::{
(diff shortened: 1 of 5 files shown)


rust/src/scannerctl/interpret/mod.rs
```diff
@@ -116,11 +119,25 @@ async fn run_on_storage<S: ContextStorage, L: Loader>(
...
     ports: Ports,
     script: &Path,
 ) -> Result<(), CliErrorKind> {
     let scan_id = ScanID(format!("scannerctl-{}", script.to_string_lossy()));
     let filename = script;
+    let kbctx = (
+        scan_id.clone(),
+        scannerlib::storage::Target(target.ip_addr().to_string()),
+    );
+    kb.iter().for_each(|s| {
```
Tehforsch: Personal preference: I find a for loop more natural than `for_each` (rust-analyzer can convert between the two)

rust/src/scannerctl/interpret/mod.rs
```diff
@@ -116,11 +119,25 @@ async fn run_on_storage<S: ContextStorage, L: Loader>(
...
+        if let Some((k, v)) = s.split_once("=") {
+            let storage_ctx = KbContextKey(kbctx.clone(), k.into());
+            let _ = storage
+                .dispatch(storage_ctx, v.into())
+                .map_err(CliErrorKind::StorageError);
```
Tehforsch: Should we not handle the error here? Since this is scannerctl, I think even an `unwrap` is fine, but silently ignoring it feels wrong.
If you want to return the result, changing the `for_each` to a for loop will also make it easier

Tehforsch (changes requested):
Nice!

PR #1940, opened 2025-06-27
Add: rust implementation of http nasl builtin functions 

jjnicola:
**What**:
Add: rust implementation of http nasl builtin functions
SC-1333

**Why**:

**How**:

**Checklist**:

- [ ] Tests


rust/src/nasl/builtin/http/mod.rs
```diff
@@ -25,14 +26,47 @@ use tokio::{
...
+    NaslSockets,
+    network::{
+        Port,
+        socket::{close_shared, open_sock_tcp_shared},
+    },
+};
+
+fn get_user_agent(context: &ScanCtx) -> Result<String, FnError> {
+    match context.get_single_kb_item(&KbKey::Custom(
+        "global_settings/http_user_agent".to_string(),
```
Tehforsch: Since this is an established key, maybe it's worth adding a variant to `KbKey`

rust/src/scanner/preferences/preference.rs
```diff
@@ -374,6 +374,22 @@ impl ScanPrefs {
...
+            self.0.push(ScanPreference {
+                id: String::from("vendor_version"),
+                value: vendor,
+            })
+        } else {
+            self.0.push(ScanPreference {
+                id: String::from("vendor_version"),
+                value: format!("{}_{}", env!("CARGO_PKG_NAME"), env!("CARGO_PKG_VERSION")),
+            });
+        };
```
Tehforsch:
```suggestion rust
        self.0.push(ScanPreference {
            id: String::from("vendor_version"),
            value: vendor.unwrap_or_else(|| format!("{}_{}", env!("CARGO_PKG_NAME"), env!("CARGO_PKG_VERSION"))),
        })
```

rust/src/nasl/builtin/http/mod.rs
```diff
@@ -408,17 +446,168 @@ impl NaslHttp {
...
+fn http_req_shared(
+    context: &ScanCtx,
+    keyword: Method,
+    port: Port,
+    item: String,
+    data: Option<String>,
+) -> Result<NaslValue, FnError> {
+    let p: u16 = port.into();
+    let tmp_key = format!("http/{p}");
+    let mut request = match context.get_kb_item(&KbKey::from(tmp_key))?.first() {
```
Tehforsch: I didn't check, but you could probably use `context.get_single_kb_item::<u16>(&KbKey::from(tmp_key))` instead (unless we expect multiple entries, in which case I'm not sure `.first()` is the right logic to begin with

rust/src/nasl/builtin/http/mod.rs
```diff
@@ -25,14 +26,47 @@ use tokio::{
...
+            {
+                Some(vendor) => format!("Mozilla/5.0 [en] (X11, U; {vendor})"),
+                _ => format!(
+                    "Mozilla/5.0 [en] (X11, U; {}_{}",
+                    env!("CARGO_PKG_NAME"),
+                    env!("CARGO_PKG_VERSION")
+                ),
+            };
+            context
+                .set_single_kb_item("global_settings/http_user_agent".into(), ua.clone())
```
Tehforsch: so we could re-use it here (see above)

Tehforsch (changes requested):
Thanks!

rust/src/nasl/builtin/http/mod.rs
```diff
@@ -25,14 +26,48 @@ use tokio::{
...
+    match context.get_single_kb_item(&KbKey::GlobalSettings(GlobalSettings::HttpUserAgent)) {
+        Ok(ua) => Ok(ua),
+        _ => {
+            let ua = match context
+                .scan_preferences
+                .get_preference_string("vendor_version")
+            {
+                Some(vendor) => format!("Mozilla/5.0 [en] (X11, U; {vendor})"),
+                _ => format!(
+                    "Mozilla/5.0 [en] (X11, U; {}_{}",
```
Tehforsch: I dont know if this matters, but it looks like there is a missing parenthesis.

Tehforsch (dismissed):

rust/src/nasl/builtin/http/mod.rs
jjnicola: ups!

PR #1944, opened 2025-07-08
Begin implementation of `find_service` builtin

Tehforsch:
Add `run_find_service` implementation.

This is a very rudimentary implementation of the `find_service` builtin. It only includes a limited number of services but hopefully lays the foundation of something ... great.

In this implementation, service detection is not hardcoded, but determined by a list of services in a JSON file. In this JSON, the services are defined as metadata (such as name and description), possible ports, along with some detection methods as well as custom KB entries to add when the service is detected.

For now, only a smaller number of services are detected:
FTP, telnet, SSH, web/http, SMTP, POP3, IMAP and NetBus.

For testing, a dockerfile along with a NASL script and a bash script to
start the dockerfile and run the NASL script against the container is
provided.}

jira: SC-1209

### DIFF
--- a/rust/src/nasl/builtin/find_service/mod.rs
+++ b/rust/src/nasl/builtin/find_service/mod.rs
@@ -309,6 +309,8 @@ fn read_from_tcp_at_port(target: IpAddr, port: u16) -> Result<ReadResult, FindSe
 }
 
 fn try_http_request(target: IpAddr, port: u16) -> Result<Vec<u8>, FindServiceError> {
+    // TODO: Use a more robust approach that also takes HTTP2 into
+    // account (such as using h2)
     let mut socket = make_tcp_socket(target, port, 0)?;
     let http_request = b"GET / HTTP/1.0\r\n\r\n";
     match socket {
--- a/rust/src/nasl/builtin/find_service/mod.rs
+++ b/rust/src/nasl/builtin/find_service/mod.rs
@@ -229,8 +229,11 @@ impl ServiceDetector {
         port: u16,
     ) -> Result<(), FnError> {
         let banner = service.banner;
-        // This is O(n^2), but simpler than using a HashMap and performance
-        // should really not matter here.
+        // The following lookup is in O(n^2). However, since the
+        // number of services is likely to be small, this is very
+        // unlikely to become a performance problem, so that I prefer
+        // the simplicity of this approach (as opposed to using a
+        // HashMap).
         let service = self
             .services
             .services
--- a/rust/src/nasl/builtin/find_service/mod.rs
+++ b/rust/src/nasl/builtin/find_service/mod.rs
@@ -1,0 +1,415 @@
+// SPDX-FileCopyrightText: 2023 Greenbone AG
+//
+// SPDX-License-Identifier: GPL-2.0-or-later WITH x11vnc-openssl-exception
+
+use std::{
+    collections::HashMap,
+    io::{self, Write},
+    net::IpAddr,
+    time::Duration,
+};
+
+use serde::{Deserialize, Serialize};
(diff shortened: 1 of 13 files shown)


rust/src/nasl/builtin/find_service/mod.rs
```diff
@@ -0,0 +1,415 @@
...
+    }
+}
+
+struct ServiceDetector {
+    services: ServiceDefinitions,
+}
+
+impl ServiceDetector {
+    fn new() -> Result<Self, FindServiceError> {
+        let json_content = include_str!("../../../../data/service_definitions.json");
```
nichtsfrei: question: wouldn't it be more stable to use
```rust
concat!(env!("CARGO_MANIFEST_DIR"), "/data/service_definitions.json")
```
?

rust/src/nasl/builtin/find_service/mod.rs
```diff
@@ -0,0 +1,415 @@
...
+
+    fn handle_detected_service(
+        &self,
+        context: &ScanCtx<'_>,
+        service: DetectedService,
+        port: u16,
+    ) -> Result<(), FnError> {
+        let banner = service.banner;
+        // This is O(n^2), but simpler than using a HashMap and performance
+        // should really not matter here.
```
nichtsfrei: nit: although I really understand the annoyance, but a later reader may not. 

A more neutral tone that explains why micro optimization to a hashmap would be wasted would be better.

rust/src/nasl/builtin/find_service/mod.rs
```diff
@@ -0,0 +1,415 @@
...
+fn try_http_request(target: IpAddr, port: u16) -> Result<Vec<u8>, FindServiceError> {
+    let mut socket = make_tcp_socket(target, port, 0)?;
+    let http_request = b"GET / HTTP/1.0\r\n\r\n";
+    match socket {
+        NaslSocket::Tcp(ref mut tcp_conn) => {
+            tcp_conn
+                .write_all(http_request)
+                .map_err(|e| SocketError::IO(e))?;
+
+            // Read response
```
nichtsfrei: Nit: it literally says read_with_timeout 2 lines beneath. Unnecessary comment.

rust/src/nasl/builtin/find_service/mod.rs
```diff
@@ -0,0 +1,415 @@
...
+            Ok(ReadResult::Data(buf))
+        }
+        Err(e) if e.kind() == io::ErrorKind::TimedOut => Ok(ReadResult::Timeout),
+        Err(e) => Err(SocketError::IO(e).into()),
+    }
+}
+
+fn try_http_request(target: IpAddr, port: u16) -> Result<Vec<u8>, FindServiceError> {
+    let mut socket = make_tcp_socket(target, port, 0)?;
+    let http_request = b"GET / HTTP/1.0\r\n\r\n";
```
nichtsfrei, line 315: question: what about http2? Wouldn't it be better to use a http client for that?

rust/src/nasl/builtin/find_service/mod.rs
Tehforsch: I wasnt quite sure what you meant by annoyance, but I tried to rephase the comment to hopefully sound more neutral.

rust/src/nasl/builtin/find_service/mod.rs
Tehforsch: Probably. I wanted to keep this simple for this prototype, but I guess doing this manually might just make things more complicated. Should I change this to use `h2` instead? Or something different entirely?

nichtsfrei (approved):

PR #1951, opened 2025-07-11
Add: support more alive test method

jjnicola:
**What**:
Add: support ICMPv6 alive test method
SC-1340

**Why**:
They are missing

**How**:

ICMPv6 (you need to configure the IPv6 network):
`sudo target/debug/scannerctl alivetest --icmp -t 5858::2`

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/data/nasl_syntax/smb_nt.inc
+++ b/rust/data/nasl_syntax/smb_nt.inc
@@ -21,7 +21,7 @@ ntlmv2_flag  = string( get_kb_item( "SMB/dont_send_ntlmv1" ) );
 ct_flag      = string( get_kb_item( "SMB/dont_send_in_cleartext" ) );
 
 ## An ISSUE reported as #71142, registry methods of smb_nt.inc are currently failing
-## against a Windows Vista Ulimate SP1 and Windows Server (R) 2008 SP1 systems. Moreover,
+## against a Windows Vista Ultimate SP1 and Windows Server (R) 2008 SP1 systems. Moreover,
 ## Windows Vista has already reached End of Life.
 ## Adding a workaround to check if target is Windows Vista Service Pack 1 or Windows Server
 ## (R) 2008 Service Pack 1 and if so force target to use SMB1 instead of SMB2, which is


rust/src/alive_test/alive_test.rs
```diff
@@ -151,17 +83,18 @@ impl TryFrom<&[u8]> for EtherTypes {
...
 }
 
-fn process_ip_packet(packet: &[u8]) -> Result<Option<AliveHostCtl>, AliveTestError> {
+fn process_ipv4_packet(packet: &[u8]) -> Result<Option<AliveHostCtl>, AliveTestError> {
     let pkt = Ipv4Packet::new(&packet[16..]).ok_or_else(|| {
         AliveTestError::CreateIpPacketFromWrongBufferSize(packet.len() as i64 - 16)
     })?;
-    let hl = pkt.get_header_length() as usize;
+
+    let hl = pkt.get_header_length() as usize * 4 + 16; // in bytes: hl *32 bits words / 8bits per byte
```
Tehforsch: I stared at this comment for a while before I understood it, because I thought the second part was explaining the `16 byte` offset. 
Maybe we could rewrite this more explicitly like

```rust
let offset = 16; // maybe add some info for people unfamiliar with the protocol like me
let bytes_per_header_byte = 4; // 32 bit words / (8 bits per byte).
let hl = pkt.get_header_length() as usize * bytes_per_header_byte + offset;
```

Tehforsch: The `16` should be a named constant, since it appears many times

rust/src/alive_test/alive_test.rs
```diff
@@ -175,6 +108,66 @@ fn process_ip_packet(packet: &[u8]) -> Result<Option<AliveHostCtl>, AliveTestErr
...
+            return Ok(Some(AliveHostCtl {
+                ip: pkt.get_source().to_string(),
+                detection_method: AliveTestMethods::Icmp,
+            }));
+        } else if icmp_pkt.get_icmpv6_type() == Icmpv6Types::NeighborAdvert {
+            return Ok(Some(AliveHostCtl {
+                ip: pkt.get_source().to_string(),
+                detection_method: AliveTestMethods::Arp,
+            }));
+        }
```
Tehforsch: I don't quite understand the logic here in general - if the icmpv6 type is not `EchoReply` or `NeighbourAdvert`, we simply ignore the packet?

Also, this could maybe be rewritten by using a closure like

```rust
let make_alive_host_ctl = |pkt, method| {
    Ok(Some(AliveHostCtl {
        ip: pkt.get_source().to_string(),
        detection_method: method,
    }));
};

...

match icmp_pkt.get_icmpv6_type() {
    Icmpv6Types::EchoReply => return make_alive_host_ctl(pkt, AliveTestMethods::Icmp),
    Icmpv6Types::NeighbourAdvert => return make_alive_host_ctl(pkt, AliveTestMethods::Arp),
    _ => {}
}
```

rust/src/alive_test/alive_test.rs
```diff
@@ -3,117 +3,49 @@
...
-const ICMP_ECHO_REQ_CODE: u8 = 0;
-const IP_PPRTO_VERSION_IPV4: u8 = 4;
 
 pub struct AliveTestCtlStop;
 
+#[derive(Clone)]
 pub struct AliveHostCtl {
     ip: String,
     detection_method: AliveTestMethods,
 }
```
Tehforsch: Q: Why is this called `AliveHostCtl`?

rust/src/alive_test/alive_test.rs
```diff
@@ -227,33 +220,92 @@ async fn capture_task(
...
     if methods.contains(&AliveTestMethods::Icmp) {
-        for t in trgt.iter() {
+        for t in target.iter() {
             count += 1;
-            let dst_ip = match t.to_string().parse::<Ipv4Addr>() {
-                Ok(ip) => ip,
-                Err(_) => {
-                    continue;
+            match t
+                .to_string()
```
Tehforsch: Unnecessary `.to_string()`

rust/src/alive_test/alive_test.rs
```diff
@@ -227,33 +220,92 @@ async fn capture_task(
...
-            alive_test_send_icmp_packet(icmp)?;
         }
     }
     if methods.contains(&AliveTestMethods::TcpSyn) {
-        //unimplemented
+        for t in target.iter() {
+            for port in DEFAULT_PORT_LIST.iter() {
+                count += 1;
+                match t
+                    .to_string()
```
Tehforsch: same here

rust/src/alive_test/alive_test.rs
```diff
@@ -227,33 +220,92 @@ async fn capture_task(
...
+                IpAddr::V4(ipv4) => {
+                    forge_arp_request(ipv4)?;
+                }
+                IpAddr::V6(ipv6) => {
+                    let ndp = forge_neighbor_solicit(ipv6)?;
+                    alive_test_send_v6_packet(ndp)?;
+                }
+            };
+        }
     }
```
Tehforsch: It could help with code clarity if we do the target ip addr parsing only once, upfront:
```rust
    let target: HashSet<IpAddr> = target
        .into_iter()
        .map(|val| {
            val.parse::<IpAddr>()
                .map_err(|_| AliveTestError::InvalidDestinationAddr)
        })
        .collect::<Result<_, _>>()?;
    if methods.contains(&AliveTestMethods::Icmp) {
        for t in target.iter() {
            count += 1;
            match t {
                IpAddr::V4(ipv4) => {
                    let icmp = forge_icmp(*ipv4);
                    alive_test_send_v4_packet(icmp)?;
                }
                IpAddr::V6(ipv6) => {
                    let icmp = forge_icmp_v6(*ipv6)?;
                    alive_test_send_v6_packet(icmp)?;
                }
            };
        }
    }
    ... // rest of the methods
```

rust/src/alive_test/arp.rs
```diff
@@ -0,0 +1,35 @@
...
+
+use super::AliveTestError;
+use pcap::Capture;
+
+pub fn forge_arp_request(dst_ip: Ipv4Addr) -> Result<(), AliveTestError> {
+    let src_ip = get_source_ipv4(dst_ip).map_err(|_| AliveTestError::InvalidDestinationAddr)?;
+    let iface =
+        get_interface_by_local_ip(src_ip.into()).map_err(|_| AliveTestError::GetDeviceList)?;
+    let local_mac_address =
+        get_local_mac_address(&iface.name).map_err(|_| AliveTestError::GetMacAddress)?;
```
Tehforsch: All these error variants ignore the internal error. If it's not too much of a hassle, adding the internal error to the respective variants could help with clarity if this error occurs.

rust/src/alive_test/arp.rs
```diff
@@ -0,0 +1,35 @@
...
+
+    let arp_frame = forge_arp_frame(local_mac_address, src_ip, dst_ip);
+
+    if let Ok(capture) = Capture::from_device(iface.clone()) {
+        if let Ok(mut capture) = capture.timeout(DEFAULT_TIMEOUT).open() {
+            if let Ok(_) = capture.sendpacket(arp_frame) {
+                return Ok(());
+            }
+        };
+    };
```
Tehforsch: Also here, we're silently ignoring all the error variants, collapsing them all into the `SendArpRequest`. This might be a little too reductive and will be hard to debug if this ever occurs.

Tehforsch (changes requested):
Thanks! Looks good, but I had a few remarks here and there. 

Another nitpicky comment: There are still some cases where the `v4` case is the default (i.e. `forge_tcp_ping` and `forge_tcp_ping_v6`). Would prefer if the former was renamed to `forge_tcp_ping_v4`). Other examples of this are `EtherTypes` and `new_raw_socket`

rust/src/alive_test/alive_test.rs
jjnicola: Yes, we are only interested in echo reply or neighbor advertisement.

rust/src/alive_test/alive_test.rs
jjnicola: Changed to AliveHostInfo

Kraemii (approved):

PR #1965, opened 2025-08-05
Add: File type script preferences related functions and other nasl builtin

jjnicola:
**What**:
Add: File type script preferences related functions and other nasl builtin
SC-1352

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -3752,7 +3752,6 @@ dependencies = [
  "sha2",
  "socket2 0.6.0",
  "sysinfo",
- "tempfile",
  "thiserror 2.0.16",
  "time",
  "tokio",
--- a/rust/src/nasl/builtin/preferences/mod.rs
+++ b/rust/src/nasl/builtin/preferences/mod.rs
@@ -5,8 +5,6 @@ use crate::models::PreferenceValue;
 use crate::nasl::prelude::*;
 use crate::scanner::preferences::preference::PREFERENCES;
 use base64::Engine as _;
-use std::io::Write;
-use tempfile::NamedTempFile;
 
 fn script_get_preference_shared(
     register: &Register,
@@ -56,37 +54,6 @@ fn script_get_preference_file_content_shared(
     base64::engine::general_purpose::STANDARD
         .decode(content)
         .ok()
-}
-
-/// Given a preference name or ID of file type script preference, stores the
-/// preference value in a temporary file and returns the path.
-#[allow(unused)]
-pub fn get_plugin_preference_fname(
-    register: &Register,
-    config: &ScanCtx,
-    name: Option<String>,
-    id: Option<usize>,
-) -> Result<String, FnError> {
-    let mut tmp = NamedTempFile::with_prefix("openvas-file-upload.")
-        .map_err(|e| BuiltinError::Preference(e.to_string()))?;
-    match script_get_preference_file_content_shared(register, config, name, id) {
-        Some(file_content) => {
-            if tmp.write_all(&file_content).is_ok() {
-                Ok(tmp.path().to_string_lossy().into_owned())
-            } else {
-                Err(BuiltinError::Preference(format!(
-                    "get_plugin_preference_fname: Could not write to temporary file at {:?}",
-                    tmp.path()
-                ))
-                .into())
-            }
-        }
-        None => Err(BuiltinError::Preference(format!(
-            "get_plugin_preference_fname: Could not create temporary file for {:?}",
-            tmp.path()
-        ))
-        .into()),
-    }
 }
 
 #[nasl_function(named(id))]
--- a/rust/src/scanner/preferences/preference.rs
+++ b/rust/src/scanner/preferences/preference.rs
@@ -233,6 +233,13 @@ lazy_static! {
     pub static ref PREFERENCES_JSON: String = serde_json::to_string(&PREFERENCES).unwrap();
 }
 
+pub fn pref_is_true(prefs: &ScanPrefs, key: &str) -> Option<bool> {
+    prefs
+        .iter()
+        .find(|x| x.id == key)
+        .map(|x| matches!(x.value.as_str(), "true" | "1" | "yes"))
+}
+
 #[derive(Deserialize, Serialize, Debug, Clone, Eq, PartialEq)]
 #[serde(untagged)]
 pub enum ScanPrefValue {
@@ -385,10 +392,7 @@ impl ScanPrefs {
     }
 
     pub fn get_preference_bool(&self, key: &str) -> Option<bool> {
-        self.0
-            .iter()
-            .find(|x| x.id == key)
-            .map(|x| matches!(x.value.as_str(), "true" | "1" | "yes"))
+        pref_is_true(self, key)
     }
 
     pub fn iter(&self) -> impl Iterator<Item = &ScanPreference> {
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -107,7 +107,6 @@ rustls-platform-verifier = "0.6.1"
 rustls-webpki = "0.103.4"
 pathdiff = "0.2.3"
 rand_core = "0.6.4"
-tempfile = "3.20.0"
 
 [workspace]
 members = ["crates/smoketest", "crates/nasl-function-proc-macro"]
--- a/rust/src/nasl/utils/scan_ctx.rs
+++ b/rust/src/nasl/utils/scan_ctx.rs
@@ -11,7 +11,7 @@ use crate::models::{AliveTestMethods, Port, PortRange, Protocol, ScanPreference}
 use crate::nasl::builtin::{KBError, NaslSockets};
 use crate::nasl::syntax::Loader;
 use crate::nasl::{FromNaslValue, WithErrorInfo};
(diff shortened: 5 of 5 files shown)


rust/src/nasl/builtin/misc/mod.rs
```diff
@@ -281,6 +281,26 @@ fn vendor_version(context: &ScanCtx<'_>) -> Result<NaslValue, MiscError> {
...
+fn safe_checks(context: &ScanCtx<'_>) -> NaslValue {
+    NaslValue::Boolean(
+        context
+            .scan_preferences
+            .get_preference_bool("safe_checks")
+            .unwrap_or(false),
+    )
+}
+
+/// Are safe checks enabled ?
```
Tehforsch: Remove

rust/src/nasl/builtin/misc/mod.rs
```diff
@@ -281,6 +281,26 @@ fn vendor_version(context: &ScanCtx<'_>) -> Result<NaslValue, MiscError> {
...
+#[nasl_function]
+fn get_script_oid(context: &ScanCtx<'_>) -> NaslValue {
+    match context.nvt().clone() {
+        Some(nvt) => NaslValue::Data(nvt.oid.into_bytes()),
+        None => NaslValue::Null,
+    }
```
Tehforsch: If you return `Option<Vec<u8>>` from the function, you can get rid of the `NaslValue::Data` and write `context.nvt().map(|nvt| nvt.oid.clone().into_bytes())`. This should avoid having to clone the entire `Nvt`.

rust/src/nasl/builtin/misc/mod.rs
```diff
@@ -281,6 +281,26 @@ fn vendor_version(context: &ScanCtx<'_>) -> Result<NaslValue, MiscError> {
...
+/// Are safe checks enabled ?
+#[nasl_function]
+fn safe_checks(context: &ScanCtx<'_>) -> NaslValue {
+    NaslValue::Boolean(
+        context
+            .scan_preferences
+            .get_preference_bool("safe_checks")
+            .unwrap_or(false),
+    )
+}
```
Tehforsch: You can just return a `bool` and get rid of the `NaslValue::Boolean`

rust/src/nasl/builtin/preferences/mod.rs
```diff
@@ -42,6 +46,89 @@ fn script_get_preference(
...
+    name: Option<String>,
+    id: Option<usize>,
+) -> Option<NaslValue> {
+    let content = script_get_preference_shared(register, config, name, id)?;
+    let content = content.as_string().unwrap();
+    match base64::engine::general_purpose::STANDARD.decode(content) {
+        Ok(c) => Some(c.into()),
+        Err(_) => None,
+    }
+}
```
Tehforsch: Changing the return type to `Option<Vec<u8>>` simplifies this a bit:

1. You can rewrite the last expression as `base64::engine::general_purpose::STANDARD.decode(content).ok()`
2. You dont have to do any conversion in `get_plugin_preference_fname`
3. You can return a `Vec<u8>` in `script_get_preference_file_content`.

rust/src/nasl/builtin/preferences/mod.rs
```diff
@@ -42,6 +46,89 @@ fn script_get_preference(
...
+// The C implementation of this function, is internally called by nasl_builtin_find_service to
+// store the TLS stuff in temporary files, which are later use to create a tls socket.
+#[allow(unused)]
+pub fn get_plugin_preference_fname(
+    register: &Register,
+    config: &ScanCtx,
+    name: Option<String>,
+    id: Option<usize>,
+) -> Result<String, FnError> {
+    let mut tmp = match NamedTempFile::with_prefix("openvas-file-upload.") {
```
Tehforsch: Is the trailing `.`  intentional? If so, ignore this comment.

rust/src/nasl/builtin/preferences/mod.rs
```diff
@@ -42,6 +46,89 @@ fn script_get_preference(
...
+        Ok(f) => f,
+        Err(e) => return Err(BuiltinError::Preference(e.to_string()).into()),
+    };
```
Tehforsch:
```rust
NamedTempFile::with_prefix("openvas-file-upload").map_err(|e| BuiltinError::Preference(e.to_string()))?
```

rust/src/nasl/builtin/preferences/mod.rs
```diff
@@ -42,6 +46,89 @@ fn script_get_preference(
...
+        && tmp.write_all(file_content.to_string().as_bytes()).is_ok()
+    {
+        return Ok(tmp.path().to_string_lossy().into_owned());
+    }
+
+    Err(BuiltinError::Preference(format!(
+        "get_plugin_preference_fname: Could not create temporary file for {:?}",
+        tmp.path()
+    ))
+    .into())
```
Tehforsch: Super nitpicky, but I don't like the early return style here. It makes the good logical flow (returning the path) seem like the exception. I would prefer something like this (which also shows slightly different error messages depending on what happened)

```rust
    match script_get_preference_file_content_shared(register, config, name, id) {
        Some(file_content) => {
            if tmp.write_all(&file_content).is_ok() {
                Ok(tmp.path().to_string_lossy().into_owned())
            } else {
                Err(BuiltinError::Preference(format!(
                    "get_plugin_preference_fname: Could not write to temporary file at {:?}",
                    tmp.path()
                ))
                .into())
            }
        }
        None => Err(BuiltinError::Preference(format!(
            "get_plugin_preference_fname: Could not create temporary file for {:?}",
            tmp.path()
        ))
        .into()),
    }
```

Also, you could change the return type of the function to be `BuiltinError`, which allows you to get rid of the `.into()` calls.

rust/src/nasl/builtin/preferences/mod.rs
```diff
@@ -42,6 +46,89 @@ fn script_get_preference(
...
+        Ok(c) => Some(c.into()),
+        Err(_) => None,
+    }
+}
+
+/// Given a preference name or ID of file type script preference, stores the
+/// preference value in a temporary file and returns the path.
+// The C implementation of this function, is internally called by nasl_builtin_find_service to
+// store the TLS stuff in temporary files, which are later use to create a tls socket.
+#[allow(unused)]
```
Tehforsch: Why should this be here?

rust/src/nasl/builtin/preferences/mod.rs
```diff
@@ -42,6 +46,89 @@ fn script_get_preference(
...
+    }
+
+    Err(BuiltinError::Preference(format!(
+        "get_plugin_preference_fname: Could not create temporary file for {:?}",
+        tmp.path()
+    ))
+    .into())
+}
+
+#[allow(unused)]
```
Tehforsch: Also unused?

rust/src/scanner/preferences/preference.rs
```diff
@@ -386,6 +386,13 @@ impl ScanPrefs {
         self.0.iter().find(|x| x.id == key).map(|x| x.value.clone())
     }
 
+    pub fn get_preference_bool(&self, key: &str) -> Option<bool> {
+        self.0
+            .iter()
+            .find(|x| x.id == key)
+            .map(|x| matches!(x.value.as_str(), "true" | "1" | "yes"))
```
Tehforsch: This logic exists multiple times throughout the code (in osp and in `scan_ctx`). We could maybe think about using a single function to unify all these.

Tehforsch (changes requested):

rust/Cargo.lock
Tehforsch, line 1: Resetting this caused some issues with the criterion deprecation of `black_box`.

rust/src/nasl/builtin/preferences/mod.rs
jjnicola: Yes, the point is part of the prefix.

rust/src/nasl/builtin/preferences/mod.rs
jjnicola: Deleted.

rust/src/nasl/builtin/preferences/mod.rs
jjnicola: For now, unused. Later, should be used for storing credentials. I didn't know where to leave this comments. All the certificates stuff via script preferences for authentication is a bit confusing. To be honest, the comment above and this functions are like a documentation for further development. Deleted.

rust/src/scanner/preferences/preference.rs
jjnicola: ~~I would prefer to do it in separated PR.~~
Done within https://github.com/greenbone/openvas-scanner/pull/1965/commits/39ba053ed789f047923034aa411abdc815fb8a3a

Only found in `preferences.rs` and `scan_ctx`

rust/Cargo.lock
jjnicola: You have already fixed it #1975

rust/src/nasl/builtin/preferences/mod.rs
Tehforsch: It's still here?

PR #1967, opened 2025-08-11
Nasl ntlm crypto

Kraemii:
SC-1349

### DIFF
--- a/rust/src/nasl/builtin/cryptographic/ntlm.rs
+++ b/rust/src/nasl/builtin/cryptographic/ntlm.rs
@@ -416,6 +416,58 @@ fn nt_owf_gen(pass: StringOrData) -> Vec<u8> {
     hasher.finalize().to_vec()
 }
 
+#[nasl_function(named(owf, login, domain))]
+fn ntv2_owf_gen(
+    owf: StringOrData,
+    login: StringOrData,
+    domain: StringOrData,
+) -> Result<Vec<u8>, FnError> {
+    use hmac::{Hmac, Mac};
+    use md5::Md5;
+
+    if login.0.is_empty() {
+        return Err(FnError::wrong_unnamed_argument(
+            "login of length >= 1",
+            &format!("login of length {}", login.0.len()),
+        ));
+    }
+
+    if domain.0.is_empty() {
+        return Err(FnError::wrong_unnamed_argument(
+            "domain of length >= 1",
+            &format!("domain of length {}", domain.0.len()),
+        ));
+    }
+
+    if owf.0.len() != 16 {
+        return Err(FnError::wrong_unnamed_argument(
+            "owf of length 16",
+            &format!("owf of length {}", owf.0.len()),
+        ));
+    }
+
+    let user: Vec<u8> = login
+        .0
+        .to_uppercase()
+        .as_bytes()
+        .iter()
+        .flat_map(|b| [*b, 0u8])
+        .collect();
+    let domain: Vec<u8> = domain
+        .0
+        .to_uppercase()
+        .as_bytes()
+        .iter()
+        .flat_map(|b| [*b, 0u8])
+        .collect();
+
+    // We can unwrap here, as the key is always 16 bytes.
+    let mut hmac = Hmac::<Md5>::new_from_slice(owf.0.as_bytes()).unwrap();
+    hmac.update(&user);
+    hmac.update(&domain);
+    Ok(hmac.finalize().into_bytes().to_vec())
+}
+
 pub struct Ntlm;
 
 function_set! {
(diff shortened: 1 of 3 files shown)


rust/Cargo.toml
```diff
@@ -45,7 +46,11 @@ md4 = "0.10.2"
 num_cpus = "1.16.0"
 pbkdf2 = { version = "0.12.2", features = ["password-hash"] }
 pkcs8 = { version = "0.10.2", features = ["encryption", "pem", "std"] }
-quick-xml = { version = "0.37.1", features = ["serde", "serde-types", "serialize"] }
+quick-xml = { version = "0.37.1", features = [
+    "serde",
+    "serde-types",
+    "serialize",
+] }
```
Tehforsch: `.toml` autoformatting?

rust/src/nasl/builtin/cryptographic/ntlm.rs
```diff
@@ -0,0 +1,425 @@
...
+    let cryptkey = cryptkey.0.as_bytes();
+    let passhash = passhash.0.as_bytes();
+
+    ep24(cryptkey, passhash)
+}
+
+/// Unfortunately as this function is based on random generated data, it is not easy to test,
+/// but was manually verified by comparing it with an altered function with fixed data.
+#[nasl_function(named(cryptkey, passhash, length))]
+fn ntlmv2_hash(
```
Tehforsch: Hm, what about extracting the body of the function into a non-NASL function that takes a RNG as an argument, so it becomes a pure function? This would make it easy to test, by initializing it with a fixed seed in tests.

rust/src/nasl/builtin/cryptographic/ntlm.rs
```diff
@@ -0,0 +1,425 @@
...
+
+    lm_response.append(&mut nt_response);
+    lm_response.append(&mut session_key);
+
+    Ok(lm_response)
+}
+
+/// This function is broken in openvas-nasl and is currently not used in the feed.
+/// Also it relies on random generated data, so it is impossible to test, if it
+/// works correctly
```
Tehforsch: Same comment as above, testing should be quite easy if we take RNG as an argument.

rust/src/nasl/builtin/cryptographic/ntlm.rs
```diff
@@ -0,0 +1,425 @@
...
+    cryptkey: StringOrData,
+    user: StringOrData,
+    domain: StringOrData,
+    ntlmv2_hash: StringOrData,
+    address_list: StringOrData,
+    address_list_len: usize,
+) -> Result<Vec<u8>, FnError> {
+    let cryptkey = cryptkey.0.as_bytes();
+    let _ = user;
+    let _ = domain;
```
Tehforsch: Just for my understanding: Is this because the function historically takes user and domain as argument but doesn't actually need them?

rust/src/nasl/builtin/cryptographic/ntlm.rs
```diff
@@ -0,0 +1,425 @@
...
+    let ntlmv2_hash = ntlmv2_hash.0.as_bytes();
+    let address_list = address_list.0.as_bytes();
+    let _ = address_list_len;
```
Tehforsch: Same here

rust/src/nasl/builtin/cryptographic/ntlm.rs
```diff
@@ -0,0 +1,425 @@
...
+    Ok(lm_response)
+}
+
+#[nasl_function(named(cryptkey, session_key, nt_hash))]
+fn key_exchange(
+    cryptkey: StringOrData,
+    session_key: StringOrData,
+    nt_hash: StringOrData,
+) -> Result<Vec<u8>, FnError> {
+    let _ = cryptkey;
```
Tehforsch: Same question as above here. 

Nitpick: I think if there is a lot of this, we could also use attributes in argument position i.e.
```rust
fn key_exchange(
    #[allow(unused)]
    cryptkey: StringOrData,
    session_key: StringOrData,
    #[allow(unused)]
    nt_hash: StringOrData,
) -> Result<Vec<u8>, FnError> { 
```

which I think communicates the intent a little bit more clearly.

rust/src/nasl/builtin/cryptographic/ntlm.rs
Kraemii: They weren't used in the c nasl code as well, so I just let them in for compatibility reasons

rust/src/nasl/builtin/cryptographic/ntlm.rs
Kraemii: Same here

PR #1983, opened 2025-09-05
Fix: use snake case for object under tag key in vts

jjnicola:
**What**:
Fix: use snake case for object under tag key in vts
SC-1363

**Why**:

**How**:

**Checklist**:

- [ ] Tests


Kraemii:
This PR has no changes anymore

PR #1914, opened 2025-05-20
Add: AppStream module support

n-thumann:
**What**:
This PR adds a partial support for AppStream modules to the RPM package type.

**TODO:**
This PR only contains the adjustments in `rpm.rs`. The part to load the adjusted Notus JSON (see linked ticket) is still missing (lack of Rust skills + I was unsure how you'd prefer to integrate it into your models). Can you take over this PR and implement it?

**Why**:
This is being done to correctly compare RPM packages of different AppStream modules.

Jira: SC-1299.

**How**:
Compiled openvasd and tested it with some custom Notus files to confirm that my overall idea is at least working.

**Checklist**:

- [x] Tests
- [x] PR merge commit message adjusted

PR #1986, opened 2025-09-11
Add: support for handling AppStream modules for RPM (notus)

Kraemii:
This adds support for AppStream modules for RPM packages, so that packages from differen modules are not comparable. This missing lead to a lot of fals positives. A new format for the Products notus file is established, which now allows an Optional field for defining the AppStream module.

Jira: SC-1299

closes https://github.com/greenbone/openvas-scanner/pull/1914

### DIFF
--- a/rust/crates/greenbone-scanner-framework/src/models/product.rs
+++ b/rust/crates/greenbone-scanner-framework/src/models/product.rs
@@ -51,16 +51,28 @@ pub enum FixedPackage {
     ByFullName {
         full_name: String,
         specifier: Specifier,
+        module: Option<Module>,
     },
     /// Contains a version and a specifier
     ByNameAndFullVersion {
         name: String,
         full_version: String,
         specifier: Specifier,
+        module: Option<Module>,
     },
 
     /// Contains a version Range
-    ByRange { name: String, range: Range },
+    ByRange {
+        name: String,
+        range: Range,
+        module: Option<Module>,
+    },
+}
+
+#[derive(Deserialize, Debug)]
+pub struct Module {
+    pub name: String,
+    pub stream: String,
 }
 
 /// A specifier can be one of: >, <, >=, <=, =
--- a/rust/src/notus/packages/rpm.rs
+++ b/rust/src/notus/packages/rpm.rs
@@ -20,16 +20,22 @@ static RE: Lazy<Regex> = lazy_regex!(
         :
     )?
     (?P<version>
-        [^-]+
+        [^-@]+
     )
     -
     (?P<release>
-        [^-]+
+        [^-@]+
     )
     \.
     (?P<arch>
-        [^-]+
-    )$"
+        [^-@]+
+    )
+    (
+        @
+        (?P<module_name>[^:]+)
+        :
+        (?P<module_stream>.+)
+    )?$"
 );
 /// Used for parsing the full version of a package
 static RE_VERSION: Lazy<Regex> = lazy_regex!(
@@ -49,8 +55,14 @@ static RE_VERSION: Lazy<Regex> = lazy_regex!(
     )
     \.
     (?P<arch>
-        [^-]+
-    )$"
+        [^-@]+
+    )
+    (
+        @
+        (?P<module_name>[^:]+)
+        :
+        (?P<module_stream>.+)
+    )?$"
 );
 
 /// Represent a based Redhat package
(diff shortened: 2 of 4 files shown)


rust/src/notus/packages/rpm.rs
```diff
@@ -486,10 +571,52 @@ mod rpm_tests {
...
+            epoch: 0,
+            version: PackageVersion("1.2.3".to_string()),
+            release: PackageVersion("4".to_string()),
+            arch: "x86_64".to_string(),
+            module: ("bar".to_string(), "1338".to_string()),
+        };
+        assert!(package2.partial_cmp(&package1).is_none());
+        assert!(package1.partial_cmp(&package2).is_none());
+    }
+
```
n-thumann, line 631: I just tested and noticed this scan result:
```
Vulnerable package:   ruby
Installed version:    ruby-2.7.6-138.module_el8.6.0+3263+904da987.x86_64@ruby:2.7
Fixed version:      >=ruby-2.7.8-139.module_el8.8.0+3578+2b4b06da.aarch64@ruby:2.7
```

I don't think they should be comparable, because they're still different architectures (the test below is failing)
```suggestion
    
    #[test]
    pub fn test_compare_same_module_different_arch() {
        let package1 =
            Rpm::from_full_name("ruby-2.7.6-138.module_el8.6.0+3263+904da987.x86_64@ruby:2.7").unwrap();
        let package2 =
            Rpm::from_full_name("ruby-2.7.8-139.module_el8.8.0+3578+2b4b06da.aarch64@ruby:2.7").unwrap();
        assert!(package2.partial_cmp(&package1).is_none());
        assert!(package1.partial_cmp(&package2).is_none());
    }
```

rust/src/models/product.rs
```diff
@@ -49,16 +49,28 @@ pub enum FixedPackage {
...
+        name: String,
+        range: Range,
+        app_stream: Option<AppStream>,
+    },
+}
+
+#[derive(Deserialize, Debug)]
+pub struct AppStream {
+    pub name: String,
+    pub version: String,
```
n-thumann:
```suggestion
    pub stream: String,
```
Can we call this `stream`? The reason for that is that an AppStream can also have a version, next to the stream:
```
  ~ docker run --rm -it almalinux:8 dnf -q module info ruby:2.7 | head -n 5 
Name             : ruby
Stream           : 2.7
Version          : 8080020230630114654
Context          : 9edba152
Architecture     : x86_64
```
```
➜  ~ curl -s https://errata.almalinux.org/8/errata.full.json | jq ".data[0].modules[0]"
{
  "name": "mariadb",
  "arch": "i686",
  "stream": "10.5",
  "version": "8060020220803092659",
  "context": "9edba152"
}
```

rust/src/notus/packages/rpm.rs
Kraemii: Really good catch. As the regex does a greedy approach, adding the `@` symbol at the end and don't allow it in the `arch` capture lead to a match looking like:
```
Rpm {
    name: "ruby",
    epoch: 0,
    version: PackageVersion(
        "2.7.6",
    ),
    release: PackageVersion(
        "138.module_el8.6.0+3263+904da987.x86_64@ruby:2",
    ),
    arch: "7",
    module: (
        "",
        "",
    ),
}
```

Allowing the `@` only as the separator to the AppStream Module fixes this.

rust/src/models/product.rs
Kraemii: Right somehow I took the wrong name for this field

Kraemii:
Thanks for the feedback!

rust/crates/greenbone-scanner-framework/src/models/product.rs
```diff
@@ -49,16 +49,28 @@ pub enum FixedPackage {
     ByFullName {
         full_name: String,
         specifier: Specifier,
+        app_stream: Option<AppStream>,
```
n-thumann:
```suggestion
        module: Option<Module>,
```
Any objection to call this `module` or `app_stream_module`? I'm asking because "AppStream" is only the RPM repository, but we're comparing modules (which are part of the AppStream repo).

n-thumann (approved):
Tested various cases (AlmaLinux with different modules and different streams together with the other required changes on the NASL and Notus side) and can confirm that this works as expected! Thanks 🤝

(Note: I used a temporary branch with this and https://github.com/greenbone/openvas-scanner/pull/2000 for testing)

PR #1988, opened 2025-09-15
Nasl big numbers

Kraemii:
Adds function bn_random, which generates a random number of any lenght needed in bits, but rounded up to whole bytes. So `bn_random(need: 1)` will produce the same numbers as `bn_random(need: 8)`

Adds function bn_cmp to compare two big numbers

Both functions are covered by tests and also produce the same result as the c implementation.

Here is a small script, to test functionality is same as in the c implementation:
```c#
k1 = bn_random(need: 1);
k2 = bn_random(need: 8);

display("k1: " + hexstr(k1));
display("k2: " + hexstr(k2));

cmp = bn_cmp(key1: k1, key2: k2);

if (cmp < 0) {
  display("k1 > k2");
} else if (cmp == 0) {
  display("k1 == k2");
} else
{
  display("k1 < k2");
}
```

### DIFF
--- a/doc/manual/nasl/built-in-functions/cryptographic/bn_cmp.md
+++ b/doc/manual/nasl/built-in-functions/cryptographic/bn_cmp.md
@@ -12,7 +12,7 @@
 
 ## DESCRIPTION
 
-Compares two big numbers given as Bytes.
+Compares two big numbers given as Bytes interpreted in big endian.
 
 ## RETURN VALUE
 
--- a/rust/src/nasl/builtin/cryptographic/misc.rs
+++ b/rust/src/nasl/builtin/cryptographic/misc.rs
@@ -51,14 +51,14 @@ fn bn_cmp(key1: StringOrData, key2: StringOrData) -> i64 {
 /// A big number is an integer, that is probably to big for any primitive data type. In case of the
 /// c implementation the mpi (multi-precision integer) type of libgcrypt was used.
 #[nasl_function(named(need))]
-fn bn_random(need: u64) -> Option<Vec<u8>> {
+fn bn_random(need: u64) -> Vec<u8> {
     let digits = need / 8;
     let rem = need % 8;
 
     let digits = if rem > 0 { digits + 1 } else { digits };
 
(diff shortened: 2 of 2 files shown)


rust/src/nasl/builtin/cryptographic/misc.rs
```diff
@@ -27,11 +30,44 @@ fn insert_hexzeros(register: &Register) -> Result<Vec<u8>, FnError> {
...
+
+/// This function generates a big number (bn) with the given amount of bits.
+/// As this function generates the number as bytes, it actually takes the number of bits, divides them by 8
+/// and rounds them up. It does not only generate the desired number of bits for the last byte, but a whole
+/// random byte.
+///
+/// A big number is an integer, that is probably to big for any primitive data type. In case of the
+/// c implementation the mpi (multi-precision integer) type of libgcrypt was used.
+#[nasl_function(named(need))]
+fn bn_random(need: u64) -> Option<Vec<u8>> {
```
Tehforsch: Couldn't this function simply return a `Vec<u8>`?

doc/manual/nasl/built-in-functions/cryptographic/bn_cmp.md
```diff
@@ -2,21 +2,21 @@
...
-*str* **bn_cmp**(key: str, buf: str);
+*str* **bn_cmp**(key1: str, key2: str);
 
-**bn_cmp** It takes two named arguments key1, key2. Returns 0 when they're equal, -1 when key1 > key2, +1 when key2 > key1.
+**bn_cmp** It takes two named arguments
 
 ## DESCRIPTION
 
-Compares key1 with key2.
+Compares two big numbers given as Bytes.
```
Tehforsch: I think as a consumer of this function you need to know that the numbers are going to be interpreted as big endian, so I'd add this to the documentation.

Tehforsch (commented):
Thanks! Also thanks for improving the docs

Tehforsch (approved):

PR #1989, opened 2025-09-16
Add fork kinds.

Tehforsch:
This allows the forking NASL function to specify what the "origin" of the fork is (i.e. VHosts). An interpreter that runs into a fork will check whether it has already forked previously for that fork kind and return the value it was spawned with if it was. Otherwise it will proceed with the fork as before.

Currently, only the VHost related functions have a fork kind, but others can easily be added in the future.

Jira: SC-1364


Kraemii (approved):

jjnicola (commented):
It seems that the interpreter is broken. The bug is not from this PR.
Feel free to fix now or leave it for separated PR.

This will work
```
set_kb_item(name: "port", value: 1);
set_kb_item(name: "port", value: 2);
set_kb_item(name: "host", value: "a");
set_kb_item(name: "host", value: "b");
a= get_kb_item("port");
b=get_kb_item("host");
display(a);
display(b);
```

while this won't work:
```
set_kb_item(name: "port", value: 1);
set_kb_item(name: "port", value: 2);
set_kb_item(name: "host", value: "a");
set_kb_item(name: "host", value: "b");
display(get_kb_item("port"));
display(get_kb_item("host"));
```

PR #1993, opened 2025-09-24
Fix bug with local function arguments.

Tehforsch:
Previously, nasl_functions would look up the value of a function argument via `Register::nasl_value` which starts the lookup in the local scope, but goes all the way to the global scope. Omitting a named argument to a function could result in wrong values being taken out of the global context, i.e. in a set up like this:

```rust
fn foo(x: usize) -> x {
    x
}
```

```nasl
x = 3;
display(foo());
```

where the latter script would result in "3".

### DIFF
--- a/rust/src/feed/update/mod.rs
+++ b/rust/src/feed/update/mod.rs
@@ -82,7 +82,7 @@ pub async fn feed_version(
 
     let feed_version = interpreter
         .register()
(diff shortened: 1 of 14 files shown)


jjnicola (approved):

jjnicola:
Thanks! It worked like a charm!

PR #1995, opened 2025-09-24
Fix: on range=1 without `-` it means from and not to this value

nichtsfrei:
Fixes the logic bug on GET /scans/id/results?range=0

SC-1374


rust/crates/greenbone-scanner-framework/src/get_scans_id_results.rs
```diff
@@ -174,8 +170,13 @@ mod tests {
                     io::Error::other("oh no"),
                 )))]));
             }
+            let to = if from.is_some() {
+                // for the test case ?range=99
+                to.unwrap_or(101)
+            } else {
+                to.unwrap_or_default()
+            };
```
Tehforsch, line 178: Why does the test case get special treatment?

nichtsfrei: Because otherwise it won't have any data for testing: 99..0 is none.

nichtsfrei: soo is there anything you want me to do or is that question resolved?

Tehforsch: i dont really understand the answer but its fine

PR #1996, opened 2025-09-24
Fix: check the nasl file encoding before converting results to utf-8

jjnicola:
**What**:
check if the original nasl file is encoded in UTF-8 or not
SC-737

**Why**:
We convert everything to UTF-8 before sending the results to client.
To avoid double conversion, we check now that nasl files are in UTF-8 encoding.

**How**:

**Checklist**:

- [ ] Tests


cfi-gb:
Only for reference, follow-up fix in #2093

PR #2005, opened 2025-09-29
Add: extend nasl builtin find service.

jjnicola:
**What**:
Try with a tls socket within nasl builtin find service 
Jira: SC-1378

**Why**:

**How**:

**Checklist**:

- [ ] Tests


rust/src/nasl/builtin/find_service/mod.rs
```diff
@@ -361,12 +417,60 @@ fn scan_port(
...
+    let mut cert = match get_plugin_preference_fname(register, context, None, Some(CERT_FILE)) {
+        Ok(cert) => cert,
+        _ => "".to_string(),
+    };
+    let mut key = match get_plugin_preference_fname(register, context, None, Some(KEY_FILE)) {
+        Ok(key) => key,
+        _ => "".to_string(),
+    };
+
+    dbg!(&cert);
```
Tehforsch: remove?

rust/src/nasl/builtin/find_service/mod.rs
```diff
@@ -361,12 +417,60 @@ fn scan_port(
...
+        if key.is_empty() {
+            key = cert.clone();
+        };
+        if cert.is_empty() {
+            cert = key.clone();
+        };
+
+        plug_set_ssl_cert(context, cert)?;
+        plug_set_ssl_key(context, key)?;
+    };
```
Tehforsch: with this logic here anyways, wouldnt it be clearer to leave both as an option (instead of initializing as "") and then to do 

```
plug_set_ssl_cert(context, cert.unwrap_or(key))?;
plug_set_ssl_key(context, key.unwrap_or(cert))?;

```

rust/src/nasl/builtin/find_service/mod.rs
```diff
@@ -361,12 +417,60 @@ fn scan_port(
...
+
+    if let Ok(pem_pass) = get_plugin_preference_fname(register, context, None, Some(PEM_PASS)) {
+        plug_set_ssl_password(context, pem_pass)?;
+    };
+
+    if let Ok(ca) = get_plugin_preference_fname(register, context, None, Some(CA_FILE)) {
+        plug_set_ssl_ca_file(context, ca)?;
+    };
```
Tehforsch, line 441: This will silently ignore all errors, is this what we want? We could also add a log of some kind.

rust/src/nasl/builtin/find_service/mod.rs
jjnicola: ups!

rust/src/nasl/builtin/find_service/mod.rs
jjnicola: We don't care if the preferences `PEM_PASS`  and `CA_FILE` are not set. Therefore we are ignoring the error in this case.

Tehforsch (approved):
Thanks!

PR #2009, opened 2025-09-30
Add dockerfile for openvas/ospd-openvas testing.

Tehforsch:
Test via
```bash
cd openvasd-tests
make openvas # for openvas backend
make ospd # for ospd backend

make discovery # run discovery script
```

### DIFF
--- a/rust/openvasd-tests/Dockerfile
+++ b/rust/openvasd-tests/Dockerfile
@@ -2,51 +2,16 @@
 
 FROM registry.community.greenbone.net/community/gvm-libs:edge
 
-RUN apt-get update && apt-get install --no-install-recommends --no-install-suggests -y \
-    redis-server \
-    redis-tools \
-    curl \
-    patchelf \
-    gettext-base \
-    build-essential \
-    cmake \
-    make \
-    gcc \
-    g++ \
-    pkg-config \
-    bison \
-    libglib2.0-dev \
-    libgpgme-dev \
-    libgnutls28-dev \
-    libgcrypt20-dev \
-    libpcap-dev \
-    libksba-dev \
-    libmagic-dev \
-    file \
-    libssh-dev \
-    libhiredis-dev \
-    libxml2-dev \
-    libnet1-dev \
-    libpaho-mqtt-dev \
-    libjson-glib-dev \
-    libcurl4-gnutls-dev \
-    libbsd-dev \
-    uuid-dev \
-    libldap2-dev \
-    libradcli-dev \
-    gcc-mingw-w64 \
-    rsync \
-    nmap \
-    libsnmp-dev \
-    python3 \
-    python3-pip \
-    python3-impacket \
-    mosquitto \
-    && rm -rf /var/lib/apt/lists/*
-
 # Build from src
 COPY . /opt/openvas-build/
 WORKDIR /opt/openvas-build
+
+RUN .github/install-openvas-dependencies.sh
+
+RUN apt-get update && apt-get install --no-install-recommends --no-install-suggests -y \
+    redis-server \
+    redis-tools \
+    patchelf
 
 RUN mkdir -p build && \
     cd build && \
@@ -57,9 +22,10 @@ RUN mkdir -p build && \
 COPY rust/openvasd-tests/entrypoint.sh /entrypoint.sh
 
 COPY .docker/openvas.conf /etc/openvas/
-COPY rust/openvasd-tests/openvasd.toml.template /etc/openvas/openvasd.toml.template
 
 RUN chmod +x /entrypoint.sh
+
+RUN sed 's/redis-openvas/redis/g' /opt/openvas-build/config/redis-openvas.conf > /etc/redis/redis.conf
 
 EXPOSE 3000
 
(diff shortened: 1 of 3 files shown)


rust/openvasd-tests/Makefile
```diff
@@ -0,0 +1,56 @@
...
+	@docker rm -f $(CONTAINER_NAME) 2>/dev/null || true
+	docker run -d \
+		--name $(CONTAINER_NAME) \
+		-p $(PORT):3000 \
+		-e SCANNER_TYPE=$(SCANNER_TYPE) \
+		-e API_KEY=$(API_KEY) \
+		-v "$(realpath ../target):/opt/openvasd_target" \
+		$(DOCKER_IMAGE)
+	@echo "Container started. Waiting for service to be ready..."
+	@sleep 5
```
nichtsfrei: wouldn't a loop be better than just waiting for 5 seconds?

nichtsfrei: Also there is a status command, why don't you reuse that?

rust/openvasd-tests/Makefile
```diff
@@ -0,0 +1,56 @@
...
+	@echo "Building Docker image..."
+	docker build -t $(DOCKER_IMAGE) -f Dockerfile ../..
+	@echo "Build complete!"
+
+start-scanner: build
+	@echo "Starting openvasd with $(SCANNER_TYPE) backend..."
+	@docker rm -f $(CONTAINER_NAME) 2>/dev/null || true
+	docker run -d \
+		--name $(CONTAINER_NAME) \
+		-p $(PORT):3000 \
```
nichtsfrei, line 24: when you use `--rm` the container would be automatically deleted when stopped.

rust/openvasd-tests/entrypoint.sh
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env bash
+set -e
+
+OPENVASD_PATH=/opt/openvasd_target/debug/openvasd
```
nichtsfrei: I would recommend to use release not debug?

rust/openvasd-tests/entrypoint.sh
```diff
@@ -0,0 +1,42 @@
...
+
+redis-server --daemonize yes --bind 127.0.0.1 --port 6379
```
nichtsfrei: I would recommend to use the configuration within `config/redis-openvas.conf` and sed to follow the defaults, see:
https://github.com/greenbone/openvas-scanner/blob/main/.devcontainer/Dockerfile#L51
as an example. 

That way we wouldn't need an additional open port.

rust/openvasd-tests/entrypoint.sh
```diff
@@ -0,0 +1,42 @@
...
+
+until redis-cli ping > /dev/null 2>&1; do
```
nichtsfrei: if you change to unix socket the command would change to redis-cli -S /run/redis/redis.sock ping, just a note if you change that.

rust/openvasd-tests/entrypoint.sh
```diff
@@ -0,0 +1,42 @@
...
+    echo "Waiting for Redis..."
+    sleep 1
+done
+
+export API_KEY=${API_KEY:-changeme}
+
+if [ -f  $OPENVASD_PATH ]; then
+    patchelf --set-interpreter /lib64/ld-linux-x86-64.so.2 $OPENVASD_PATH 2>/dev/null || true
```
nichtsfrei, line 21: Please add a comment why this is necessary.

rust/openvasd-tests/Dockerfile
```diff
@@ -0,0 +1,66 @@
+# to be built with openvas-scanner top level dir as base
+
+FROM registry.community.greenbone.net/community/gvm-libs:edge
+
+RUN apt-get update && apt-get install --no-install-recommends --no-install-suggests -y \
```
nichtsfrei: is there a reason for not using install-openvas-dependencies?

https://github.com/greenbone/openvas-scanner/blob/main/.devcontainer/Dockerfile#L40

rust/openvasd-tests/Makefile
Tehforsch: absolutely

rust/openvasd-tests/Makefile
Tehforsch: good to know, thanks! we still have to keep the `docker rm` around though, since openvasd will be running "infiinitely"

rust/openvasd-tests/entrypoint.sh
Tehforsch: i intentionally used release, since this would be mostly used for debugging, but i can default to release and keep an option for debug around, if you prefer

rust/openvasd-tests/Dockerfile
Tehforsch: ignorance. fixed.

rust/openvasd-tests/entrypoint.sh
nichtsfrei: considering that the behavior of openvasd is very different and we have run into issues that turned out to be false on release mode I think release would be safer to use release as a default?

Maybe we can think about adding debug symbols to release though?

nichtsfrei (dismissed):

PR #2013, opened 2025-10-02
Fix: multiple scan starts

nichtsfrei:
This fixes a bug where a scan could be started again even though it was already started.

It is not enough to only change the status from `requested` to `running` to stop this from happening.

The cause is SQLite default `DEFERRED` transactions. If the first statement is a `SELECT`, the transaction is marked as read and only later tries to switch to write. If one transaction commits a write and a later transaction already started, it may still use the old snapshot without the new `running` state and will try again to move the scan from `requested` to `running`.

SQLite has ways to avoid this, for example using `IMMEDIATE` to start a write transaction at once. But with `sqlx` and a pool this is not simple because the pool may hold or open transactions in the background.

As a workaround I added a manual check. We now keep a vector of scans that were already set to `running`. Only if the scan is not in that list we call `scanner.start_scan`. This prevents starting the same scan more than once.

I also changed the tick logic from burst to delay. This makes sure that `on_schedule` is not called right away when a tick was missed.


rust/src/openvasd/scans/scheduling.rs
```diff
@@ -353,6 +411,7 @@ where
         tracing::warn!(%error, "Unable to set not stopped runs from a previous session to failed.")
     }
     let mut interval = tokio::time::interval(check_interval);
+    interval.set_missed_tick_behavior(MissedTickBehavior::Delay);
```
Tehforsch, line 416: If I understand this correctly, the reason for this is that is (presumably/hopefully) reduces the chance of these db race conditions happening? If so, maybe a small comment would be nice.

rust/src/openvasd/scans/scheduling.rs
```diff
@@ -27,6 +27,15 @@ struct ScanScheduler<Scanner, Cryptor> {
...
     scanner: Arc<Scanner>,
     max_concurrent_scan: usize,
+    // we need to cache the ids that are just set from requested to running to not accidentally
+    // start a scan twice when the interval is too short for the sqlitedb to flush that change.
+    // This means on: `requested_to_running` we add the id when we set that id to running, when
+    // fetching requested scans we filter out each id that is stored in this guard and
+    // on scan_import_results it is removed because it is already flushed by sqlitedb.
+    //
+    // This is a hack, if there are any better solutions that work as reliably this can be removed
+    // without functional harm.
```
Tehforsch: I think without the context of your PR description, I wouldn't really understand this comment. 

maybe adding a few explanatory lines like

```
    // Exists to prevent a bug in which scans were accidentally started twice due to deferred writes in sqlite.
    // To address this bug, this field keeps track of all requested scans manually, instead of relying on
    // sqlite for this information.

```

would help.

Tehforsch (approved):
Nice catch. Really unfortunate that we have to do this :|

rust/src/openvasd/scans/scheduling.rs
nichtsfrei: can you check if the comment is sufficient?

Tehforsch (approved):

PR #2018, opened 2025-10-07
Fix: Tests: always free suite

mattmundell:
## What

Always free the suite in the tests.

## Why

Makes the output clearer when running tests with `-fsanitize=address`.

## References

Like https://github.com/greenbone/gvm-libs/pull/939.


jjnicola (approved):

jjnicola:
It looks good. Tests fail because permissions to access the registry.

PR #2021, opened 2025-10-08
Add: Feed state orchestrator

nichtsfrei:
Introduces a vt orchestrator that verifies if a feed update is needed
and sending the message Need into a broadcast that is retrieved by the
scheduler that verifies if a feed update can be started. If so then it
answers allow back. This is a safeguard so that running scans don't get
corruped.

An orchestrator has a worker which does the actual feed update and so
on.

The orchestrator also updates the feed snapshot used by the http
responses for the feed version

SC-1377

### DIFF
--- a/rust/src/container_image_scanner/config/mod.rs
+++ b/rust/src/container_image_scanner/config/mod.rs
@@ -157,7 +157,9 @@ impl SqliteConfiguration {
             && !path.exists()
         {
             // we panic when we cannot create the dir
-            std::fs::create_dir_all(path).ok_or_else(|| panic!("Failed to create dir at {path:?}"));
+            std::fs::create_dir_all(path)
+                .ok()
+                .or_else(|| panic!("Failed to create dir at {path:?}"));
         }
 
         let options = SqliteConnectOptions::from_str(&self.location.sqlite_address(name))?
--- a/rust/src/nasl/interpreter/error.rs
+++ b/rust/src/nasl/interpreter/error.rs
@@ -139,6 +139,8 @@ pub struct IncludeSyntaxError {
     errs: Vec<ParseError>,
 }
 
+// TODO Get rid of this once we have a proper implementation of spans
+// for InterpreterError as well.
 impl std::fmt::Display for IncludeSyntaxError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         emit_errors(&self.file, self.errs.iter().cloned(), Level::Error);
--- a/rust/src/openvasd/scans/mod.rs
+++ b/rust/src/openvasd/scans/mod.rs
@@ -1,9 +1,4 @@
-use std::{
-    num::ParseIntError,
-    pin::Pin,
-    str::FromStr,
-    sync::{Arc, RwLock},
-};
+use std::{num::ParseIntError, pin::Pin, str::FromStr, sync::Arc};
 
 use futures::StreamExt;
 use greenbone_scanner_framework::{entry::Prefixed, models::AliveTestMethods, prelude::*};
@@ -12,7 +7,7 @@ use scannerlib::{
     scanner,
 };
 use sqlx::{Acquire, QueryBuilder, Row, SqlitePool, query, query_scalar, sqlite::SqliteRow};
-use tokio::sync::{broadcast, mpsc::Sender};
+use tokio::sync::mpsc::Sender;
 
 use crate::{
     config::Config,
@@ -703,8 +698,8 @@ pub(crate) fn config_to_crypt(config: &Config) -> ChaCha20Crypt {
 pub async fn init(
     pool: SqlitePool,
     config: &Config,
-    feed_status: broadcast::Sender<orchestrator::Message>,
-    feed_snapshot: Arc<RwLock<FeedState>>,
+    feed_status: orchestrator::Communicator,
+    feed_snapshot: Arc<std::sync::RwLock<FeedState>>,
 ) -> Result<Endpoints<ChaCha20Crypt>, Box<dyn std::error::Error + Send + Sync>> {
     let crypter = Arc::new(config_to_crypt(config));
     let scheduler_sender = scheduling::init(
@@ -724,10 +719,9 @@ pub async fn init(
 
 #[cfg(test)]
 mod tests {
-    use std::{
-        sync::{Arc, RwLock},
-        time::Duration,
-    };
+    use std::time::Duration;
+
+    use super::*;
 
     use futures::StreamExt;
     use greenbone_scanner_framework::{
@@ -752,9 +746,13 @@ mod tests {
     };
 
     async fn init(pool: SqlitePool, config: &Config) -> super::Endpoints<ChaCha20Crypt> {
-        let feed_snapshot = Arc::new(RwLock::new(FeedState::Synced("0".into(), "2".into())));
-        let (feed_state, _) = tokio::sync::broadcast::channel(2);
-        super::init(pool, config, feed_state, feed_snapshot)
+        let feed_snapshot = Arc::new(std::sync::RwLock::new(FeedState::Synced(
+            "0".into(),
+            "2".into(),
+        )));
+        let ignored = Default::default();
+
+        super::init(pool, config, ignored, feed_snapshot)
             .await
             .unwrap()
     }
@@ -1114,13 +1112,13 @@ mod tests {
         let (config, pool) = create_pool().await?;
 
         let crypter = Arc::new(config_to_crypt(&config));
-        let (feed, _) = tokio::sync::broadcast::channel(1);
+        let (_, _, communicator) = orchestrator::Communicator::init();
         let scheduler_sender = scheduling::init_with_scanner(
             pool.clone(),
             crypter.clone(),
             &config,
             scheduling::tests::scanner_succeeded().build(),
-            feed,
+            communicator,
         )
         .await?;
         let undertest = super::Endpoints {
--- a/rust/src/openvasd/vts/orchestrator.rs
+++ b/rust/src/openvasd/vts/orchestrator.rs
@@ -1,46 +1,120 @@
-use std::sync::{Arc, RwLock};
+use std::sync::Arc;
 use std::time::Duration;
+use tokio::sync::RwLock;
 
 use greenbone_scanner_framework::GetVTsError;
 use greenbone_scanner_framework::models::FeedType;
 use scannerlib::models::FeedState;
 use scannerlib::{PinBoxFut, feed};
-use tokio::sync::broadcast::{self, Receiver, Sender};
+use tokio::sync::mpsc;
 
 use crate::vts::FeedHashes;
 
-// Use message instead of directly FeedStatusChange if we want to add meta data later.
-#[derive(Debug, Clone)]
-pub struct Message {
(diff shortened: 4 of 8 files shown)


rust/src/container_image_scanner/config/mod.rs
```diff
@@ -153,6 +153,12 @@ impl SqliteConfiguration {
             pool::PoolOptions,
             sqlite::{SqliteConnectOptions, SqliteJournalMode},
         };
+        if let DBLocation::File(path) = &self.location
+            && !path.exists()
+        {
+            // we panic when we cannot create the dir
+            std::fs::create_dir_all(path).unwrap();
```
Tehforsch: This is going to give very cryptic error messages that could be annoying to solve for anyone who doesnt have access to the code / no backtrace.

```suggestion rust
            std::fs::create_dir_all(path).ok_or_else(|| panic!("Failed to create dir at {path:?}"));
```

rust/src/nasl/interpreter/error.rs
```diff
@@ -139,8 +139,6 @@ pub struct IncludeSyntaxError {
     errs: Vec<ParseError>,
 }
 
-// TODO Get rid of this once we have a proper implementation of spans
-// for InterpreterError as well.
```
Tehforsch, line 143: Why was this todo removed?

rust/src/openvasd/scans/mod.rs
```diff
@@ -738,11 +751,12 @@ mod tests {
         scans::{config_to_crypt, scheduling},
     };
 
-    fn feed_state() -> Pin<Box<dyn Future<Output = FeedState> + Send + 'static>> {
-        Box::pin(async { FeedState::Synced("0".into(), "2".into()) })
-    }
     async fn init(pool: SqlitePool, config: &Config) -> super::Endpoints<ChaCha20Crypt> {
-        super::init(pool, config, feed_state).await.unwrap()
+        let feed_snapshot = Arc::new(RwLock::new(FeedState::Synced("0".into(), "2".into())));
+        let (feed_state, _) = tokio::sync::broadcast::channel(2);
```
Tehforsch: Nit: Why is `feed_snapshot` of type `FeedState` (ignoring arc/rwlock) and `feed_state` of type `Sender`? Confuses me - I'd an prefer `_tx` suffix, or something to make it clear the latter is a sender

rust/src/openvasd/vts/orchestrator.rs
```diff
@@ -0,0 +1,428 @@
...
+        Message { change: value }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum FeedStatusChange {
+    // Communicate downstream that we need to synchronize a feed
+    //
+    // This means a scan scheduler should finish running scans while not allowing new scans until
+    // Synced. When all the o scans are finished then it MUST send Allow back.
```
Tehforsch:
```suggestion
    // Synced. When all the running scans are finished, it MUST send `Allow` back.
```

rust/src/openvasd/vts/orchestrator.rs
```diff
@@ -0,0 +1,428 @@
...
+}
+
+#[cfg(test)]
+pub mod test {
+
+    use super::*;
+    struct BWLer {
+        cached: Option<FeedHashes>,
+        calculated: FeedHashes,
+    }
```
Tehforsch: What is this struct name, some joke that I dont get?

rust/src/openvasd/vts/orchestrator.rs
```diff
@@ -0,0 +1,428 @@
...
+
+        fn calculated_hashes(&self) -> PinBoxFut<Result<FeedHashes, WorkerError>> {
+            let calculated = self.calculated.clone();
+            Box::pin(async move { Ok(calculated) })
+        }
+
+        fn update_feed(&self, _: FeedType, _: String) -> PinBoxFut<Result<(), WorkerError>> {
+            Box::pin(async move { Ok(()) })
+        }
+    }
```
Tehforsch: Don't really love the fact that the `Worker` trait exists only to allow creating this dummy. We could get rid of the entire `Worker` abstraction if we ditched this, so we could get rid of the generics and all the `PinBoxFut`, cleaning up the code significantly. Maybe the unit tests below could still be done with a `FeedSynchronizer` or maybe not - I'd also be fine with getting rid of them if it allows cleaning up the code.  

Just a personal preference, feel free to ignore if you think these unit tests are valuable enough

rust/src/openvasd/vts/orchestrator.rs
```diff
@@ -0,0 +1,428 @@
...
+            outer_state.clone(),
+            &[FeedType::Advisories, FeedType::NASL],
+        )
+        .await
+        .unwrap();
+
+        let outer_state = outer_state.read().unwrap();
+        assert_eq!(outer_state.nasl(), expected_nasl);
+        assert_eq!(outer_state.advisories(), expected_advisories);
+    }
```
Tehforsch: I know you dont believe in DRY but there is so much boilerplate here that I have a hard time seeing what we're even trying to test

rust/src/openvasd/vts/orchestrator.rs
```diff
@@ -0,0 +1,428 @@
...
+
+impl FeedStatusChange {
+    pub fn feed_type(&self) -> &FeedType {
+        match self {
+            FeedStatusChange::Need(feed_type)
+            | FeedStatusChange::Allow(feed_type)
+            | FeedStatusChange::Synced(feed_type) => feed_type,
+        }
+    }
+}
```
Tehforsch: Given that every variant has a `feed_type` anyways and we have a surrounding `Message` struct already, why not extract the `feed_type` to message and rename this to `MessageKind` or whatever

rust/src/openvasd/vts/orchestrator.rs
```diff
@@ -0,0 +1,428 @@
...
+        let (tx, rx) = broadcast::channel(2);
+        let tx2 = tx.clone();
+        let orc = Orchestrator {
+            outer_state,
+            worker: Arc::new(worker),
+            receiver: rx,
+            sender: tx2,
+        };
+        (tx, orc)
+    }
```
Tehforsch, line 173: This is a major point of confusion for me.

We create a channel and then return the sender. Yet we keep the sending end of the channel for ourselves, too. Then on the outside, somebody calls `sender.subscribe()` to get a receiving end. 

I find this confusing for multiple reasons:
1. It is not clear who is sending messages. In other cases I could track down the receiving end returned from the new function and immediately realize it. But now I have to track sending ends too.
2. We lack the ability to tell which message type goes in which direction. I'd much prefer having two messages that are incoming / outgoing (from the perspective of the orchestrator), so the information flow becomes clearer.

Long story short, I think the mpmc/broadcast channel is the wrong channel type here. Switching to mpsc would mean we have to have two channels and two message types, but would really help with making this less confusing, in my opinion.

rust/src/openvasd/vts/orchestrator.rs
```diff
@@ -0,0 +1,428 @@
...
+                | (false, false, true, true)
+                | (true, true, false, false) => break,
+                (sync_nasl, nasl_handle, sync_advisories, advisories_handle) => tracing::debug!(
+                    sync_nasl,
+                    nasl_handle,
+                    sync_advisories,
+                    advisories_handle,
+                    "Continue wairing for Allow message"
+                ),
+            }
```
Tehforsch, line 307: Isnt this just
```rust
if (sync_nasl && nasl_handle.is_some()) || (sync_advisories && nasl_advisories.is_some()) {
    break
}
else {
    tracing::debug!(
        sync_nasl,
        nasl_handle,
        sync_advisories,
        advisories_handle,
        "Continue wairing for Allow message"
    ),
}
```

rust/src/openvasd/vts/orchestrator.rs
nichtsfrei: The reasoning is that we are already aware of another type of worker: redis based.

That said I generally believe that in case of an independent delegation principle, work that is not directly done within that implementation, a trait definition is worth it as it makes it easier to reason about it and also to make it easier to define behaviour that is not conform of our assumption for test cases.

rust/src/openvasd/vts/orchestrator.rs
nichtsfrei: I'm wondering if that would change if I abstract the test logic away though.

rust/src/openvasd/vts/orchestrator.rs
nichtsfrei: I don't know. To me it made more sense to write and read it  'Status of Feed has changed' then 'Message about feed about status change'. But honestly I regret introducing Message already :D

rust/src/openvasd/vts/orchestrator.rs
nichtsfrei: Just a small thing: I do believe in DRY when it's about business logic, which when changed has to change everywhere. But that's not necessarily true about the test cases.

rust/src/openvasd/scans/mod.rs
nichtsfrei: `feed_snapshot` is the current value, aka a snapshot, of the feed used on responses (the value, not the state). feed-state is the broadcast channel, while called Sender it has subscribe as a function.

But yes, maybe a type definition would be better.

rust/src/openvasd/vts/orchestrator.rs
nichtsfrei: There is no mpmc, meaning we either have to introduce a new crate or use mpsc.

rust/src/openvasd/vts/orchestrator.rs
nichtsfrei: It is not. It is either:
- when NASL and advisories is requested => both must be set
- when just NASL => nasl must be set
- when just  advisories => advisories must be set

with your logic it would abort when we require both but just one is set.

nichtsfrei: That's why I decided to keep it like that.

rust/src/openvasd/vts/orchestrator.rs
nichtsfrei: introduced solution based on mpsc. Not sure if I prefer it.

rust/src/openvasd/vts/orchestrator.rs
nichtsfrei: added a helper method with a very good name. excellent name.

rust/src/openvasd/vts/orchestrator.rs
nichtsfrei: I decided to keep worker trait.

rust/src/openvasd/vts/orchestrator.rs
nichtsfrei: renamed.

rust/src/openvasd/scans/mod.rs
nichtsfrei: changed to mpsc system within structs.

rust/src/openvasd/vts/orchestrator.rs
Tehforsch: right, my bad

Tehforsch (approved):

PR #2039, opened 2025-10-20
Chore: introducing construct to ensure that the scan state has changed

nichtsfrei:
Why:
- sqlx on SQLite uses deferred transactions. Pooled connections can see
  stale rows and reorder writes. The scheduler mixed reads and writes
  across connections which produced races and ghost scans.
- We carried a requested_guard hack to mask double starts. It added
  state that diverged from the DB and still missed edge cases.
- Cached statements and long-lived transactions caused status checks to
  read old values. That led to scans stuck in requested or running.

What:
- Introduce ScanStateChange that performs compare-and-set updates and
  reads through one detached connection guarded by a mutex.
- Add retry on IO errors with connection replacement and cache clearing.
- Provide primitives: change_state, change_state_all, fetch_scans_in_state,
  scan_get_status, scan_update_status, count_scans_in_state.

How it improves correctness:
- Status transitions become atomic CAS operations on a single connection.
- Eliminates cross-connection visibility lag and stale reads.
- Removes requested_guard and the double-start class of bugs.
- Startup cleanup uses change_state_all to flip orphaned running -> failed.
- Scheduling computes capacity from authoritative running count and fetches
  requested IDs via the same connection. No mixed-connection race.

Scope:
- Replace scattered SQL in scheduling with ScanStateChange calls.
- No schema changes. Tests updated and new ones added for single and bulk
  transitions.

Operational effect:
- Fewer stuck scans. No duplicate starts. Clearer logs with affected row
  counts. Safer shutdown and restart behavior.

SC-1377

### DIFF
--- a/rust/src/openvasd/scans/mod.rs
+++ b/rust/src/openvasd/scans/mod.rs
@@ -7,10 +7,7 @@ use scannerlib::{
     scanner,
 };
 use sqlx::{
-    Acquire, QueryBuilder, Row, Sqlite, SqlitePool, query,
-    query::Query,
-    query_scalar,
-    sqlite::{SqliteArguments, SqliteRow},
+    Acquire, QueryBuilder, Row, Sqlite, SqlitePool, query, query_scalar, sqlite::SqliteRow,
 };
 use tokio::sync::mpsc::Sender;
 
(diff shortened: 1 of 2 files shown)


rust/src/openvasd/scans/mod.rs
```diff
@@ -595,12 +601,16 @@ AND result_id = ?"#,
...
-    let scan_row = query(r#"
+pub(crate) fn status_query<'a>(id: i64) -> Query<'a, Sqlite, SqliteArguments<'a>> {
+    sqlx::query(r#"
         SELECT created_at, start_time, end_time, host_dead, host_alive, host_queued, host_excluded, host_all, status
         FROM scans
         WHERE id = ?
-        "#).bind(id).fetch_one(pool).await?;
+        "#)
+        .bind(id)
+}
```
Tehforsch: Maybe we could move this query to the state_change file to keep this file clean of the sqlx details

rust/src/openvasd/scans/state_change.rs
```diff
@@ -0,0 +1,315 @@
...
+    //             .await?;
+    //
+    //         if status.as_deref() == Some(on) {
+    //             ids.swap_remove(0);
+    //         }
+    //
+    //         tokio::time::sleep(self.retry).await;
+    //     }
+    //     Ok(())
+    // }
```
Tehforsch: remove?

rust/src/openvasd/scans/state_change.rs
```diff
@@ -0,0 +1,315 @@
...
+            .await?;
+        tracing::debug!(
+            affected = rows.rows_affected(),
+            from,
+            to,
+            id,
+            "Scan status change"
+        );
+
+        //        self.block_on(to, vec![id]).await?;
```
Tehforsch: remove?

Tehforsch (dismissed):

PR #2040, opened 2025-10-21
Fix: SQL Variable limit

Kraemii:
In some cases it can happen, that the limit of SQLite Variables is reached, e.g. when inserting OIDs of VTs or huge lists of Ports. In such cases the values are splitted into chunks depending on the Variable Limit divided by the number of fields. The limit for SQLite is 32766 (https://docs.rs/sqlx-core/0.8.6/sqlx_core/query_builder/struct.QueryBuilder.html#method.push_bind)

SC-1483

### DIFF
--- a/rust/src/container_image_scanner/scheduling/db/mod.rs
+++ b/rust/src/container_image_scanner/scheduling/db/mod.rs
@@ -7,9 +7,12 @@ use sqlx::{
 };
 use tracing::debug;
 
-use crate::container_image_scanner::{
-    ExternalError,
-    image::{Credential, Image, ImageID, ImageParseError},
+use crate::{
+    SQLITE_LIMIT_VARIABLE_NUMBER,
+    container_image_scanner::{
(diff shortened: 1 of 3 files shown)


rust/src/container_image_scanner/scheduling/db/mod.rs
```diff
@@ -222,26 +225,29 @@ pub async fn store_results(
...
-            .push_bind(detail.name)
-            .push_bind(detail.value)
-            .push_bind(detail.source.s_type)
-            .push_bind(detail.source.name)
-            .push_bind(detail.source.description);
-    });
-
-    let query = builder.build();
-    query.execute(&mut *tx).await?;
+    for results in results.chunks(SQLITE_LIMIT_VARIABLE_NUMBER / 14) {
```
nichtsfrei, line 228: q: I'm wondering if it's worth to put that behind a helper function?

PR #2044, opened 2025-10-22
Make NASL rust interpreter NASL C backward compatible

jjnicola:
**What**:
Add the `naslv2` feature.
Local variables inside inner scopes, belong to a parent root scope or global. 
SC-1389

**Why**:

**How**:

Build with and `naslv2` features and compare with openvas-nasl the following script
```
if (1) {
	a = 100; # it is inside an if scope, but still global.
}
display(" AAA: ", a);

function new_scope() {
	f = 100;
	display("FFF inside",f);
	display("Trying AAAA inside function", a); # a belongs to global scope

}
display("FFFF: ",f); # f is set in the fn scope
new_scope();

a = 1;
if (a) {
    local_var a;
    a = 23;
}
display(a);

a = 1;
b = 2;
{
  local_var a;
  a = 3; b = 4;
}
display(a);
display(b);
```

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/src/nasl/builtin/misc/mod.rs
+++ b/rust/src/nasl/builtin/misc/mod.rs
@@ -311,7 +311,6 @@ fn scanner_status() -> NaslValue {
     NaslValue::Null
 }
 
-
 pub struct Misc;
 
 function_set! {


Tehforsch (commented):
General comment: I think instead of making this a compile time feature, we should keep this at runtime (which is were it has to be if we want scripts to be able to opt in to different versions).

The `ForkingInterpreter` already has the `with_version` which we can use, but there isnt a concept for this in the `interpreter_test_ok!` macro yet.  For `interpreter_test_err!` I introduced a second variant `interpreter_test_err_v2` which sets the version to v2. You could introduce the equivalent `interpreter_test_ok_v2` and then use this to test the new behavior.

rust/src/nasl/interpreter/register.rs
```diff
@@ -81,17 +89,43 @@ impl Register {
...
+                if parent_scope.is_fn_scope || parent_scope.parent_index() == 0 {
+                    return parent_index;
+                } else if parent_scope.parent_index() == 0 {
+                    return 0;
+                } else {
+                    parent_index = parent_scope.parent_index();
+                }
+            }
+        }
+    }
```
Tehforsch: Looking at this, I wonder if its simpler to skip the `create_child` and `drop_last` calls in `resolve_block` (dependent on a boolean flag which is always true when `resolve_block` is called from `resolve_fn` and always false otherwise) so that only functions create scopes in NASL v1.

rust/src/nasl/interpreter/tests/mod.rs
```diff
@@ -38,7 +38,39 @@ macro_rules! interpreter_test_ok {
...
+        }
+    };
+}
+
+#[macro_export]
+macro_rules! interpreter_test_ok_v2 {
+    ($name: ident, $code: literal, $($expected: expr),* $(,)?) => {
+        #[test]
+        fn $name() {
+            let mut results = $crate::nasl::interpreter::tests::interpret($code, $crate::nasl::prelude::NaslVersion::V2);
```
Tehforsch: I think I would prefer having one 

`interpreter_test_ok!` macro that writes the test and accepts a version argument (similar to the second match in the current implementation) and then `interpreter_test_ok_v1!` and `interpreter_test_ok_v2!` as small wrappers that use it.

rust/src/nasl/interpreter/mod.rs
```diff
@@ -499,7 +499,14 @@ impl<'ctx> Interpreter<'ctx> {
     }
 
     pub(crate) async fn resolve_block(&mut self, block: &Block<Statement>) -> Result {
-        self.register.create_child();
+        let mut fncall = false;
+        if let Some(Statement::FnDecl(_)) = block.items.first() {
+            fncall = true;
+        }
```
Tehforsch: I don't think this is the right logic. This checks whether the first statement in the block is a function decl, so this block would match:

```
function foo() {
    function bar() { }
}
```

but this one wouldnt:

```
function foo() {
    x += 1;
    bar()
}
```

what we really want is to check if the "reason" for opening this block was because it's the block of a function, or an if/for/... 

P.S.: Looking at the code, I realize now that function calls already create the child scope anyways (interpreter/call.rs:127),
so we can actually get rid of this entire logic altogether, if my guess is correct. Sorry, didn't realize this when I wrote the last review.

The only other blocks would then be if/for/while/do, which we don't want to create local scopes in NASL::V1 (unless this is specific to `if`)

rust/src/nasl/interpreter/mod.rs
```diff
@@ -499,7 +499,14 @@ impl<'ctx> Interpreter<'ctx> {
...
+
+        if (fncall && self.version == NaslVersion::V1) || self.version == NaslVersion::V2 {
```
Tehforsch: I recommend matching over versions instead of checking them in an if, so that we will automatically be notified of all the sites in the code we need to check if we add `NaslVersion::V3` at some point:

```
let add_scope = match self.version { 
    V1 => false,
    V2 => true,
};
```

or, also acceptable:

```
let add_scope = !matches!(self.version, NaslVersion::V1);
```

but not 

```
let add_scope = matches!(self.version, NaslVersion::V2);
```

since the latter would silently fail for (the hypothetical) V3

rust/src/nasl/version.rs
```diff
@@ -1,4 +1,4 @@
-#[derive(Clone, Copy, Default, Debug)]
+#[derive(Clone, Copy, Default, Debug, PartialEq)]
```
Tehforsch: In the sense of what I wrote above, I'd remove the `PartialEq` derive here, so that we are incentivized to use `match` and `matches` whenever possible.

rust/src/nasl/interpreter/mod.rs
jjnicola: I have just tested it. The first example above is not allowed in NASLv1 (function declaration inside functions).

rust/src/nasl/interpreter/tests/mod.rs
```diff
@@ -160,13 +179,24 @@ mod operator {
...
 interpreter_test_ok!(
     block_scope,
     "a = 1; b = 2; { local_var a; a = 3; b = 4; } a; b;",
     1,
     2,
     NaslValue::Null,
-    1,
+    3,
     4,
 );
```
Tehforsch, line 201: I'm confused about this test - given that we explicitly declare `a` as local, shouldn't it behave in a sane way, even in V1?

jjnicola: Both, C and Rust Naslv1, behaves as in the example. Like if `a` has already been declared as global, it is still global.

Tehforsch: ah, that makes sense, thanks.

jjnicola: A different situation would be if it is a function declaration. Otherwise, the scope it is still global

jjnicola: In this case, the `a` will be in the function scope as local.
``` c
a = 1;
function foo123 () {
 local_var a;
 a = 3;
 b = 4;
 display (a);
}
foo123();
display( a, "    ", b);
```

PR #2047, opened 2025-10-27
NASL Rust implementation progress and coverage documentation

Kraemii:
Meanwhile we had 3 sources for the list of NASL functions:
- Documentation
- Static List in `feed_filter`
- Static NASL script

Therefore I decided to enhance the implementation of the `feed_filter` to also be able to get statistics of the coverage of the current rust implementation. The base source for this information is the Documentation. As I also deleted the static list and NASL script, there is now only 1 source of truth regarding the list of Builtin-functions and all statistics are derived from it.

Issues became clear, when new features for the c based scanner where still developed. Examples are e.g. the Kerberos support and direct NASL call. This lead to an inconsistency between the former 3 sources.

Jira: SC-1195 SC-1394

### DIFF
--- a/rust/src/feed_filter/main.rs
+++ b/rust/src/feed_filter/main.rs
@@ -65,28 +65,24 @@ struct BuiltinStats {
 }
 
 impl BuiltinStats {
-    fn new(scripts: &Scripts, builtins: &DocumentedFunctions) -> Self {
+    fn new(scripts: &Scripts, builtins: &BuiltinFunctions) -> Self {
         let mut category_stats = HashMap::new();
-        let mut undocumented = vec![];
         let mut function_calls = HashMap::new();
 
-        for func in builtins.undocumented.iter() {
-            undocumented.push(func.clone());
+        // Initialize all built-in functions with zero calls
+        for (func, _) in builtins.unimplemented().iter() {
+            function_calls.entry(func.clone()).or_insert(0);
+        }
+        for (func, _) in builtins.implemented().iter() {
+            function_calls.entry(func.clone()).or_insert(0);
         }
 
+        // Count function calls in all scripts
         for (_, script) in scripts.iter() {
             for call in iter_fn_calls(&script.ast) {
                 let function = call.fn_name.to_string();
                 *function_calls.entry(function).or_insert(0) += 1;
             }
-        }
-
-        for (func, _) in builtins.unimplemented().iter() {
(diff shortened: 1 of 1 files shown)


rust/src/feed_filter/main.rs
```diff
@@ -38,58 +26,271 @@ fn init_logging() {
...
+
+impl BuiltinStats {
+    fn new(scripts: &Scripts, builtins: &DocumentedFunctions) -> Self {
+        let mut category_stats = HashMap::new();
+        let mut undocumented = vec![];
+        let mut function_calls = HashMap::new();
+
+        for func in builtins.undocumented.iter() {
+            undocumented.push(func.clone());
+        }
```
Tehforsch: Personally, I prefer doing this with an immutable variable via

```rust
let undocumented = builtins.undocumented.iter().cloned().collect()
```

since reading code with many mutable variables is a lot more difficult

rust/src/feed_filter/main.rs
```diff
@@ -38,58 +26,271 @@ fn init_logging() {
...
+
+/// DocumentedFunctions holds implemented and unimplemented built-in functions
+struct DocumentedFunctions {
+    /// HashMap<function_name, (category, deprecated)>
+    implemented: HashMap<String, (String, bool)>,
+    /// HashMap<function_name, (category, deprecated)>
+    unimplemented: HashMap<String, (String, bool)>,
+    /// Set of undocumented functions
+    undocumented: HashSet<String>,
+}
```
Tehforsch: I think this should be renamed, given that it also contains undocumented functions

rust/src/feed_filter/main.rs
```diff
@@ -38,58 +26,271 @@ fn init_logging() {
...
+            }
+        }
+
+        for (func, _) in builtins.unimplemented().iter() {
+            function_calls.entry(func.clone()).or_insert(0);
+        }
+
+        for (func, _) in builtins.implemented().iter() {
+            function_calls.entry(func.clone()).or_insert(0);
+        }
```
Tehforsch: Minor, unimportant nitpick: I found this a little confusing while reading this the first time - if I understand correctly this is just here to set all entries that hadn't previously been set to 0, so maybe we could move this above the for-loop, just to make the intent clearer.

Tehforsch (approved):
Thanks! Just a few minor comments that could also be ignored

Tehforsch (approved):

PR #2056, opened 2025-11-03
Implement Redis based VT cache

nichtsfrei:
When OpenVASD is started on openvas scanner it is using a redis based VT
cache instead of a sqlite based cache to not waste memory as the openvas
binary needs that data within redis anyway.

The downside is that /vts?information=true is very slow as we create a
new connection per oid when translating. Since GVMD does not rely on
that endpoint and it is an unofficial one I think that can be ignored.

Additionally storage::VT has been replaced with VTData.

https://jira.greenbone.net/browse/SC-1484

### DIFF
--- a/rust/src/openvasd/scans/scheduling.rs
+++ b/rust/src/openvasd/scans/scheduling.rs
@@ -573,30 +573,6 @@ where
                 preferences::preference::PREFERENCES.to_vec(),
             );
 
-            // TODO: maybe we are in the state that it is actually easier to write a redis worker?
-
-            // let (sender, mut receiver, black_mirror) = orchestrator::Communicator::init();
-            //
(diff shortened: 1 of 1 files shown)


rust/src/openvasd/scans/scheduling.rs
```diff
@@ -689,32 +573,31 @@ where
...
+            //                    tracing::warn!(%error, "Unable to inform orchestrator about allow. VTs are stuck");
+            //                 }
+            //
+            //
+            //             }
+            //
+            //         }
+            //     }
+            // });
+            //
```
jjnicola: Do you want to leave this commented or should it be removed ?

nichtsfrei: Oops, is deleted.

PR #2061, opened 2025-11-05
Fix  Add or Update detect-hidden-unicode.yml

greenbonebot:


nichtsfrei:
Nit: would it be possible to get an easier method to white-list files that are allowed to have hidden characters? We do quite a bit of wonky testing, as provide a NASL interpreter, so there is a possibility that we write will write a NASL test that contains a hidden character, which in return could be quite confusing to debug why it is failing with this action enabled.

PR #2062, opened 2025-11-05
Compose stuff

nichtsfrei:
To test this you need to first build an local image (to have the fix provided in this pr within the openvas image)

```
podman build \
  -t registry.community.greenbone.net/community/openvas-scanner:stable \
  -f .docker/prod.Dockerfile .
```

Adds docker-compose to start openvasd and some test cases.

To start openvasd with http call:

```
podman-compose -f base.yaml
```

To start openvasd with https but without mtls:

```
podman-compose -f base.yaml -f tls.yaml
```

To start openvasd with mTLS:

```
podman-compose -f base.yaml -f mtls.yaml
```

To override the used certificate set the environment variables:

- ${OPENVASD_SERVER_PEM:-./openvasd-server.pem}:/var/lib/openvasd/certs/server.pem:ro
- ${OPENVASD_SERVER_KEY:-./openvasd-server.key}:/var/lib/openvasd/certs/server.key:ro
- ${OPENVASD_CLIENT_KEYS:-./client-certs}:/var/lib/openvasd/certs/clients:ro

If you want to verify your setup you can also deploy the
tests/victim.yaml:

```
podman-compose -f base.yaml -f mtls.yaml -f tests/victim.yaml
```

and use the Makefile within tests to create scans:

```
make create-victim-simple-auth-ssh
make start-victim-simple-auth-ssh
make results-victim-simple-auth-ssh
make status-victim-simple-auth-ssh
```

The naming scheme of that Makefile is
- create-
- start-
- results-
- stop-
- rm-

and the name of the json within `tests/scans` without the json suffix.

DEVOPS-1813

### DIFF
--- a/compose/README.md
+++ b/compose/README.md
@@ -13,22 +13,50 @@ It is divided into three definitions:
 2. tls.yaml - overrides the settings within base.yaml to start OpenVASD in TLS mode
 3. mtls.yaml - overrides the settings within base.yaml to start OpenVASD in mTLS mode (preferred)
 
-Which means to start openvasd with http:
+## How to start
+
+### HTTP
+
+To start the scanner compose with http you can use the base.yaml:
 
 ```
-podman-compose -f base.yaml
+podman-compose -f base.yaml up
 ```
 
-in TLS mode
+### HTTPS
+
+By default the compose definitions (tls.yaml as well as mtls.yaml) use the certificates:
+- ./openvasd-server.key
+- ./openvasd-server.pem
+the directory
+- ./client-certs
+for the client certificates.
+
+You can either copy your certificates into that location or set the environment variables:
+- OPENVASD_SERVER_KEY - the key file of your certificate
+- OPENVASD_SERVER_PEM - the pem file of your certificate
+- OPENVASD_CLIENT_CERTS - the directory containing public certificates of clients that are allowed to use OpenVASD 
+
+NOTE: if the files specified by OPENVASD_SERVER_KEY, OPENVASD_SERVER_PEM or the
+directory OPENVASD_CLIENT_CERTS are not available directories with that path
+may be created.
+
+If you don't have certificates you can use the Makefile to create some:
 
 ```
-podman-compose -f base.yaml -f tls.yaml
+make
 ```
 
-in mTLS mode:
+To start scanner compose with TLS but without client-certificates:
 
 ```
-podman-compose -f base.yaml -f mtls.yaml
+podman-compose -f base.yaml -f tls.yaml up
+```
+
+To start scanner compose with mTLS:
+
+```
+podman-compose -f base.yaml -f mtls.yaml up
 ```
 
 when in mTLS mode you have to provide the client-certificate and the corresponding key when connecting to OpenVASD:
@@ -40,30 +68,15 @@ curl -vk \
     https://localhost:3000/scans
 ```
 
-On default the compose definitions (tls.yaml as well as mtls.yaml) use the certificates:
-- ./openvasd-server.key
-- ./openvasd-server.pem
-the directory
-- ./client-certs
-for the client certificates.
+## How to test
 
-You can either copy your certificates into that location or set the environment variables:
-- OPENVASD_SERVER_KEY - the key file of your certificate
-- OPENVASD_SERVER_PEM - the pem file of your certificate
-- OPENVASD_CLIENT_KEYS - the directory containing public certificates of clients that are allowed to use OpenVASD 
-
-If you don't have certificates, but want to test OpenVASD you can call:
-
-```
-make
-```
 
 Additionally to the OpenVASD compose definition we also provide a possibility to verify the setup.
 
-This is done by adding `test/victim.yaml` to the compose chain like
+This is done by adding `tests/victim.yaml` to the compose chain like
 
 ```
-podman-compose -f base.yaml -f mtls.yaml -f tests/victim.yaml
+podman-compose -f base.yaml -f mtls.yaml -f tests/victim.yaml up
 ```
 
 and then use `make` within the `tests` directory to create and start predefined scans:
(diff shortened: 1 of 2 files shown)


compose/README.md
```diff
@@ -0,0 +1,102 @@
...
+
+It is divided into three definitions:
+1. base.yaml - contains the actual definition of the services starts OpenVASD in http mode
+2. tls.yaml - overrides the settings within base.yaml to start OpenVASD in TLS mode
+3. mtls.yaml - overrides the settings within base.yaml to start OpenVASD in mTLS mode (preferred)
+
+Which means to start openvasd with http:
+
+```
+podman-compose -f base.yaml
```
henning: minor detail: 
to actually start the service, users must also add an "up" to make this an actual full working command ;)
(at least with my podman-compose version 1.0.6  on ubuntu - maybe other versions have up as a default that doesn't need to be typed?)

compose/README.md
```diff
@@ -0,0 +1,102 @@
...
+in TLS mode
+
+```
+podman-compose -f base.yaml -f tls.yaml
+```
+
+in mTLS mode:
+
+```
+podman-compose -f base.yaml -f mtls.yaml
```
henning: if this command will only succeed after the things described below are handled, the command could be mentioned after them... otherwise impatient users will just run the command now and fail

compose/README.md
```diff
@@ -0,0 +1,102 @@
...
+the directory
+- ./client-certs
+for the client certificates.
+
+You can either copy your certificates into that location or set the environment variables:
+- OPENVASD_SERVER_KEY - the key file of your certificate
+- OPENVASD_SERVER_PEM - the pem file of your certificate
+- OPENVASD_CLIENT_KEYS - the directory containing public certificates of clients that are allowed to use OpenVASD 
+
+If you don't have certificates, but want to test OpenVASD you can call:
```
henning: proposal so readers understand what this does:

"...you can run this in order to have some generated for you suited for testing/playing around"

compose/README.md
henning: actually probably more important than i thought, as when i stopped it again, and tried to make it work, I had a directory (yes, not a file) openvasd-server.key that i first had to delete and then the certificates got generated correctly, service could be started and the make commands from the test folder succeeded

henning:
Trying to test this, after 

podman-compose -f base.yaml -f mtls.yaml -f tests/victim.yaml up

it seems like there is a scanner running, but to verify it as the description above says i get: 

```
$ make create-victim-simple-auth-ssh
curl -vk --key ../client-keys/client1.key --cert ../client-certs/client1.pem  -X POST -H "Content-Type: application/json" \
	--data-binary @scans/victim-simple-auth-ssh.json \
	https://localhost:3000/scans | sed 's/"//g' > known-scans/victim-simple-auth-ssh
Note: Unnecessary use of -X or --request, POST is already inferred.
* Host localhost:3000 was resolved.
* IPv6: ::1
* IPv4: 127.0.0.1
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying [::1]:3000...
* connect to ::1 port 3000 from ::1 port 53358 failed: Connection refused
*   Trying 127.0.0.1:3000...
* connect to 127.0.0.1 port 3000 from 127.0.0.1 port 47848 failed: Connection refused
* Failed to connect to localhost port 3000 after 0 ms: Couldn't connect to server
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
* Closing connection
curl: (7) Failed to connect to localhost port 3000 after 0 ms: Couldn't connect to server
```

could be solved by reading the full README from the changed files here, create the certs etc... then it runs.

so besides the above comments this looks good ;)

compose/README.md
nichtsfrei: thanks, how did you stop it to trigger the dir creation?

compose/README.md
nichtsfrei: thanks.

compose/README.md
nichtsfrei: ya, no, that's a typo, thanks for pointing it out.

henning:
one more thing: while looking into the containers etc in order to try to reproduce the "directories named like the key files that should be there" problem i saw that the victim container still uses an image from your personal account: docker.io/nichtsfrei/victim:latest

compose/README.md
henning: ok so if I delete all eventually previously started containers to ensure i'm in a real clean state, remove eventually pre-existing client-certs, openvasd-server.key, openvasd-server.pem files/dirs, and then just run 

`podman-compose -f base.yaml -f mtls.yaml up`

without preparing the client certs in any way, then I will get directories named client-certs, openvasd-server.key, openvasd-server.pem created somehow.  So that seems to be happening in the conatiner.

Likely reason: if volume is defined to be mounted into the container, but the source isn't there on the host, apparently podman tries to be forgiving and creates a directory on the host and mounts that into the container instead of failing and telling it cannot mount a nonexisting dir/file.

One idea as said is to only have this command in the docs *after* explaining the setup of required certs, and another one to avoid these complications for users might be to provide make targets that ensure everything is done correctly for the simple self-generated certs case so 

make start-with-autogenerated-cert

will get users up and running to play around... if they then need something special, they can still deep dive into everything and do their adjustments.

nichtsfrei:
That it is hosted from my account isn't currently considered a problem, as it is only used for testing and is required for running.  it is done that way because there is no greenbone equivalent. This was the case also for https://github.com/greenbone/scanner-lab. There is already a ticket to think about a proper solution, currently there is none.

compose/README.md
nichtsfrei: As this is a non changeable behavior, neither for docker nor podman, and openvasd does abort with a warning that it is a directory instead of a file  i will just add a note in the README.

PR #2064, opened 2025-11-10
Fix: don't check hashsum during feed update

jjnicola:
**What**:
Fix: don't check hashsum during feed update when signature check is disabled.
Jira: SC-1494

**Why**:

**How**:

**Checklist**:

- [ ] Tests


rust/src/notus/loader/hashsum.rs
```diff
@@ -131,7 +131,9 @@ impl AdvisoryLoader for HashsumAdvisoryLoader {
             .ok_or_else(|| Error::UnknownProduct(os.to_string()))?
             .map_err(Error::HashsumLoadError)?;
 
-        file_item.verify().map_err(Error::HashsumLoadError)?;
+        if signature_check {
```
nichtsfrei, line 134: nit: I don't like that disable_signature check disables the signature verification (gpg signature of hashsumfile) and the sha256 verification. I would rather have a config file only option to disable the sha256 sum verification separately.

nichtsfrei (approved):

rust/src/notus/loader/hashsum.rs
jjnicola: I think it is not necessary a second dedicated option for checking the hash sum. Once you disable the signature check what it is in the sha256 file can be changed and it is not important anymore. Content will be just used for loading the feed. 
We can discussed this later if you want, may be there is a reason I am not considering .

PR #2065, opened 2025-11-11
Fix: scannerctl scan-config translator

jjnicola:
**What**:
Fix: scannerctl scan-config translator
SC-1498

**Why**:
scannerctl crashes when there is no preferences in the src XML scan config

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/src/scannerctl/scan_config.rs
+++ b/rust/src/scannerctl/scan_config.rs
@@ -93,7 +93,6 @@ async fn execute(
     };
     let mut vts = HashSet::new();
     for a in config.iter().map(|f| {
-        tracing::info!("{:?}", &f);
         as_bufreader(f)
             .and_then(|r| parse_vts(r, storage.as_ref(), &scan.vts).map_err(|e| map_error(f, e)))
     }) {


rust/src/scannerctl/scan_config.rs
```diff
@@ -245,7 +246,7 @@ struct ScanConfigNvtSelector {
 
 #[derive(Clone, Debug, Deserialize, PartialEq)]
 struct ScanConfigPreferences {
-    preference: Vec<ScanConfigPreference>,
+    preference: Option<Vec<ScanConfigPreference>>,
```
Tehforsch: Is there a good reason we couldn't do

```rust
    #[serde(default)]
    preference: Vec<ScanConfigPreference>,
```

instead? This would default to an empty vec which sounds better to me than using an option

rust/src/scannerctl/scan_config.rs
```diff
@@ -93,6 +93,7 @@ async fn execute(
     };
     let mut vts = HashSet::new();
     for a in config.iter().map(|f| {
+        tracing::info!("{:?}", &f);
```
nichtsfrei: That does not look like an info statement?

jjnicola: ups!

PR #2070, opened 2025-11-18
Codename lookup on OS detection when no VERSION got found.

nichtsfrei:
Fix: add codename lookup when version is not set

Some distributions (Debian) do not set a Version on their testing image,
that's why we need a codename lookup when there is no Version found.

### DIFF
--- a/rust/src/openvasd/container_image_scanner/detection.rs
+++ b/rust/src/openvasd/container_image_scanner/detection.rs
@@ -70,6 +70,7 @@ where
     async fn detect_operating_system(
         self,
     ) -> Result<OperatingSystem, OperatingSystemDetectionError> {
+        #[rustfmt::skip]
         const CODE_NAME_LOOKUP: &[(&str, &str, &str)] = &[
             // At least for debian: as soon as they are stable we can delete that
             // To improve that we could generate a lookup table automatically based on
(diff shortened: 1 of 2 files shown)


rust/src/openvasd/container_image_scanner/scheduling/mod.rs
```diff
@@ -281,45 +281,42 @@ where
         let scans = Self::set_images_to_scanning(config.clone(), &pool)
             .await
             .unwrap();
-        let mut join_handler = FuturesUnordered::new();
+        let mut join_handler = Vec::with_capacity(scans.len());
```
Tehforsch: just fyi, using iterator+collect instead of a forloop+push makes this obsolete, since the vec will be allocated using the iterators `size_hint`

rust/src/openvasd/container_image_scanner/detection.rs
```diff
@@ -70,42 +70,61 @@ where
...
+            ("forky", "14", "14 (forky)"),
+            ("duke", "15", "15 (duke)"),
+            // dear rustfmt don't collapse
+        ];
+        const OS_RELEASE_KEYS: &[&str] = &[
+            "ID",
+            "VERSION_ID",
+            "VERSION",
+            "VERSION_CODENAME",
+            // dear rustfmt don't collapse
```
Tehforsch: FYI: `#[rustfmt::skip]` exists

nichtsfrei: I know, but I found it more funny that way ... will apply your professional solution :)

PR #2071, opened 2025-11-19
Tidy Up Crypto

Kraemii:
Tidy up the crypto builtins:

1. Adjust `StringOrData` Type to not automatically convert stuff
2. Remove custom parameter handle and use `nasl_function` macro instead


rust/src/nasl/builtin/cryptographic/aes_ccm.rs
```diff
@@ -37,19 +37,22 @@ where
 }
 
 /// Base function for ccm en- and decryption. Sets the tag length to 16.
-fn ccm<D>(register: &Register, crypt: Crypt, auth: bool) -> Result<NaslValue, FnError>
+fn ccm<D>(
+    key: &[u8],
+    data: &[u8],
+    nonce: &[u8],
```
nichtsfrei, line 44: q: Wouldn't it be nicer to create a struct for key, data, nonce? They are basically everywhere.

Kraemii: Not sure if that would help, as we still need them everywhere for the nasl function parameter. So yes the function head would be look cleaner, but then you had to pack the stuff into a struct every time and also unpack the stuff right at the beginning of the function

rust/src/nasl/builtin/cryptographic/misc.rs
```diff
@@ -9,15 +9,8 @@ use rand::{Rng, rng};
...
-    let data = register
-        .local_nasl_value("in")
-        .map_err(|_| ArgumentError::MissingNamed(vec!["in".into()]))?;
-    let data = match data {
-        NaslValue::Data(x) => x,
-        NaslValue::String(x) => x.as_bytes(),
-        _ => return Err(ArgumentError::WrongArgument("expected Data.".to_string()).into()),
-    };
+    // As in `is` a keyword in rust, we cannot use the nasl_function annotation for named arguments.
+    let data = StringOrData::from_nasl_value(register.local_nasl_value("in")?).map(|x| x.data())?;
```
Tehforsch, line 13: You could use `r#in` as an identifier and then add support for that in the proc macro. I believe we had a similar problem in another builtin function somewhere, but I can't find it right now

rust/src/nasl/builtin/cryptographic/pem_to.rs
```diff
@@ -11,9 +11,11 @@ use nasl_function_proc_macro::nasl_function;
 use rsa::{RsaPrivateKey, pkcs8::DecodePrivateKey, traits::PrivateKeyParts};
 
 #[nasl_function(named(passphrase))]
-fn pem_to_rsa(register: &Register, passphrase: String) -> Result<NaslValue, FnError> {
-    let ori_pem = StringOrData::from_nasl_value(register.local_nasl_value("priv")?).map(|s| s.0)?;
-    let decrypted_key = match RsaPrivateKey::from_pkcs8_encrypted_pem(&ori_pem, passphrase) {
+fn pem_to_rsa(register: &Register, passphrase: StringOrData) -> Result<NaslValue, FnError> {
```
Tehforsch, line 14: See above, I think adding support for raw idents and using `r#priv` would be ideal

rust/src/nasl/builtin/cryptographic/aes_ccm.rs
nichtsfrei: why would you need to unpack if you could just add the functions to that struct then?

PR #2074, opened 2025-11-20
Fix: host progress

jjnicola:
**What**:
Fix host progress. 
Add a new DB table to keep track of the scanning hosts and their progress.
Jira: SC-1502

**Why**:
progress was not working. For a single host scan it is in 0% until it finishes.

**How**:
run a scan via gvmd/gsa and check that the scan progress works. Also, it can be tested with curl.

**Checklist**:

- [ ] Tests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ message ("-- Configuring the Scanner...")
 
 # VERSION: Always include major, minor and patch level.
 project (openvas
-  VERSION 23.32.1
+  VERSION 23.32.2
   LANGUAGES C)
 
 if (POLICY CMP0005)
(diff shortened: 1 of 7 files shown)


rust/src/openvasd/scans/state_change.rs
```diff
@@ -64,14 +64,27 @@ pub(crate) fn row_to_models_status(scan_row: SqliteRow) -> models::Status {
     let dead = scan_row.get("host_dead");
     let alive = scan_row.get("host_alive");
     let finished = excluded + dead + alive;
+    let host_scanning: String = scan_row.get("host_scanning");
+
+    let host_progress: HashMap<String, i32> = host_scanning
```
nichtsfrei: I know that it is a draft, but I would highly recommend to create a table for host_scanning so that we don't have to do this comma separation thingy everywhere and you don't need to alter a table.

```
CREATE TABLE scan_currently_running(
    id INTEGER,
    host TEXT NOT NULL,
    PRIMARY KEY (id, host),
    FOREIGN KEY (id) REFERENCES client_scan_map(id) ON DELETE CASCADE
);
```

PR #2078, opened 2025-11-27
Clean up usage of the loader trait.

Tehforsch:
Previously, we used a mix of proper generics (`L: Loader`)  and dynamic dispatch (`Box<dyn Loader>`). These are now replaced  with a single struct (`Loader`) encapsulating dynamic dispatch  internally. All `L: Loader` generics were removed.

The point of the `Loader` (now `NaslLoader`) trait is to allow  dependency injection for file loading in tests. There are two use cases  for this trait:
1. Loading directly from a file system in normal operation (sometimes also used in tests).
2. Loading fake files from given strings in tests.

Number 2 was either via an impl of the trait for `Fn(&str) -> String` closures or via custom trait impls at the test site. These are now provided via a single `TestLoader` that is visible via the `Loader::test()` API.

### DIFF
--- a/rust/src/scanner/scanner_stack.rs
+++ b/rust/src/scanner/scanner_stack.rs
@@ -2,8 +2,6 @@
 //
 // SPDX-License-Identifier: GPL-2.0-or-later WITH x11vnc-openssl-exception
 
-use crate::nasl::syntax::Loader;
-
 use crate::nasl::utils::scan_ctx::ContextStorage;
 use crate::scheduling::{ConcurrentVT, ConcurrentVTResult, SchedulerStorage, VTError};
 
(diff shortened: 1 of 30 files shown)


rust/src/scanner/scanner_stack.rs
```diff
@@ -15,16 +13,14 @@ pub trait Schedule: Iterator<Item = ConcurrentVTResult> + Sized {
 
 impl<T> Schedule for T where T: Iterator<Item = ConcurrentVTResult> {}
 
+// TODO: Remove this trait, now that it is just one associated type?
 pub trait ScannerStack {
     type Storage: ContextStorage + SchedulerStorage + Clone + 'static;
```
Tehforsch, line 18: This trait once unified `Loader`, `Storage` and `Scheduler` to reduce the amount of generics, but it has really lost its purpose by now. I didn't want to remove it in this PR, but I think this should be done at some point.

Kraemii (dismissed):

PR #2088, opened 2025-12-09
Add: log messages (results) for internally generated information

nichtsfrei:
Adds internal log messages with the key:
`openvasd/container_image_scanner`
within the oid field so that consumer can filter them.

The generated log messages include:
- found operating system
- errors

Notus erros as well as GZIP error will not fail a scan anymore.

https://jira.greenbone.net/browse/SC-1513

### DIFF
--- a/rust/src/openvasd/container_image_scanner/endpoints/scans.rs
+++ b/rust/src/openvasd/container_image_scanner/endpoints/scans.rs
@@ -890,12 +890,26 @@ mod test {
             let result = fakes.entry.get_scans_id_results(scan_id, None, None);
             let result: Vec<_> = result.collect().await;
 
-            let result: Vec<_> = result
-                .into_iter()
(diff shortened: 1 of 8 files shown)


rust/src/openvasd/container_image_scanner/endpoints/scans.rs
```diff
@@ -889,13 +889,28 @@ mod test {
             let (scan_id, _) = fakes.create_start_results(&client_id, scan).await;
             let result = fakes.entry.get_scans_id_results(scan_id, None, None);
             let result: Vec<_> = result.collect().await;
+            dbg!(&result);
```
jjnicola: Remove?

nichtsfrei:
Contained in: https://github.com/greenbone/openvas-scanner/pull/2092

PR #2089, opened 2025-12-10
Changes default for container-image-scanner parallel scans / batch size

nichtsfrei:
Changes the default amount of parallel scans within container-image-scanner to 23 and the batch size to 5.

Which means it could potentially open 23 * 5 concurrent connections.


nichtsfrei:
Ended up in: https://github.com/greenbone/openvas-scanner/pull/2088

PR #2090, opened 2025-12-10
Fix overflow on head request

jjnicola:
**What**:
- fix overflow: When request has no path, len() is 0 and usize type. Therefore, it is not possible to extract 1 and the request ends with `attempt to subtract with overflow`
- Fix: HEAD entry: `HEAD /`  method was not reached and `HEAD /scans` returns 200 instead of 204.
Jira: SC-1518

**Why**:

**How**:
request HEAD / or GET with an empty path

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/crates/greenbone-scanner-framework/src/entry/mod.rs
+++ b/rust/crates/greenbone-scanner-framework/src/entry/mod.rs
@@ -561,7 +561,7 @@ mod tests {
         );
         let req = test_utilities::empty_request(Method::HEAD, "/test////authn//////");
         let resp = entry_point.call(req).await.unwrap();
-        assert_eq!(resp.status(), StatusCode::NO_CONTENT);
+        assert_eq!(resp.status(), StatusCode::OK);
         let headers = resp.headers();
         assert_eq!(
             headers.get("authentication").unwrap(),
@@ -682,7 +682,7 @@ mod tests {
             .unwrap();
         let resp = entry_point.call(req).await.unwrap();
 
-        assert_eq!(resp.status(), StatusCode::NO_CONTENT);
+        assert_eq!(resp.status(), StatusCode::OK);
     }
 
     struct PrefixedAuth {}
@@ -725,7 +725,7 @@ mod tests {
             .unwrap();
         let resp = entry_point.call(req).await.unwrap();
 
-        assert_eq!(resp.status(), StatusCode::NO_CONTENT);
+        assert_eq!(resp.status(), StatusCode::OK);
     }
 
     #[tokio::test]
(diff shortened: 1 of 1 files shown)


nichtsfrei (approved):

rust/crates/greenbone-scanner-framework/src/entry/mod.rs
```diff
@@ -208,9 +214,6 @@ impl RequestHandlers {
                             rh.prefix(),
                             rh.path_segments().join("/")
                         );
-                        if req.method() == Method::HEAD {
```
nichtsfrei, line 211: There are tests to verify that HEAD only works on known paths. Either remove the test or move the HEAD check back to the loop.

nichtsfrei (changes requested):

rust/crates/greenbone-scanner-framework/src/entry/mod.rs
```diff
@@ -228,6 +235,9 @@ impl RequestHandlers {
                     }
                 }
             }
+            if req.method() == Method::HEAD && !has_match {
+                return BodyKind::no_content(StatusCode::OK);
```
nichtsfrei: I'm a bit confused about that. Isn't that already handled in the loop? Why this has_match variable which is only true when a segment watches anyway? Also the return code differs which seems incorrect as well?

rust/crates/greenbone-scanner-framework/src/entry/mod.rs
```diff
@@ -208,8 +213,10 @@ impl RequestHandlers {
                             rh.prefix(),
                             rh.path_segments().join("/")
                         );
-                        if req.method() == Method::HEAD {
-                            return BodyKind::no_content(StatusCode::OK);
+                        if req.method() == Method::HEAD && is_authenticated {
+                            return BodyKind::no_content(StatusCode::NO_CONTENT);
+                        } else if req.method() == Method::HEAD && !is_authenticated {
+                            return BodyKind::no_content(StatusCode::UNAUTHORIZED);
```
nichtsfrei: isn't the unauthorized case already handled in the else case of !needsauthentication || is_authenticated?

rust/crates/greenbone-scanner-framework/src/entry/mod.rs
jjnicola: The loop doesn't handle `HEAD /` (without path). [This case](https://greenbone.github.io/scanner-api/#/general/get_info)
For `HEAD /scans`, it is indeed handled inside loop with two different codes.
With this patch, openvasd responds with the codes following the specs

rust/crates/greenbone-scanner-framework/src/entry/mod.rs
jjnicola: you are right! fixed with [e8d5653](https://github.com/greenbone/openvas-scanner/pull/2090/commits/e8d5653ae2162c5a50e13c0925e8faa275ab0874)

PR #2129, opened 2026-02-09
Builtin notus

Kraemii:
Implements:
- notus_error
- notus
- security_notus

Jira: SC-1560

Run test script with `scannerctl execute script -n ~/var/lib/notus/products/ notus.nasl`
```c#
pkg_list = "libc6-1.28-10+deb10u2, libc-dev-bin-2.28-10+deb10u2";

results = notus(product: "debian_10", pkg_list: pkg_list);

if(!isnull(results)) {
    foreach res (results) {
        oid = res["oid"];
        msg = res["message"];
        display(oid);
        display(msg);
    }
} else {
    ret = notus_error();
    display(ret);
}
```
Just play a little bit around.

### DIFF
--- a/rust/src/nasl/builtin/notus/mod.rs
+++ b/rust/src/nasl/builtin/notus/mod.rs
@@ -8,6 +8,7 @@ mod tests;
 use std::{collections::HashMap, net::SocketAddr};
 
 use greenbone_scanner_framework::models::FixedVersion;
+use http::StatusCode;
 use nasl_function_proc_macro::nasl_function;
 use serde::{Deserialize, Serialize};
 use serde_json;
@@ -75,6 +76,14 @@ impl NaslNotus {
             .send()
             .await
             .map_err(|e| HttpError::Custom(e.to_string()))?;
+
+        if response.status() != StatusCode::OK {
+            return Err(HttpError::Custom(format!(
+                "Notus service returned status code {}",
+                response.status()
+            ))
+            .into());
+        }
 
         // Parse JSON array of results
         let results: Vec<NotusResult> = response


rust/src/nasl/builtin/notus/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+        &self,
+        addr: &SocketAddr,
+        pkg_list: &[String],
+        product: &str,
+    ) -> Result<NaslValue, FnError> {
+        let mut sock = TcpStream::connect(addr).map_err(|e| HttpError::IO(e.kind()))?;
+        let pkg_json = serde_json::to_string(pkg_list).unwrap();
+
+        let request = format!(
+            "POST /notus/{} HTTP/1.1\r\nContent-Length: {}\r\n\r\n{}",
```
nichtsfrei: why don't use hyper::client (used by us) or reqwest (used by docker-registry)? 

This seems very low level and could get annoying when TLS is used or if http2 is used (like we do) and so on. Also status code verification is missing. I think a higher level abstraction would be better although I appreciate the low levelness personally :)

rust/src/nasl/utils/scan_ctx.rs
```diff
@@ -266,6 +267,13 @@ impl<T> ContextStorage for RedisStorage<T> where
 }
 impl<T> ContextStorage for Arc<T> where T: ContextStorage {}
 
+#[derive(Clone)]
+pub enum NotusCtx {
+    Direct(Arc<Mutex<Notus<HashsumProductLoader>>>),
+    Address(SocketAddr),
+    // We might want to add more contexts here in the future, e.g. for other plugins.
```
nichtsfrei: isn't that normal for an enum? :)

rust/src/scannerctl/execute/mod.rs
```diff
@@ -5,21 +5,23 @@
 use std::fs::File;
 use std::io::stdin;
 use std::path::PathBuf;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
```
nichtsfrei, line 8: mhm I'm wondering if you introduce a new mutex and since we are using it in an async context if it wouldn't be better to introduce tokios variant?

rust/src/scannerctl/interpret/mod.rs
```diff
@@ -123,6 +124,7 @@ fn load_feed_by_json(store: &InMemoryStorage, path: &PathBuf) -> Result<(), CliE
     Ok(())
 }
 
+#[allow(clippy::too_many_arguments)]
```
nichtsfrei, line 128: Please add a TODO that we should redesign that.

rust/src/scannerctl/interpret/mod.rs
```diff
@@ -208,7 +214,17 @@ pub async fn run(
             } else {
                 load_feed_by_exec(&storage, &loader).await?
             }
-            run_on_storage(storage, loader, target, kb, ports, script, scan_preferences).await
+            run_on_storage(
```
nichtsfrei, line 218: Because this screams 'confusion' to me :D, currently it is fine because they variables are readable but I don't like that this method call is only understandable if the variable names follow the function naming convention.

rust/src/nasl/builtin/notus/mod.rs
```diff
@@ -0,0 +1,167 @@
...
+        Ok(NaslValue::Array(ret))
+    }
+
+    #[nasl_function]
+    fn notus_error(&self) -> Option<String> {
+        self.last_error.clone()
+    }
+
+    #[nasl_function(named(pkg_list, product))]
+    fn notus(
```
nichtsfrei: I think those should be async, although we currently allow sync functions I remember it only be the case because of older functions but we should use async whenever possible to make transition easier. And also allows us to use more abstract IO functionality.

rust/src/scannerctl/execute/mod.rs
Kraemii: I did a little research: If we do not do any async operations (.await) with the Data behind a Mutext, it is best to just stick to the `std::sync::Mutex` https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#which-kind-of-mutex-should-you-use
As the internal Notus implementation, as well as the Socket Data are not used for an async operation I think it would be better to stick to the `std::sync::Mutex`

rust/src/nasl/builtin/notus/mod.rs
Kraemii: I use reqwest instead now

nichtsfrei (approved):

PR #2133, opened 2026-02-12
Implement Kerberos in Rust NASL using existing C code

Kraemii:
In the first step I just did a tidy up of the build scripts already in use for building C code, we already depend on. This includes:
- Move all build artifacts to target directory, so they will be cleaned with a `cargo clean`
- Remove shell script for building `libgcrypt` and `libgpg-error`. Instead building it within `build.rs`
- Build c files in the sys crate
- Provide bindings in sys crate
- Re-export sys crate in nas-c-lib to make bindings available

Next I implemented libopenvas-krb5-sys, which includes a build script to:
- Download krb5 source
- Build krb5 libs from source
- Compile openvas-krb5 implementation
- Generate bindings

Finally I added all Nasl builtin functions for Kerberos:
- krb5_error_code_to_string
- krb5_find_kdc
- krb5_gss_init
- krb5_gss_prepare_context
- krb5_gss_session_key
- krb5_gss_update_context
- krb5_gss_update_context_needs_more
- krb5_gss_update_context_out
- krb5_is_failure
- krb5_is_success

Jira: SC-1136

For testing I provide a small and simple test script. Please use environment variables to fill all the necessary parameters for the Krb5 credentials. Here a template:
```
KRB5_CONFIG=/etc/openvasd/krb5.conf \
    KRB5_KDC=KDC \
    KRB5_TARGET_HOST=Hostname \
    KRB5_REALM=REALM \
    KRB5_USER=User \
    KRB5_PASSWORD=Password \
    target/release/scannerctl execute script -t <IP> examples/feed/krb5/login.nasl
```

### DIFF
--- a/rust/crates/nasl-c-lib/libopenvas-krb5-sys/build.rs
+++ b/rust/crates/nasl-c-lib/libopenvas-krb5-sys/build.rs
@@ -6,8 +6,7 @@ use std::env;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 
-const KRB5_VERSION: &str = "1.20";
-const KRB5_PATCH: &str = "1";
+const KRB5_VERSION: &str = "1.22.2";
 
 fn main() {
     println!("cargo:rerun-if-changed=build.rs");
@@ -51,7 +50,12 @@ fn main() {
 }
 
 fn fetch_krb5(destination: &Path) -> PathBuf {
-    let folder_name = format!("krb5-{}.{}", KRB5_VERSION, KRB5_PATCH);
+    let krb5_version = env::var("KRB5_VERSION").unwrap_or_else(|_| KRB5_VERSION.to_string());
+    let (krb5_version, krb5_patch) = krb5_version.rsplit_once('.').expect(&format!(
+        "Invalid KRB5_VERSION format: expected major.minor.patch, got {}",
+        krb5_version
+    ));
+    let folder_name = format!("krb5-{}.{}", krb5_version, krb5_patch);
     let extract_path = destination.join(&folder_name);
     let tar_path = extract_path.with_added_extension("tar.gz");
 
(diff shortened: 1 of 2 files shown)


rust/crates/nasl-c-lib/libgcrypt-sys/build.rs
```diff
@@ -2,34 +2,76 @@
...
+    if let Ok(val) = env::var("IN_CROSS")
+        && val != "0"
+        && target == "aarch64-unknown-linux-gnu"
+    {
+        unsafe {
+            env::set_var("CC", "aarch64-linux-gnu-gcc");
+            env::set_var("CHOST", "arm64");
         }
-        Err(e) => panic!("Unexpected error: {e}"),
     }
```
Kraemii: I am not feeling well about this part. I think for cross compiling it would be better, that `CC` and `CHOST` is set accordingly by the one building this. I think this will be very error prone.

rust/crates/nasl-c-lib/libcrypt-sys/build.rs
```diff
@@ -2,34 +2,76 @@
...
+                .status()
+                .unwrap();
+        }
+
+        Command::new("tar")
+            .args(["-xf", &tar_path.to_string_lossy()])
+            .current_dir(destination)
+            .status()
+            .unwrap();
+    }
```
Kraemii, line 67: Instead of using Command calls it would also be possible to use Rust libraries handling this. What is preferred here?

rust/crates/nasl-c-lib/libopenvas-krb5-sys/build.rs
```diff
@@ -0,0 +1,116 @@
...
+        ))
+        .current_dir(src)
+        .status()
+        .expect("Failed to run configure")
+        .success()
+    {
+        panic!("Configure failed");
+    }
+
+    // Ignore status, as it make process might fail, but libs will still be built
```
nichtsfrei: Isn't that comment missleading as expect will fail?

rust/src/nasl/builtin/krb5/mod.rs
```diff
@@ -0,0 +1,523 @@
...
+        user: String,
+        message: String,
+        code: OKrb5ErrorCode,
+    },
+    #[error("Unable to prepare GSS context")]
+    Context,
+}
+
+struct Krb5Credentials {
+    _config_path: String,
```
nichtsfrei, line 80: why starting with _?

rust/crates/nasl-c-lib/libopenvas-krb5-sys/build.rs
Kraemii: Not exactly, the expect just checks if running the command lead to an error, that does not produce an return code, like `command not found`, `permission denied` and so on. In case the make comnmand actually runs, but produces errors and does not exit with a 0, we just ignore that, as that is even expected here. But I can also just delete the comment.

rust/src/nasl/builtin/krb5/mod.rs
Kraemii: As the created `OKrb5Credential` type does only store pointers to existing data, I have to keep these in memory. They are not actually used anyway and are automatically freed, when dropped.

PR #2139, opened 2026-02-23
Fix: Use newer product/protocol names in some service reportings.

cfi-gb:
**What**:
- `MySQL` got changed to `MySQL/MariaDB` as the relevant code is actually already covering MariaDB via `|| strstr (buffer, "mariadb") != NULL)))`
- `Postgres` is called `PostgreSQL` these days and one text part `A PostgreSQL server is running on this port` already covered that
- `SSL` got changed to `SSL/TLS` as the `switch` / `case` call already covered both SSL 2.0 through SSL 3.0 as well as TLS 1.0 through TLS 1.3 and we should reflect this

**Why**:
Better / more up2date naming.

**How**:
N/A

**Checklist**:

- [ ] Tests

### DIFF
--- a/misc/network.c
+++ b/misc/network.c
@@ -1778,7 +1778,7 @@ get_encaps_through (openvas_encaps_t code)
     case OPENVAS_ENCAPS_TLSv12:
     case OPENVAS_ENCAPS_TLSv13:
     case OPENVAS_ENCAPS_TLScustom:
-      return " through SSL";
+      return " through SSL/TLS";
     default:
       snprintf (str, sizeof (str),
                 " through unknown transport layer - code %d (0x%x)", code,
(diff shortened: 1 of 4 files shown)


ArnoStiefvater:
Hey @cfi-gb, thanks a lot for the PR.

Will users on SCAN/GEA/GSA run into problems because delta of reports will differ or does this not happen or is it irrelevant? This was the only thing that came to mind regarding this PR by us.

cfi-gb:
Hey,

> Hey @cfi-gb, thanks a lot for the PR.
> 
> Will users on SCAN/GEA/GSA run into problems because delta of reports will differ or does this not happen or is it irrelevant? This was the only thing that came to mind regarding this PR by us.

thanks a lot for the question / remark. AFAICT / from the code these are only messages within reports with a "log level" result and i can't find any references that something is trying to parse them somehow.

So the only difference a user will see is the improved messages. This for sure caused a "changed" result in the delta reports but this is something which is happening daily for VT maintenance work and users should be familiar with it.

:+1:

PR #2150, opened 2026-03-06
Add: Create krb5 config per IP

Kraemii:
The default for the krb5 config was /etc/krb5.conf which. This now changes and for each IP a separate config is created in /tmp/krb5_<ip>.conf. These are also cleaned up.

Jira: SC-1284

### DIFF
--- a/src/openvas.c
+++ b/src/openvas.c
@@ -26,7 +26,6 @@
 #include "../misc/plugutils.h"     /* nvticache_free */
 #include "../misc/scan_id.h"       /* to manage global scan_id */
 #include "../misc/vendorversion.h" /* for vendor_version_set */
-#include "../nasl/nasl_krb5.h"     /* for nasl_okrb5_clean */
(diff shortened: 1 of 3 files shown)


src/openvas.c
```diff
@@ -641,7 +640,6 @@ openvas (int argc, char *argv[], char *env[])
 
       gvm_close_sentry ();
       destroy_scan_globals (globals);
-      nasl_okrb5_clean ();
```
nichtsfrei, line 644: Don't we have to cleanup krb stuff when the scan is done?

nichtsfrei: nevermind it just moved.

nichtsfrei (approved):

src/openvas.c
Kraemii: Yeah to wrap it up, we have to clean it up per script fork, not per scan. So it moved to the cleanup op scripts.

PR #2153, opened 2026-03-10
Fix pending image scheduling to favor least-progressed scans

nichtsfrei:
Previously, pending image downloads were selected in FIFO order, which could favor already running scans when multiple scans were started.

Change the SQLite pending-image query to order by host_finished so images from the least-progressed scans are picked first. This makes image scheduling fairer across concurrent scans.

### DIFF
--- a/rust/src/openvasd/container_image_scanner/scheduling/db/sqlite/images.rs
+++ b/rust/src/openvasd/container_image_scanner/scheduling/db/sqlite/images.rs
@@ -118,14 +118,14 @@ impl<'o> Execute<Vec<(ImageID, Option<Credential>)>> for DBImages<'o, (usize, us
                 max => max as i64,
             };
 
-            // TODO: add ordering here: we should prioritize images that belong to a scan with the
-            // least amount of finished on scan.
             let rows = sqlx::query(
                 r#"
     SELECT i.id, i.image, c.username, c.password
     FROM images i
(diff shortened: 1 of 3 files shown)


rust/src/openvasd/container_image_scanner/scheduling/db/sqlite/images.rs
```diff
@@ -239,3 +239,95 @@ impl<'o> Fetch<Vec<ProcessingImage>> for DBImages<'o, usize> {
...
+        };
+
+        dbg!(&ids);
+        validate(3, ids.clone()).await;
+        ids.remove(1);
+        dbg!(&ids);
+        validate(2, ids.clone()).await;
+        ids.remove(1);
+        dbg!(&ids);
+        validate(5, ids.clone()).await;
```
Tehforsch, line 331: my no "`dbg!()` statements in the codebase"-pre-commit-script will cry. but its fine, it already cries anyways

Tehforsch (approved):

rust/src/openvasd/container_image_scanner/scheduling/db/sqlite/images.rs
nichtsfrei: Unconditional solidarity with dbg!


I can remove them if you like, but I actually like them in test cases though. Makes debugging a bit easier.

PR #2155, opened 2026-03-11
Fix: insert alive test methods and ports

jjnicola:
**What**:
insert alive test methods and ports to scans table
Jira: SC-1579

**Why**:
Are not inserted and the default `icmp` methods is always used

**How**:
run a scan with a different method 

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/src/openvasd/database/sqlite/scans.rs
+++ b/rust/src/openvasd/database/sqlite/scans.rs
@@ -140,14 +140,13 @@ where
                     port.range
                         .clone()
                         .into_iter()
-                        .map(move |r| (port.protocol, r))
+                        .map(move |r| (port.protocol.as_ref(), r))
                 }),
                 |mut b, (protocol, range)| {
                     b.push_bind(&mapped_id)
                         .push_bind(match protocol {
                             None => "udp_tcp",
-                            Some(models::Protocol::TCP) => "tcp",
-                            Some(models::Protocol::UDP) => "udp",
+                            Some(x) => x.as_ref(),
                         })
                         .push_bind(range.start as i64)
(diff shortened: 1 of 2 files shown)


rust/src/openvasd/database/sqlite/scans.rs
```diff
@@ -158,6 +158,54 @@ where
...
+
+            query.execute(&mut *tx).await?;
+        }
+    }
+
+    if !scan.target.alive_test_methods.is_empty() {
+        let mut builder = QueryBuilder::new("INSERT INTO alive_methods (id, method)");
+        builder.push_values(&scan.target.alive_test_methods, |mut b, method| {
+            b.push_bind(&mapped_id).push_bind(match method {
+                AliveTestMethods::TcpAck => "tcp_ack",
```
nichtsfrei: Wouldn't it be better to implement `AsRef<str>` for AliveTestMethods? Or you could use `.to_string()` would be a bit of a waste, but I think better than reimplementing it here?

rust/src/openvasd/database/sqlite/scans.rs
```diff
@@ -158,6 +158,54 @@ where
...
+                    port.range
+                        .clone()
+                        .into_iter()
+                        .map(move |r| (port.protocol, r))
+                }),
+                |mut b, (protocol, range)| {
+                    b.push_bind(&mapped_id)
+                        .push_bind(match protocol {
+                            None => "udp_tcp",
+                            Some(models::Protocol::TCP) => "tcp",
```
nichtsfrei: I think for the some cases it would be better to either implement `AsRef<str>` for Protocol or use `.to_string()`. The first one would be better, the second one is already implemented.

PR #2160, opened 2026-03-12
Fix: version comparison for RPM

Kraemii:
In case two RPM versions are compared and have a different separator they are checked lexicographical. In reality they should not be compared and are interchangeable. To solve issues with this, we now normalize the version string to only contain a single '.' as a separator.

SC-1572

### DIFF
--- a/rust/src/notus/packages/rpm.rs
+++ b/rust/src/notus/packages/rpm.rs
@@ -65,7 +65,7 @@ static RE_VERSION: Lazy<Regex> = lazy_regex!(
     )?$"
 );
 
-/// Represent a based Redhat package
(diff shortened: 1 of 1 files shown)


rust/src/notus/packages/rpm.rs
```diff
@@ -87,6 +87,23 @@ fn find_any_exception(name: &str) -> String {
     "".to_string()
 }
 
+fn normalize(version: &str) -> String {
```
nichtsfrei, line 90: Is that true for all RPM spec or just Redhat?

Kraemii: I have the information from the official documentation for rpm: https://rpm.org/docs/6.0.x/man/rpm-version.7

PR #2163, opened 2026-03-13
Add: Host discovery for large IPv6 networks

jjnicola:
**What**:
With this patch, the host discovery for large ipv6 networks is introduced together with a new alive test method (int 32).
Instead of creating a host list from an IPv6 network address and mask (e.g. 5858::0/64), which would take years to scan and it would consume all the RAM, it just call a new function in gvm-libs to perform an ICMPv6 to a multicast IPv6 address (ff02::1).
Later, the found host are set as target and they will be handle as before. To avoid a double alive test, they are just considered alive.

For now, it only supports only one IPv6 network address. This means it can not be combined with other networks or unicast addresses.

Also, it is required that the local network has ICMPv6 enabled, otherwise the discovery won't work.

Jira: SC-1574

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/src/alive_test/alive_test.rs
+++ b/rust/src/alive_test/alive_test.rs
@@ -4,7 +4,9 @@
 
 use crate::alive_test::AliveTestError;
 use crate::alive_test::arp::forge_arp_request;
-use crate::alive_test::icmp::{forge_icmp_v4, forge_icmp_v6, forge_neighbor_solicit};
+use crate::alive_test::icmp::{
+    forge_icmp_v4, forge_icmp_v6, forge_icmp_v6_for_host_discovery, forge_neighbor_solicit,
+};
 use crate::nasl::raw_ip_utils::{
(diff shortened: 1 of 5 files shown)


jjnicola:
It depends on https://github.com/greenbone/gvm-libs/pull/1058

rust/src/alive_test/alive_test.rs
```diff
@@ -231,13 +235,51 @@ async fn capture_task(
...
+    if methods.contains(&AliveTestMethods::HostDiscoveryIpv6) {
+        if let Some(t) = target.iter().next()
+            && target.len() == 1
+        {
+            return Some(
+                Ipv6Cidr::from_str(t)
+                    .map_err(|e| AliveTestError::InvalidDestinationAddr(e.to_string())),
+            );
+        } else {
+            tracing::debug!("Only one IPv6 network address is allowed for host discovery");
```
Kraemii: I thing a warn would be better here

PR #2170, opened 2026-03-17
Async storage

Tehforsch:
Sorry for the massive PR, but this PR is basically the minimum set of changes that are demanded by putting async in front of retrieve, dispatch and remove.

To make my life easier, I removed all unnecessary trait bounds that we demanded of various types using the storage traits (`ContextStorage` in particular), so I could keep the trait bounds that needed to be added to a minimum.

Jira: SC-1511

### DIFF
--- a/rust/src/nasl/utils/scan_ctx.rs
+++ b/rust/src/nasl/utils/scan_ctx.rs
@@ -247,7 +247,7 @@ pub trait ContextStorage:
     /// KbKey, the value is appended to the existing list. This function is used to replace the
     /// existing entry with the new one.
     async fn dispatch_replace(&self, key: KbContextKey, item: KbItem) -> Result<(), StorageError> {
-        self.remove(&key)?;
+        self.remove(&key).await?;
         self.dispatch(key, item).await
     }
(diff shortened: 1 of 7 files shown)


rust/src/nasl/utils/scan_ctx.rs
```diff
@@ -618,7 +648,11 @@ impl Drop for ScanCtx<'_> {
     fn drop(&mut self) {
         let mut nvt = self.nvt.lock().unwrap();
         if let Some(nvt) = nvt.take() {
-            self.dispatch_nvt(nvt);
+            // TODO: This might very well not work.
+            // Do we really need this Drop impl anyways?
+            futures::executor::block_on(async {
+                self.dispatch_nvt(nvt).await;
+            });
```
Tehforsch, line 666: This is a complicated issue, I would have liked to remove this drop impl but that requires rethinking the entire logic about setting the nvt fields in the ScanCtx (which arguably should be done in the ScriptCtx instead?) and it got too big for this already big PR. The current solution works, although it is ugly.

PR #2176, opened 2026-03-24
Kerberos CCache

Kraemii:
This PR enables the CCache. It is set to the location `/tmp/krb5cc_<ip>` and the env variable `KRB5CCNAME` is set accordingly. The default can be overwritten, if the env variable is set manually.

SC-1509

### DIFF
--- a/nasl/nasl_krb5.c
+++ b/nasl/nasl_krb5.c
@@ -106,6 +106,12 @@ build_krb5_credential (lex_ctxt *lexic)
       snprintf (default_config_path, sizeof (default_config_path),
                 "/tmp/krb5_%s.conf", ip_str);
       setenv ("KRB5_CONFIG", default_config_path, 1);
+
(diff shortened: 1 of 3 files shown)


nasl/nasl_krb5.c
```diff
@@ -106,6 +106,12 @@ build_krb5_credential (lex_ctxt *lexic)
       snprintf (default_config_path, sizeof (default_config_path),
                 "/tmp/krb5_%s.conf", ip_str);
```
ArnoStiefvater: read out the variable first
don't hardcode /tmp

Kraemii: `/tmp/...` is only the default value. It can be overwritten by either the env variable or within the nasl script when calling `krb5_gss_prepare_context`. I could do the same for the CCache

nichtsfrei (approved):

PR #2181, opened 2026-03-26
Fix: set and show openvasd version at start

jjnicola:
**What**:
set and show openvasd version at start
SC-1405

**Why**:
to show it

**How**:
Start openvasd and read the log

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/src/openvasd/config/mod.rs
+++ b/rust/src/openvasd/config/mod.rs
@@ -607,7 +607,6 @@ impl Config {
             )
             .arg(
                 clap::Arg::new("version")
-                    .short('V')
                     .action(ArgAction::SetTrue)
                     .help("Show openvasd version and exit."),
             )


rust/src/utils/version.rs
```diff
@@ -0,0 +1,10 @@
+// SPDX-FileCopyrightText: 2026 Greenbone AG
+//
+// SPDX-License-Identifier: GPL-2.0-or-later WITH x11vnc-openssl-exception
+
+pub fn show_version(binary_name: &str) {
+    tracing::info!(
+        "Running {binary_name} version {}",
+        option_env!("VERGEN_GIT_DESCRIBE").unwrap_or("unknown")
```
nichtsfrei, line 8: doesn't that just show whatever is in the evn? Like if  i call VERGEN_GIT_DESCRIBE="hallo" openvasd -v it shows 'hallo'?

rust/src/openvasd/config/mod.rs
```diff
@@ -686,6 +694,12 @@ impl Config {
         if let Some(enable) = cmds.get_one::<bool>("enable-get-scans") {
             config.endpoints.enable_get_scans = *enable;
         }
+        if let Some(version) = cmds.get_one::<bool>("version") {
+            config.version = *version;
+        }
+        if let Some(version) = cmds.get_one::<bool>("version") {
```
nichtsfrei: duplicate?

rust/src/openvasd/config/mod.rs
```diff
@@ -603,6 +605,12 @@ impl Config {
                     .short('L')
                     .help("Level of log messages to be shown. TRACE > DEBUG > INFO > WARN > ERROR"),
             )
+            .arg(
```
nichtsfrei, line 608: Do we need that as an argument? I think printing it startup is sufficient?

rust/src/openvasd/config/mod.rs
jjnicola: ups

rust/src/openvasd/config/mod.rs
jjnicola: I like that binaries can show the version without having to start a daemon.

nichtsfrei: mhm, ok, but wouldn't be long form be sufficient? Like -V is a bit unusual and --version should be fine?

jjnicola: removed -V short arg with https://github.com/greenbone/openvas-scanner/pull/2181/commits/85986fde22fbfda2bfa9f060f87c9764c0af8b8f

PR #2182, opened 2026-03-30
Simplify Notus product loading and relax feed requirements

nichtsfrei:
Refactor Notus loading by unifying product handling into a single ProductLoader and removing legacy loader implementations, caching, and implicit verification logic.

The main goal is to allow using Notus feeds without requiring a sha256sums file, which previously made testing and local development unnecessarily difficult.

At the same time, the default is now to have signature_check enabled. When it is enabled, behavior remains unchanged and feeds are still verified as before. Verification is now optional and handled explicitly, resulting in a simpler, more predictable, and flexible loading pipeline.

SC-1581

### DIFF
--- a/rust/src/notus/notus.rs
+++ b/rust/src/notus/notus.rs
@@ -36,13 +36,13 @@ impl Notus {
     fn load_product(&self, os: &str) -> Result<Product, Error> {
         let product = self.loader.load_product(os)?;
 
-        match Product::try_from(product) {
-            Ok(adv) => Ok(adv),
-            Err(Error::VulnerabilityTestParseError(_, pkg)) => {
-                Err(Error::VulnerabilityTestParseError(os.to_string(), pkg))
+        Product::try_from(product).map_err(|e| {
+            if let Error::VulnerabilityTestParseError(_, pkg) = e {
+                Error::VulnerabilityTestParseError(os.to_string(), pkg)
+            } else {
+                e
             }
-            Err(err) => Err(err),
-        }
+        })
     }
 
     fn parse<P: Package>(packages: &[String]) -> Result<Vec<P>, Error> {


rust/examples/GBCommunitySigningKey.asc
```diff
@@ -0,0 +1,29 @@
...
+dvIXivKLVDV4D7Z+QGU+140mYaIRfKagzfGzcke/3IC9ppUATA0R1r0l4rmyD6bF
+v9JnS41hSL4HZCH9LJaaPN9k7fRpSmUHyn05+2iW6UIz1aona6Y3nRvQ2XZxsg7u
+6mEjEou9TD3ayVt/uewEQ2mHxr07YHYq/jZFHXTRz5R8fPwUu1NpSL+zI6CFf6cx
+Et4bcRLeW7CD4/7UcUXGp+SECvj4/IeVkZABjxpC9ikLX2FkgJD1G6/gt3eEjmEF
+2g8l9tYDIJqclm1gpQk6TMRuw43Rqf1Vhd9USGBXK2lTo6ZU3LYUQzz2ycYmF8Ut
+x693sUpNCCWgEV3k21O8lDeUbrtrBdEcOYbbG73mV4UjLKxM0dJ8V8j0CDRprzD8
+kFeelLFq/j6iglHMVXhQQX5nTc0ieNlxOj2O7HFu1yR/B1CIBT8H0ab2aJjEsZPh
+hq67V5EJ4c1l3hOJIEyNdkG90uEHGKFABlQ9TaVjCDlGI5LKmNYb+bq39QARAQAB
+tCZHcmVlbmJvbmUgQ29tbXVuaXR5IEZlZWQgaW50ZWdyaXR5IGtleYkCTgQTAQgA
+OBYhBIrkvkKbYKWbMRwuc5gj+qYO0eWABQJZsAVfAhsDBQsJCAcCBhUICQoLAgQW
```
Tehforsch, line 15: Should this really be `OBYhBIrkvkKbYKWbMRwuc5gj`? I think we should rewrite it as `OCYhBIrkvkKbYKWbMRwuc5gk`.

rust/src/notus/notus.rs
```diff
@@ -25,39 +23,21 @@ use super::{
...
-        let (product, stamp) = self.loader.load_product(os)?;
+    fn load_product(&self, os: &str) -> Result<Product, Error> {
+        let product = self.loader.load_product(os)?;
 
         match Product::try_from(product) {
-            Ok(adv) => Ok((adv, stamp)),
+            Ok(adv) => Ok(adv),
             Err(Error::VulnerabilityTestParseError(_, pkg)) => {
                 Err(Error::VulnerabilityTestParseError(os.to_string(), pkg))
             }
```
Tehforsch: This is `map_err` now, but whatever

rust/src/scannerctl/execute/mod.rs
```diff
@@ -165,8 +165,8 @@ async fn script(args: ScriptArgs) -> Result<(), CliError> {
     let notus = args.notus.map(|x| match x {
         NotusArgs::Address(addr) => NotusCtx::Address(addr),
         NotusArgs::Internal(path) => NotusCtx::Direct(Arc::new(Mutex::new(Notus::new(
```
Tehforsch, line 167: Not important: Making this mutex/arc wrapping a `direct` method on `NotusCtx` would look a little cleaner, but whatever

nichtsfrei: true, but the only usages are within scannerctl and a test, so, dunno. Coming up with a name does not seem to be worth the trouble?

PR #2184, opened 2026-04-01
Add: ip_reverse_lookup() nasl builtin rust implementation

jjnicola:
**What**:
Add: ip_reverse_lookup() nasl builtin rust implementation
SC-1582

**Why**:
missing and used in the feed

**How**:
Run the following NASL script with scannerctl and openvas-nasl. Compare the output
```
$ cat ip_reverse_lookup.nasl 
display(ip_reverse_lookup("192.168.1.1"));
display(ip_reverse_lookup());
```
```
openvas-nasl -X ip_reverse_lookup.nasl # without target
openvas-nasl -X ip_reverse_lookup.nasl -t 192.168.1.1 #has hostname
openvas-nasl -X ip_reverse_lookup.nasl -t 192.168.1.50 #does not have a hostname

target/debug/scannerctl execute script ~/my_nasl/ip_reverse_lookup.nasl
target/debug/scannerctl execute script ~/my_nasl/ip_reverse_lookup.nasl -t 192.168.1.1
target/debug/scannerctl execute script ~/my_nasl/ip_reverse_lookup.nasl -t 192.168.1.50
```

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/src/nasl/builtin/host/mod.rs
+++ b/rust/src/nasl/builtin/host/mod.rs
@@ -4,8 +4,8 @@
 
 #[cfg(test)]
 mod tests;
-
 use dns_lookup::lookup_addr;
+use std::{net::IpAddr, str::FromStr};
 use thiserror::Error;
 
 use crate::nasl::interpreter::{Fork, ForkKind};
@@ -20,6 +20,8 @@ pub enum HostError {
     EmptyAddress,
     #[error("Target is not a hostname.")]
     TargetIsNotAHostname,
+    #[error("{0}")]
+    InvalidHost(String),
 }
 
 struct Hostname(String);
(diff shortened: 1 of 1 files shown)


rust/src/nasl/builtin/host/mod.rs
```diff
@@ -172,6 +172,18 @@ fn same_host(h1: &str, h2: &str, cmp_hostname: Option<bool>) -> Result<bool, FnE
...
+/// Return the hostname of the given ip or the target if given or localhost as default
+#[nasl_function]
+fn host_reverse_lookup(context: &ScanCtx, ip_str: Option<String>) -> Option<String> {
+    let target = if let Some(t) = ip_str {
+        IpAddr::from_str(&t).ok()
+    } else {
+        Some(context.target().ip_addr())
+    };
+    //safe to unwrap the target, since it always defaults to localhost
+    lookup_addr(&target.unwrap()).ok()
```
Tehforsch: Not sure I agree with the logic here - `target` can be `None` if `from_str` returns `Err` (if the user calls the fn with a wrong ip addr str). In this case we should throw a proper error instead of panicking. So maybe a better solution would be to do something like (didn't check if this compiles)

```rust
fn host_reverse_lookup(context: &ScanCtx, ip_str: Option<String>) -> Result<Option<String>, FnError> {
    let target = if let Some(ip_str) = ip_str { {
        IpAddr::from_str(&t)?
    } else {
        context.target().ip_addr()
    };
    Ok(lookup_addr(&target).ok())
}
```

Tehforsch (changes requested):

Tehforsch (approved):
Thanks!

PR #2198, opened 2026-05-04
Tidy up

Kraemii:
Remove some unused stuff:

- `Cross.toml` is not used anymore
- the `scanner-migration` stuff is now in the `migration` directory
- the `openvasd-tests`, which were broken anyway


Tehforsch (approved):
Thanks

PR #2207, opened 2026-05-11
Change: Various `.md` documentation updates.

cfi-gb:
**What**:
- Adjust examples for script require_udp_ports() and script_require_ports() to reflect the correct strings
- Add hint on non_simult_ports to script_require_ports() docs

**Why**:
Better documentation (e.g. using `ssh` together with `443` might be confusing, or just `sip` which doesn't exist).

**How**:
N/A

**Checklist**:
- [ ] Tests


Tehforsch (approved):
Thank you!

PR #2209, opened 2026-05-12
Fix: cached hashes

jjnicola:
**What**:
Fix: cached hashes
SC-1597

**Why**:
If the integrity check is disabled, the cached_hashes() always returns None, which produces that the feeds is update each time that the scheduler runs.

**How**:
run openvasd with `integrity_check = false` and check with `GET /health/ready` that the feed version flips from a hash to unavailable each time that the scheduler runs and checks the feed.

**Checklist**:

- [ ] Tests


Tehforsch (approved):
Thank you!

PR #2211, opened 2026-05-13
Fix: hide password on GET /scans/{id} response

jjnicola:
**What**:
Fix: hide password on GET /scans/{id} response
SC-1598

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/rust/crates/greenbone-scanner-framework/src/models/credential.rs
+++ b/rust/crates/greenbone-scanner-framework/src/models/credential.rs
@@ -151,3 +151,88 @@ impl AsRef<str> for CredentialType {
         }
     }
 }
+
+impl CredentialType {
+    pub fn hide_pass(&mut self) -> Self {
+        match self {
+            CredentialType::UP {
+                username,
+                privilege,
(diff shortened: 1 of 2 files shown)


rust/crates/greenbone-scanner-framework/src/models/credential.rs
```diff
@@ -123,6 +140,7 @@ pub enum CredentialType {
         /// The SNMP username.
         username: String,
         /// The SNMP password.
+        #[serde(serialize_with = "hide_pass")]
```
nichtsfrei: did you verify that does  not interfere with the storage and usage of credentials within container-imager-scanner and openvas scanner?

as far as I remember we use serde to store the credentials in the DB with this change it would store the stars instead of the password we need later on in the process.

jjnicola: Yes. Still working on this PR. The test pops up the issue.

nichtsfrei (approved):

PR #2212, opened 2026-05-16
openvas/pref_handler: skip VTs whose redis entry fails UTF-8 decode

arch-fan:
This PR fixes a panic in `process_vts` caused by calling `.unwrap()` on `redis_connector.get_vt()`.

With a full Greenbone Community Feed loaded via `openvas --update-vt-info`, some VT entries can fail UTF-8 conversion when read back from Redis:

```text
LibraryError("Cannot convert from UTF-8- TypeError")
```

Before this patch, one bad VT entry panicked the Tokio worker, leaving the scan stuck in `running` from the API consumer’s perspective.

The fix replaces the unwrap with a fallible `match`: unreadable VT entries are logged at `WARN` with their OID and skipped. This matches the existing scheduler behavior for VTs that are not found or handled via Notus.

I reproduced this with a full-and-fast scan against `tleemcjr/metasploitable2` using the GCF Redis cache. Pre-patch, the worker panicked shortly after scan start. Post-patch, the affected OIDs are skipped with WARN logs and the scan continues normally.

This is intentionally a defensive consumer-side fix. The underlying issue appears to be non-UTF-8 data in the NVT cache; a follow-up could consider using `from_utf8_lossy` in `scannerlib::redis` so all consumers tolerate these entries consistently.

### DIFF
--- a/RELICENSE/arch-fan.md
+++ b/RELICENSE/arch-fan.md
@@ -1,0 +1,80 @@
+Thank you for your interest in the project openvas-scanner managed by Greenbone.
+In order for you to make Contributions now or in the future to this
+project, You agree to license your Contributions under the MIT-0 license (see below).
+Please note that You remain the copyright owner, and anyone receives the right to
+use your Contribution in proprietary and open source software without any
+attribution.
+ 
+..................
+ 
+MIT No Attribution (MIT-0)
+ 
+Copyright <YEAR> <COPYRIGHT HOLDER>
+ 
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the "Software"), to deal in
+the Software without restriction, including without limitation the rights to
+use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+the Software, and to permit persons to whom the Software is furnished to do so.
+ 
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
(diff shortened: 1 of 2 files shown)


nichtsfrei:
Thank you very much for your contribution.

As you're a first time contributor may I ask you to follow the instructions in https://github.com/greenbone/openvas-scanner/tree/main/RELICENSE within this PR?

arch-fan:
Ok, sorry I didn't know that 😰 I'll be doing it throughout the day.

Btw, PR was made by Claude, but checking the issue, it seems that's an error with the UTF8 redis decoder. I think before merging, it would be interesting to mention that this would work with from_utf8_lossy. The issue is that you will have replacement characters.

Idk if it's interesting to accept them or just skip them, as they could have unexpected behaviors at parsing.

It seems they aren't a lot of them, but just in case

arch-fan:
> PR was made by Claude

Just in case, I wasn't searching an issue just to PR. I was deploying openvas on a product we're building and found this issue, so told Opus to make a PR fixing it & patch my local build. It seemed to work everything right.

arch-fan:
Done relicense commit. Also, I know here it's not the right place to ask, but would you be interested in implementing nix flakes for development & deployment? I will make a PR if so 😃

If you don't know about nix, https://nixos.org/

nichtsfrei:
I tried to bring up flakes two years ago, but I can try to make a point again.

nichtsfrei (approved):

arch-fan:
> I tried to bring up flakes two years ago, but I can try to make a point again.

Yes, pleaseeeee 🤩🤩🤩 thank you!! 

I’ll be waiting and ready to make a PR if needed.

nichtsfrei:
I asked the team and it is fine by us.

Looking forward to your PR :)

arch-fan:
> I asked the team and it is fine by us.
> 
> 
> 
> Looking forward to your PR :)

YESSS!!!



Thanks! If it’s not urgent, I’ll open the PR in a couple of weeks. I’d like to take the time to do it properly.

PR #2215, opened 2026-05-20
refactoring

nichtsfrei:
https://jira.greenbone.net/browse/SC-1610


rust/src/openvasd/container_image_scanner/endpoints/vts.rs
```diff
@@ -1,46 +1,102 @@
...
+    ) -> AppResult<Response> {
+        let client_id = Self::client_id(&ident);
+
+        if query.detailed() {
+            let vts = self.get_vts(client_id).try_collect::<Vec<_>>().await?;
+            Ok((StatusCode::OK, Json(vts)).into_response())
+        } else {
+            let oids = self.get_oids(client_id).try_collect::<Vec<_>>().await?;
+            Ok((StatusCode::OK, Json(oids)).into_response())
+        }
```
Tehforsch, line 68: Do I see this right that we don't stream anymore here?

Tehforsch (dismissed):
Preliminary approval, I will look in more detail later but we wanted to merge this quickly

rust/src/openvasd/container_image_scanner/endpoints/vts.rs
nichtsfrei: oh shoot, sorry. Initially I had everywhere the content-application header setter manually and then asked an agent to change that to axums Json thingy while I overlooked that it changed the streaming response. My bad. Should be fixed now.

Tehforsch (approved):

PR #2225, opened 2026-06-02
Add: nasl_psrp() builtin function

jjnicola:
**What**:
Add nasl_psrp() builtin function
SC-1396
SC-1397
SC-1398

**Why**:

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/nasl/nasl_smb.c
+++ b/nasl/nasl_smb.c
@@ -17,6 +17,12 @@
 
 #include "../misc/plugutils.h"
 #include "base/hosts.h"
+#include "glib.h"
+#include "glibconfig.h"
+#include "nasl_debug.h"
+#include "nasl_lex_ctxt.h"
+#include "nasl_misc_funcs.h"
+#include "nasl_var.h"
 #include "openvas_smb_interface.h"
 
 #include <arpa/inet.h>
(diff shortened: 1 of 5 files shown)


nasl/nasl_smb.c
```diff
@@ -494,3 +501,382 @@ nasl_win_cmd_exec (lex_ctxt *lexic)
...
+
+  krb5 = !g_strcmp0 (authentication, KERBEROS_AUTH_TYPE);
+  if (host == NULL && host_ip != NULL)
+    {
+      calculate_host = true;
+      host = addr6_as_str (host_ip);
+      if (host != NULL && krb5)
+        {
+          gvm_host = gvm_host_from_str (host);
+          g_free (host);
```
Tehforsch, line 666: Maybe I'm misreading things, but isn't there an execution path where `host` is double-freed between here and line 700?

Tehforsch (approved):

nasl/nasl_smb.c
jjnicola: In the following line 667, `gvm_host_reverse_lookup()` will find the hostname or set host as NULL.
If host is still NULL we leave the function in line 672, otherwise we continue and finally clean the `host` after use in line 700.
Therefore, there is no double free

Tehforsch: Got it, didn't read that right - thanks.

jjnicola: Thanks for the review!

PR #2231, opened 2026-06-04
Add: Restructure feed-filter and add dependency filter functionality

Kraemii:
This restructures the original feed-filter into separate files. Additionally a dependency filter is added. It does not filter based on builtin functions but on dependencies. A single script is given, and all include and script_dependencies are checked. From here out, the collected scripts can be copied into a specified directory.

Example: `target/release/feed-filter dep-filter --feed-out ~/dev/nasl/gather-package-list/ ~/dev/vulnerability-tests/nasl/common/ ~/dev/vulnerability-tests/nasl/common/gather-package-list.nasl`

SC-1583

### DIFF
--- a/rust/src/feed_filter/main.rs
+++ b/rust/src/feed_filter/main.rs
@@ -2,29 +2,49 @@
 //
 // SPDX-License-Identifier: GPL-2.0-or-later WITH x11vnc-openssl-exception
 
+mod builtins;
+mod dep_filter;
 mod error;
+mod output;
+mod script;
+mod stats;
+mod utils;
 
-use clap::Parser;
-use scannerlib::models::{Scan, VT};
-use scannerlib::nasl::syntax::Loader;
-use scannerlib::nasl::syntax::grammar::{Ast, Atom, Expr, FnCall, Statement};
-use scannerlib::nasl::{Code, nasl_std_executor};
-use std::collections::HashSet;
-use std::fmt::Display;
-use std::io::{self};
+use clap::{Parser, Subcommand};
+use std::fs;
 use std::path::PathBuf;
-use std::{collections::HashMap, fs, path::Path};
-use walkdir::WalkDir;
 
+use crate::builtins::BuiltinFunctions;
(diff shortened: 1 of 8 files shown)


Tehforsch:
Before I review, could you tell me what the motivation is behind this change? I understand that now instead of calling feed-filter directly, we call `feed-filter <set of nasl files>` (simplified) and then get those scripts along with their dependencies? Is this expected to be the default use case (as opposed to filtering the entire feed)? If no, we should probably make this an optional argument instead `feed-filter --scripts <set of nasl files>`

Kraemii:
We need this in order to pursuit the next step for the rust only scanner. With this we can extract single scripts, including their dependencies. After that we can check, which builtin functions are missing in order to be able to execute this stack. Also making adjustments, removing not needed dependencies for testing, removing unwanted functionality is much easier within a the extracted smaller feed. In general testing becomes much easier and faster.

rust/src/feed_filter/main.rs
```diff
@@ -2,29 +2,49 @@
...
+    command: Command,
+}
+
+#[derive(Subcommand)]
+enum Command {
+    /// Filter feed based on built-in function coverage.
+    BuiltinFilter(BuiltinFilterArgs),
+    /// Filter feed based on given root scripts and their dependencies.
+    DepFilter(DepFilterArgs),
+}
```
Tehforsch, line 44: Given that we call the binary as `feed-filter` already, I feel like we shouldn't repeat the `filter` word here. 

Maybe the variants could be called 
`ByBuiltin` and `byDependency` or something

Tehforsch (changes requested):

rust/src/feed_filter/main.rs
Kraemii: I just removed the filter part, so the sub command are now `builtin` and `dep`. As their purpose is explained, when calling `feed-filter --help` I guess this is enough

Tehforsch (approved):

PR #2243, opened 2026-06-10
Fix: store the mtime in the redis nvticache for backward compatibility

jjnicola:
Jira: SC-1608

### DIFF
--- a/rust/src/openvasd/scans/scheduling.rs
+++ b/rust/src/openvasd/scans/scheduling.rs
@@ -1,4 +1,6 @@
+use std::path::PathBuf;
 use std::sync::Arc;
+use fslock::LockFile;
 use tokio::sync::RwLock;
 
 use greenbone_scanner_framework::models::{self, Scan};
@@ -9,7 +11,7 @@ use scannerlib::{
     osp,
     scanner::{
         OpenvasdScanner, ScanDeleter, ScanResultFetcher, ScanResultKind, ScanStarter, ScanStopper,
-        preferences,
+        ScannerType, preferences,
     },
 };
 
@@ -28,6 +30,8 @@ use crate::{
     },
     vts::orchestrator::{self, FeedStatusChange},
 };
+
+const LOCK_FILE: &str = "feed-update.lock";
 
 #[derive(Default, Debug)]
 struct IsInProgress {
@@ -111,6 +115,7 @@ struct ScanScheduler<Scanner, Cryptor> {
     // otherwise we would need to store two separate lists.
     feed_sync_in_progress: Arc<RwLock<IsInProgress>>,
     scan_state: ScanStateController,
+    lock_file_dir: String,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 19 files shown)


rust/src/scanner/scanner.rs
```diff
@@ -28,6 +28,13 @@ pub struct ScanResults {
     pub results: Vec<models::Result>,
 }
 
+#[async_trait]
+pub trait ScannerType {
+    fn scanner_type(&self) -> String {
+        String::from("openvasd")
+    }
+}
+
```
Tehforsch, line 39: I'd prefer if we could re-use the `ScannerType` enum from openvasd for this (would have to move it into scannerlib and publically export it from there). We can still have this trait if we rename it (although I'd merge all these Scanner-related types at some point, because there is no point in their existence, but we don't need to do that now), but I'd make it return `ScannerType`

rust/src/openvasd/scans/scheduling.rs
```diff
@@ -175,9 +180,28 @@ impl<T, C> ScanScheduler<T, C> {
     }
 }
 
+fn is_file_locked(path: String) -> bool {
+    let mut file = LockFile::open(&path).expect("Valid path to lock file");
```
Tehforsch: The worst named function in the rust standard library strikes again: 

```suggestion
    let mut file = LockFile::open(&path).expect("Invalid path to lock file");
```

rust/src/openvasd/scans/scheduling.rs
```diff
@@ -358,10 +382,17 @@ where
         let result = match message {
             FeedStatusChange::Need(_) => {
                 let count_running: i64 = self.get_running_count().await;
-                if count_running == 0 {
+                let filelocked = {
```
Tehforsch: Nit: rename to `file_locked` or `is_file_locked`

rust/src/openvasd/vts/mod.rs
```diff
@@ -252,30 +270,138 @@ where
...
+                oid: "fake_oid".to_string(),
+                filename: element
+                    .strip_prefix(&dir_path)
+                    .unwrap()
+                    .to_str()
+                    .unwrap()
+                    .to_string(),
+                ..Default::default()
+            };
+
```
Tehforsch, line 340: Just to help future us: Maybe add a TODO comment here that this is terrible ;)

rust/src/scanner/scanner.rs
```diff
@@ -139,6 +146,12 @@ impl LambdaBuilder {
     }
 }
 
+impl ScannerType for Lambda {
+    fn scanner_type(&self) -> String {
+        String::from("lambda")
+    }
+}
+
```
Tehforsch: Might have to add  `cfg(test)` gated `Lambda` enum variant to `ScannerType` if you do decide to re-use the `ScannerType` enum as above

rust/src/storage/redis/mod.rs
```diff
@@ -375,20 +375,21 @@ mod tests {
                             );
                         }
                         "filename:test.nasl" => {
+                            dbg!(&values);
```
Tehforsch: Remove

Tehforsch (changes requested):

rust/src/storage/redis/mod.rs
jjnicola: ups!

rust/src/scanner/scanner.rs
jjnicola: Move ScannerType from openvasd::config to scannerlib::utils. The trait was renamed and now returns the scanner type instead of the string.

rust/src/openvasd/vts/mod.rs
jjnicola: added a TODO. Will add a jira issue as wel. The whole thing must be done again.

Tehforsch (approved):
thanks!

PR #2249, opened 2026-06-17
Add: Opt-in endpoint hardening (auth, health IP allow-list, declined-response headers)

mde-gb:
## What

Adds three opt-in `[endpoints]` settings to `openvasd`, all defaulting to off so
existing deployments are unaffected:

- `require_authentication` — require the configured auth (mTLS or API key) for
  `/notus`, `/notus/{os}`, and `/vts` (previously always unauthenticated).
- `health_ip_allowlist` — restrict the implemented health probe routes
  (`GET`/`HEAD` on `/health/alive`, `/health/ready`, `/health/started`) to direct
  TCP peer IPs or CIDR networks. `X-Forwarded-For` is intentionally ignored.
- `hide_declined_response_headers` — omit the `authentication`, `api-version`, and
  `feed-version` metadata headers on `401` and `403` responses.

Each is configurable via TOML, CLI flag, and environment variable. Startup fails
fast if `require_authentication` is set without an API key or complete mTLS config.
OpenAPI, README, example configs, the default config snapshot, and tests are updated.

## Why

Today `openvasd` serves the Notus and VT endpoints without authentication and
answers health probes for any caller. That is acceptable behind a trusted gateway,
but risky for any instance that is — intentionally or accidentally — reachable from
an untrusted network. This change lets operators harden such deployments without
breaking the default behavior.

Concrete risks on a publicly reachable instance:

- **Unauthenticated Notus/VT endpoints** disclose which OS advisories and
  vulnerability tests the scanner knows about. An attacker can fingerprint the
  feed and OS coverage to infer what the operator can (and cannot) detect, and use
  that to plan evasion. These endpoints are also comparatively expensive, so open
  access invites resource abuse and denial-of-service amplification.
- **Open health endpoints** (`/health/alive`, `/health/ready`, `/health/started`)
  leak liveness and readiness state. That confirms a live service to scanners and
  helps an attacker time activity (e.g., probe while the service is starting or
  not ready) and map infrastructure during reconnaissance.
- **Metadata headers on declined requests.** Returning `authentication`,
  `api-version`, and `feed-version` on `401`/`403` responses hands version and
  capability details to callers who have *not* authenticated. This is information
  disclosure / version fingerprinting — OWASP A02:2025 Security Misconfiguration —
  and contradicts the least-information and defense-in-depth principles. Precise
  `api-version`/`feed-version` values let an attacker correlate the instance with
  known issues in specific versions (cf. OWASP WSTG information-gathering / version
  disclosure). Suppressing them on declined responses reduces the pre-auth
  attack surface while keeping the headers available to authenticated clients.

Because every option defaults to off, current setups are unchanged; operators opt
in only where the threat model requires it.

## How

- New `EndpointPolicy` in `greenbone-scanner-framework` holds the allow-list and the
  hide-headers flag. `allows_health_peer` matches the peer IP against the CIDR set;
  `is_health_probe_route` gates only the implemented health routes; metadata-header
  suppression is centralized in `should_hide_metadata_headers`.
- The accept loop now forwards the peer `SocketAddr` through `make_service` into the
  entry point so the allow-list can be enforced per connection.
- Request handlers (`GetVTsHandler`, Notus handlers) gain a `require_authentication`
  flag threaded through `RuntimeBuilder` (new `api_key`, `endpoint_policy`, and
  `require_authentication` builder methods).
- Config parsing validates allow-list entries (`IpInet`) and rejects invalid values
  for both TOML and CLI input.

## Contribution checklist

- [x] Conventional commit tag (`Add:` → minor release)
- [x] AI use disclosed (`Co-authored-by: AI (codex/partial)` and
      `Co-authored-by: AI (copilot/partial)`)
- [x] RELICENSE: not required — contributor is a Greenbone member; if a RELICENSE
      file is nonetheless required, it is covered by greenbone/openvas-scanner#2248.

## Misc

Jira: SC-1635

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -2017,6 +2017,7 @@ checksum = "0cc23270f6e1808e30a928bdc84dea0b9b4136a8bc82338574f23baf47bbd280"
 name = "greenbone-scanner-framework"
 version = "0.1.0"
 dependencies = [
+ "cidr",
  "futures",
  "futures-util",
  "http-body",
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -30,6 +30,7 @@ tracing-subscriber = { version = "0.3.17", features = ["env-filter"] }
 tracing-appender = "0.2.4"
 uuid = { version = "1", features = ["v4", "fast-rng", "serde"] }
 once_cell = "1"
+cidr = "0.3.2"
 
 [dependencies]
 aes = "0.8.2"
@@ -137,7 +138,7 @@ tar = "0"
 bzip2 = "0"
 zstd = "0"
 rustls-pki-types = { version = "1.14.0", features = ["std"] }
-cidr = "0.3.2"
+cidr = { workspace = true }
 
 [workspace]
 members = ["crates/nasl-function-proc-macro"]
--- a/rust/api/openapi.yml
+++ b/rust/api/openapi.yml
@@ -71,6 +71,8 @@ paths:
           description: "Ok"
         "403":
           description: "Forbidden by configured health IP allow-list"
+        "503":
+          description: "Service Unavailable"
     get:
       description: "Get application's liveness information. Access can be restricted by the configured health IP allow-list."
       operationId: "get_health_alive"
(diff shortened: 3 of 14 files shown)


ArnoStiefvater:
Hey,

I am closing the PR for now as it follows another strategy of securing the rest of the endpoints than envisioned by us. We might look into it as reference though. Thank you for investing your time and effort!
It is also linked with the created Jira issue.

PR #2251, opened 2026-06-18
Check credentials for duplicates

### LINKED ISSUE #1783: Add sanity check for credentials
As long as the openvas implementation is used it is reusing credentials of the same type rather than iterating over them when they does not work as the credentials are getting exposed via a NASL script that gets the preferences and exposes them as kb item, as an example:
```
krb5_login    = script_get_preference( "KRB5 login:", id:1 );
krb5_password = script_get_preference( "KRB5 password:", id:2 );
krb5_realm    = script_get_preference( "KRB5 realm:", id:3 );
krb5_kdc      = script_get_preference( "KRB5 kdc:", id:4 );

if( krb5_login )    set_kb_item( name:"KRB5/login_filled/0", value:krb5_login );
if( krb5_password ) set_kb_item( name:"KRB5/password_filled/0", value:krb5_password );
if( krb5_realm )    set_kb_item( name:"KRB5/realm_filled/0", value:krb5_realm );
if( krb5_kdc )      set_kb_item( name:"KRB5/kdc_filled/0", value:krb5_kdc );

exit( 0 );
```

This limits the credential per target and credential type to one.

However neither the API documentation nor openvasd prevents adding multiple entries for the same service.

A scan with multiple entries per service
```
    "credentials": [
      {
        "service": "ssh",
        "up": {
          "username": "tesme",
          "password": "testme"
        },
        {
        "service": "ssh",
        "up": {
          "username": "testus",
          "password": "praktikus"
        }
    ],

```

Is accepted although it cannot work as the caller intended (try first, when not working try the second).

The expected behavior would be a 400 with a reasoning that ssh got defined twice.

Tehforsch:
Closes #1783

Jira: SC-1634

### DIFF
--- a/rust/crates/greenbone-scanner-framework/src/models/credential.rs
+++ b/rust/crates/greenbone-scanner-framework/src/models/credential.rs
@@ -58,8 +58,8 @@ pub enum Service {
     Generic,
 }
 
-impl AsRef<str> for Service {
-    fn as_ref(&self) -> &str {
+impl Service {
+    pub fn to_str(&self) -> &str {
         match self {
             Service::SSH => "ssh",
             Service::SMB => "smb",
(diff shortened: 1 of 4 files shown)


jjnicola (approved):
LGTM. Also tested and works like a charm

PR #2253, opened 2026-06-18
Fix: Prefer CVSSv3 over CVSSv2 in Notus advisories

timopollmeier:
When picking the value for the "severity_vector" VT tag of Notus advisories, the CVSSv3 one is now prefrerred over the CVSSv2 one.

This is more consistent with the common behavior of preferring the newest CVSS version where available.

This addresses GEA-1844, GEA-1853


jjnicola (approved):

y0urself:
With a unit test this wouldn't have been happened.

PR #2261, opened 2026-06-24
Fix: get_tcp_option() and get_tcp_element()

jjnicola:
Avoid infinit loop and OOB. 
Adds a buffer size check before memcpy()'ing
Other minor bugs and code improvement (remove dead code)

Jira: SC-1639
Jira: SC-1640


Tehforsch (approved):
Thanks!

PR #2264, opened 2026-06-25
Fix: reported issues in the scanner

jjnicola:
SC-1649

SC-1665
- Fix: useragent in nasl_http2.c. Reported as an use-after-free issue, but the problem was that I was using the wrong variable. Now uses the right one ua
SC-1666
- Fix: avoid stack overflow when password length is greater than 21
SC-1667
- Fix: check buffer lenght before memcpy
SC-1668
- Fix: OOB read in nasl_crypt.c
SC-1669
- Fix: nasl_ntlmv2_hash(). Do length validation.
SC-1670
- Fix: Unbounded VLA / stack overflow / integer overflow
SC-1671
- Fix:nasl_snmp.c: read() return stored in size_t
SC-1672
- Fix: use-after-free in nasl_snmpv1v2c_get()
SC-1673
- Fix: Stack overflow in nasl_win_cmd_exec()
SC-1674
- Fix: OOB read / potential heap overflow nasl_packet_forgery*

**IMPORTANT**

### DIFF
--- a/nasl/nasl_crypto.c
+++ b/nasl/nasl_crypto.c
@@ -254,12 +254,19 @@ nasl_get_sign (lex_ctxt *lexic)
                           "buflen:<bl>, seq_number:<s>)\n");
       return NULL;
     }
-  uint8_t calc_md5_mac[16];
-  simple_packet_signature_ntlmssp ((uint8_t *) mac_key, buf, seq_num,
-                                   calc_md5_mac);
   if (buflen != get_var_size_by_name (lexic, "buf") || buflen < 26)
     {
       nasl_perror (lexic, "OOB read/write\n");
+      return NULL;
+    }
+
+  uint8_t calc_md5_mac[16];
+  if (simple_packet_signature_ntlmssp ((uint8_t *) mac_key, buf,
+                                       (size_t) buflen, seq_num,
+                                       calc_md5_mac)
+      != 0)
+    {
+      nasl_perror (lexic, "get_signature: OOB read\n");
       return NULL;
     }
 
--- a/nasl/nasl_crypto.c
+++ b/nasl/nasl_crypto.c
@@ -29,7 +29,6 @@
 #include "smb_signing.h"
 
 #include <assert.h>
-#include <cstdint>
 #include <ctype.h>
 #include <gcrypt.h>
 #include <glib.h>
@@ -263,8 +262,7 @@ nasl_get_sign (lex_ctxt *lexic)
 
   uint8_t calc_md5_mac[16];
   if (simple_packet_signature_ntlmssp ((uint8_t *) mac_key, buf,
-                                       (size_t) buflen, seq_num,
-                                       calc_md5_mac)
+                                       (size_t) buflen, seq_num, calc_md5_mac)
       != 0)
     {
       nasl_perror (lexic, "get_signature: OOB read\n");
@@ -916,17 +914,30 @@ nasl_ntlmv2_hash (lex_ctxt *lexic)
 
   /* NTLMv2 */
 
-  /* We also get to specify some random data */
+  /* We also get to specify some random data.
+     To avoid DoS we set an uper limit for client chal length.
+     At the moment of fixing this, the max lenght used in a nasl script is 64.
+  */
+  if (client_chal_length > 4096)
+    {
+      nasl_perror (lexic, "Length to big. Max is 4096\n");
+      return NULL;
+    }
+
   ntlmv2_client_data = g_malloc0 (client_chal_length);
   for (i = 0; i < client_chal_length; i++)
     ntlmv2_client_data[i] = rand () % 256;
 
   if (hash_len != 16)
-    nasl_perror (lexic, "owf_in must have a length of 16\n");
+    {
+      nasl_perror (lexic, "owf_in must have a length of 16\n");
+      return NULL;
+    }
 
   /* Given that data, and the challenge from the server, generate a response */
-  SMBOWFencrypt_ntv2_ntlmssp (ntlm_v2_hash, server_chal, 8, ntlmv2_client_data,
-                              client_chal_length, ntlmv2_response);
+  SMBOWFencrypt_ntv2_ntlmssp (
+    ntlm_v2_hash, server_chal, get_var_size_by_name (lexic, "cryptkey"),
+    ntlmv2_client_data, client_chal_length, ntlmv2_response);
 
   /* put it into nt_response, for the code below to put into the packet */
   final_response = g_malloc0 (client_chal_length + sizeof (ntlmv2_response));
--- a/nasl/nasl_crypto.c
+++ b/nasl/nasl_crypto.c
@@ -675,6 +675,8 @@ nasl_ntlmv1_hash (lex_ctxt *lexic)
 
   if (pass_len < 16)
     pass_len = 16;
+  if (pass_len > 21)
+    pass_len = 21;
 
   bzero (p21, sizeof (p21));
   memcpy (p21, password, pass_len);
--- a/nasl/nasl_crypto.c
+++ b/nasl/nasl_crypto.c
@@ -35,6 +35,7 @@
 #include <gvm/base/logging.h>
 #include <stddef.h>
 #include <stdlib.h>
+#include <string.h>
 
 #ifndef uchar
 #define uchar unsigned char
@@ -244,19 +245,32 @@ tree_cell *
 nasl_get_sign (lex_ctxt *lexic)
 {
   char *mac_key = (char *) get_str_var_by_name (lexic, "key");
+  long int keylen = get_var_size_by_name (lexic, "key");
   uint8_t *buf = (uint8_t *) get_str_var_by_name (lexic, "buf");
-  long int buflen = get_int_var_by_name (lexic, "buflen", -1);
+  long int buflen = get_var_size_by_name (lexic, "buf");
   int seq_num = get_int_var_by_name (lexic, "seq_number", -1);
-  if (mac_key == NULL || buf == NULL || buf[0] == '\0' || buflen == -1
-      || seq_num <= -1)
+  if (mac_key == NULL || buf == NULL || buf[0] == '\0' || seq_num <= -1)
     {
+      // deprecate the buf_len argument and use the real buf length
       nasl_perror (lexic, "Syntax : get_signature(key:<k>, buf:<b>, "
-                          "buflen:<bl>, seq_number:<s>)\n");
+                          "seq_number:<s>)\n");
       return NULL;
     }
-  if (buflen != get_var_size_by_name (lexic, "buf") || buflen < 26)
+
+  if (buflen < 26)
     {
-      nasl_perror (lexic, "OOB read/write\n");
+      nasl_perror (lexic,
+                   "OOB read/write. Buffer length must be at least 26. Current "
+                   "length: %ld\n",
+                   buflen);
+      return NULL;
+    }
+
+  if (keylen < 16)
+    {
+      nasl_perror (
+        lexic, "OOB read: key must be at least 16 bytes. Current length: %ld\n",
+        keylen);
       return NULL;
     }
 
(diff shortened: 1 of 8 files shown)


nasl/nasl_crypto.c
```diff
@@ -254,9 +254,21 @@ nasl_get_sign (lex_ctxt *lexic)
                           "buflen:<bl>, seq_number:<s>)\n");
       return NULL;
     }
+  if (buflen != get_var_size_by_name (lexic, "buf") || buflen < 26)
```
Kraemii: Shouldn't we ignore the `buflen` parameter completely and just get the `buflen` from `buf`

nasl/nasl_crypto.c
```diff
@@ -254,9 +254,21 @@ nasl_get_sign (lex_ctxt *lexic)
...
+    {
+      nasl_perror (lexic, "OOB read/write\n");
```
Kraemii: Please put more information into the error handling. We exactly know, what we expect and what was actual given.

nasl/nasl_crypto.c
```diff
@@ -529,6 +541,19 @@ nasl_ntlmv2_response (lex_ctxt *lexic)
                "ntlmv2_hash:<n>, address_list:<a>, address_list_len:<len>)\n");
       return NULL;
     }
+
+  if (address_list_len <= get_var_size_by_name (lexic, "address_list"))
+    {
+      nasl_perror (lexic, "ntlmv2_response: address_list and address_list_len "
+                          "have different sized\n");
```
Kraemii:
```suggestion
                          "have different size\n");
```

nasl/nasl_crypto.c
```diff
@@ -529,6 +541,19 @@ nasl_ntlmv2_response (lex_ctxt *lexic)
...
+      return NULL;
+    }
+  if (address_list_len > UINT16_MAX)
+    {
+      nasl_perror (lexic,
+                   "ntlmv2_response: address_list_len exceed Max lenght\n");
```
Kraemii:
```suggestion
                   "ntlmv2_response: address_list_len exceed Max length\n");
```

Kraemii: Also here you can add the information how long it can be at max and what it actual was.

nasl/nasl_crypto.c
```diff
@@ -529,6 +541,19 @@ nasl_ntlmv2_response (lex_ctxt *lexic)
                "ntlmv2_hash:<n>, address_list:<a>, address_list_len:<len>)\n");
       return NULL;
     }
+
+  if (address_list_len <= get_var_size_by_name (lexic, "address_list"))
```
Kraemii: Is this even correct? We now throw an error, when the `address_list_len` is smaller than the actual `address_list`. If they have to be the same, I would suggest the same, as in the previous comment to just ignore the `address_list_len`. Else the `address_list_len` should not be more than the actual `address_list` to prevent Out of Bounds read/write, correct?

nasl/nasl_crypto.c
```diff
@@ -675,6 +700,8 @@ nasl_ntlmv1_hash (lex_ctxt *lexic)
 
   if (pass_len < 16)
     pass_len = 16;
+  if (pass_len > 21)
+    pass_len = 21;
 
   bzero (p21, sizeof (p21));
   memcpy (p21, password, pass_len);
```
Kraemii: I think this still fails, when the password is less than 16 bytes in length, as random bytes from the memory are then copied, right?

nasl/nasl_crypto.c
```diff
@@ -887,17 +914,30 @@ nasl_ntlmv2_hash (lex_ctxt *lexic)
...
   /* NTLMv2 */
 
-  /* We also get to specify some random data */
+  /* We also get to specify some random data.
+     To avoid DoS we set an uper limit for client chal length.
+     At the moment of fixing this, the max lenght used in a nasl script is 64.
+  */
+  if (client_chal_length > 4096)
+    {
+      nasl_perror (lexic, "Length to big. Max is 4096\n");
```
Kraemii:
```suggestion
      nasl_perror (lexic, "length too big. Max is 4096\n");
```

nasl/nasl_crypto.c
```diff
@@ -887,17 +914,30 @@ nasl_ntlmv2_hash (lex_ctxt *lexic)
...
+    {
+      nasl_perror (lexic, "owf_in must have a length of 16\n");
+      return NULL;
+    }
 
   /* Given that data, and the challenge from the server, generate a response */
-  SMBOWFencrypt_ntv2_ntlmssp (ntlm_v2_hash, server_chal, 8, ntlmv2_client_data,
-                              client_chal_length, ntlmv2_response);
+  SMBOWFencrypt_ntv2_ntlmssp (
+    ntlm_v2_hash, server_chal, get_var_size_by_name (lexic, "cryptkey"),
```
Kraemii: I just looked this one up, and NTLM always expects a 8 byte `cryptkey`/`server_chal`. To have it more resilient to check for the correct size of it before and throw an error, when it differs to that size.

nasl/nasl_packet_forgery.c
```diff
@@ -898,6 +902,13 @@ get_tcp_option (lex_ctxt *lexic)
...
+  if (ipsz < (int) sizeof (struct ip))
+    return NULL;
+
+  // check that ip + tcp is bigger that the original packet
+  if (ip->ip_hl * 4 + 20 > ipsz)
+    return NULL;
+
   // ip header length is given in 32 bits words = 4 bytes.
   if (ip->ip_hl * 4 > ipsz)
     return NULL; /* Invalid packet */
```
Kraemii, line 911: When the second statement if true, the first one is also always true and can be removed

```suggestion
// ip header length is given in 32 bits words = 4 bytes.
if (ip->ip_hl * 4 > ipsz)
return NULL; /* Invalid packet */
```

nasl/nasl_packet_forgery_v6.c
```diff
@@ -407,9 +407,9 @@ insert_ip_v6_options (lex_ctxt *lexic)
   int pad_len;
   char zero = '0';
   int i;
-  int pl;
+  size_t pl;
 
-  if (ip6 == NULL)
+  if (ip6 == NULL && value == NULL)
```
Kraemii:
```suggestion
  if (ip6 == NULL || value == NULL)
```

nasl/nasl_packet_forgery_v6.c
```diff
@@ -863,6 +867,12 @@ get_tcp_v6_option (lex_ctxt *lexic)
 
   /* valid ipv6 header check */
   ipsz = get_var_size_by_name (lexic, "tcp");
+  if (ipsz < (int) sizeof (struct ip6_hdr))
```
Kraemii: `ipsz` is of type `size_t`, which should be the same, as `sizeof`:

```suggestion
  if (ipsz < sizeof (struct ip6_hdr))
```

nasl/nasl_packet_forgery_v6.c
```diff
@@ -863,6 +867,12 @@ get_tcp_v6_option (lex_ctxt *lexic)
...
+    return NULL; // mal formed.
+
+  if (40 + 20 > ipsz)
```
Kraemii: What exactly is this for?

nasl/nasl_smb.c
```diff
@@ -423,9 +423,16 @@ nasl_win_cmd_exec (lex_ctxt *lexic)
   else
     {
       delimiter = strchr (kdc, ',');
+      int kdc_len = delimiter - kdc;
+      if (delimiter && (kdc_len > INET6_ADDRSTRLEN - 1))
+        {
+          g_warning ("kdc hostname value too long (max 45 chars)");
```
Kraemii, line 428: Is there a reason for `g_warning` instead of `nasl_perror`?

nasl/nasl_smb.c
```diff
@@ -423,9 +423,16 @@ nasl_win_cmd_exec (lex_ctxt *lexic)
   else
     {
       delimiter = strchr (kdc, ',');
+      int kdc_len = delimiter - kdc;
```
Kraemii: `delimiter` might be `NULL`

nasl/nasl_snmp.c
```diff
@@ -683,13 +684,19 @@ nasl_snmpv1v2c_get (lex_ctxt *lexic, int version, u_char action)
...
+      g_free (next_oid_str);
+      next_oid_str = g_strdup (result->oid_str);
     }
   else if (result->oid_str != NULL)
-    next_oid_str = result->oid_str;
-
+    {
+      g_free (next_oid_str);
+      next_oid_str = g_strdup (result->oid_str);
+    }
```
Kraemii, line 699: The same issue is still present for the v3 implementation

nasl/smb_signing.c
```diff
@@ -19,21 +19,36 @@
...
+int
 simple_packet_signature_ntlmssp (uint8_t *mac_key, const uchar *buf,
-                                 uint32 seq_number, unsigned char *calc_md5_mac)
+                                 size_t buf_len, uint32 seq_number,
+                                 unsigned char *calc_md5_mac)
 {
   const size_t offset_end_of_sig = (smb_ss_field + 8);
   unsigned char sequence_buf[8];
   struct MD5Context md5_ctx;
+  const uint32 claimed_len = smb_len (buf);
```
Kraemii, line 32: Move `smb_len` behind guards.

nasl/smb_signing.c
Kraemii, line 63: `mac_key` is unchecked. It is only checked for being `!= NULL` by the caller in `nasl_crypto.c`

Kraemii (changes requested):

nasl/nasl_crypto.c
jjnicola: I checked this in the plugins. Always is sent strlen(address_list) as lenght. So, removing.

nasl/nasl_crypto.c
jjnicola: Fixed. We copy the pass up to 16bytes. Later is zero padded.

nasl/nasl_smb.c
jjnicola: this function uses g_warning for everything instead of nasl_perror.

nasl/smb_signing.c
jjnicola: fixed, since I use now the real buffer len, not the passed by the nasl script

jjnicola:
@Kraemii thanks for the review!!!

nasl/smb_signing.c
jjnicola: checked now by the caller

nasl/nasl_crypto.c
```diff
@@ -887,13 +920,30 @@ nasl_ntlmv2_hash (lex_ctxt *lexic)
...
   if (hash_len != 16)
-    nasl_perror (lexic, "owf_in must have a length of 16\n");
+    {
+      nasl_perror (lexic, "owf_in must have a length of 16\n");
+      return NULL;
+    }
+  if (get_var_size_by_name (lexic, "cryptkey") != 8)
+    {
+      nasl_perror (lexic, "cryptkey/server_chal must have a length of 8\n");
+      return NULL;
```
Kraemii: Memory leak here, as we allocated `ntlmv2_client_data`, but in error cases it is not freed. Just move the allocation behind the checks.

nasl/nasl_packet_forgery_v6.c
```diff
@@ -863,11 +867,18 @@ get_tcp_v6_option (lex_ctxt *lexic)
...
   /* valid ipv6 header check */
   ipsz = get_var_size_by_name (lexic, "tcp");
+  if (ipsz < sizeof (struct ip6_hdr))
+    return NULL; // mal formed.
+
+  // ipv6 header has a fixed length of 40 bytes and tcp a minimum header size of
+  // 20 bytes.
+  if (ipsz < 40 + 20)
+    return NULL; /* Invalid packet */
+
```
Kraemii: I guess one check should be sufficient here:
```suggestion
  // tcp minimum header size is 20
  if (ipsz < sizeof (struct ip6_hdr) + 20)
    return NULL;
```

Kraemii (changes requested):
Just two small things here. Thanks for the work!

Kraemii (approved):
Looks good to me now :D

PR #2269, opened 2026-07-06
Fix: OpenAPI spec and scanner framework status codes

Kraemii:
Correct wrong status codes and sync the OpenAPI spec with the scanner framework:

- return 409 Conflict (was 406) for duplicate scan IDs, actions on a scan in the wrong state, and deleting a running scan
- reject malformed `range` query on GET /scans/{id}/results with 400
- notus: return 400 on package parse errors and a JSON body on 500
- document 401 (reusable Unauthorized response) on all authenticated /scans endpoints
- document 503 on /health/alive and /health/started
- drop non-idiomatic responses from the spec

Jira: SC-1592

### DIFF
--- a/rust/api/openapi.yml
+++ b/rust/api/openapi.yml
@@ -1,8 +1,7 @@
 openapi: "3.1.1"
 info:
   title: "Scanner"
-  description: "
-    # Authentication
+  description: " # Authentication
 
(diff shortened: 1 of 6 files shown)


jjnicola (approved):
LGTM

PR #2277, opened 2026-07-10
Fix: reported issues in openvas scanner

jjnicola:
FIX: 
- Fix: nasl_gettimeofday() one-byte heap overflow
- Fix: Heap overflow in insstr. Allocated size doesn't match with the copied bytes when i1>i2
- Fix: prevent buffer underflow and buffer over read in get_udp_element() and get_udp_v6_element()
- Fix: get_icmp_element() and get_icmp_v6_element(). memcpy(..., size +1) reads one byte past the ICMP payload.
- Fix: capture_next_packet(). Add guard before dereferencing/byte-swapping the IP header;  can underflow.
- Fix: forge_icmp_packet(). over-reads the source  buffer (info leak into the forged packet) when the icmp type is 13 or 14. In this cases the payload length is fixed to 12.
- Fix: eregmatch(). loop never advances, infinite loop, unbounded allocation when find_all is set to TRUE

SC-1649
 

**IMPORTANT**

### DIFF
--- a/nasl/nasl_text_utils.c
+++ b/nasl/nasl_text_utils.c
@@ -912,26 +912,29 @@ nasl_eregmatch (lex_ctxt *lexic)
       int index = 0;
       char *current_pos;
       current_pos = string;
-      while (1)
+      while (*current_pos)
         {
+          unsigned int offset = 0, i = 0;
+          size_t cur_len;
+          char *current_pos_cp;
+
           if (regexec (&re, current_pos, (size_t) NS, subs, 0) != 0)
             {
-              regfree (&re);
               break;
             }
 
-          unsigned int offset = 0, i = 0;
+          cur_len = strlen (current_pos);
+          current_pos_cp = g_malloc (cur_len + 1);
+
           for (i = 0; i < NS; i++)
             {
-              char current_pos_cp[strlen (current_pos) + 1];
-
               if (subs[i].rm_so == -1)
                 break;
 
               if (i == 0)
                 offset = subs[i].rm_eo;
 
-              strcpy (current_pos_cp, current_pos);
+              memcpy (current_pos_cp, current_pos, cur_len + 1);
               current_pos_cp[subs[i].rm_eo] = 0;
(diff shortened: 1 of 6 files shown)


Tehforsch (commented):
Could you add some documentation on what this fix is about to the commit message? In the current state I had to go to Jira to even see what this is about and that might be down in the future

nasl/nasl_text_utils.c
```diff
@@ -1045,9 +1053,10 @@ nasl_insstr (lex_ctxt *lexic)
   if (i1 > i2)
     {
       nasl_perror (lexic,
-                   " insstr: warning! 1st index %d greater than 2nd index %d\n",
+                   " insstr: warning! 1st index %ld greater than 2nd index "
+                   "%ld\n",
                    i1, i2);
-      sz3 = sz2;
+      sz3 = i1 + sz2 + (i2 < sz1 - 1 ? sz1 - 1 - i2 : 0);
```
Tehforsch, line 1059: The last example in `doc/manual/nasl/built-in-functions/string-functions/insstr.md` covers the case where `i1 > i2`. Running it with `main` prints "a", but in this new branch it prints the whole string "aabaa"

Tehforsch (changes requested):

nasl/nasl_text_utils.c
jjnicola: When we wrote the manual, we create the examples using the output of the script and with a buggy function, not from the expected result. Altough I don't really know what is the expectation, it makes sense to me to get `aabaa` instead of `a` for `insstr('aaaa', 'b', 2, 1); display(a);` with a fixed function.

jjnicola: Also, checked in the Feed, and there no use of this function with the `idx1 > idx2`.

jjnicola: Update function example in documentation

Tehforsch (approved):

PR #2278, opened 2026-07-13
Integration and api tests

Tehforsch:
TODO: Fix the test which currently asserts on `plugin_feed_info.inc` being returned by GET `/vts`

Rough structure:
- Adds openvasd integration tests add `src/openvasd/api_tests/`.
- They replace the hurl tests, so hurl tests are deleted
- The existing compose setup has been slightly refactored to contain a `base.yaml` which spawns all the necessary processes and a separate `openvasd.yaml` which just spawns `openvasd`. The tests then take place by only spawning `base.yaml` and then inserting `rust-tests.yaml` instead, which runs the unit test in a container similar to `openvasd`.
- The unit tests work by spawning individual openvasd instances which listen at different ports and sending API requests to those instances. 
- Tests that use `openvas` share a single central `openvas` instance. To make sure things work smoothly, they contend for a shared lock so that they ever run in serial. After each test, the redis cache is cleared.

Jira: SC-1655


rust/src/openvasd/api_tests/test_builder.rs
```diff
@@ -0,0 +1,594 @@
...
+
+use crate::{build_runtime, config::Config};
+
+const DEFAULT_TIMEOUT: Duration = Duration::from_secs(5);
+const DEFAULT_SLEEP_INTERVAL: Duration = Duration::from_millis(100);
+const DEFAULT_CLIENT_CERT: &str = "client1.pem";
+const DEFAULT_CLIENT_KEY: &str = "client1.key";
+static OPENVAS_TEST_LOCK: OnceLock<Arc<Mutex<()>>> = OnceLock::new();
+
+pub trait Snapshottable: Serialize {
```
Kraemii, line 45: nit:
```suggestion
pub trait SnapshotTable: Serialize {
```

rust/src/openvasd/api_tests/test_builder.rs
```diff
@@ -0,0 +1,594 @@
...
+    fn redactions() -> Vec<String> {
```
Kraemii, line 48: A small description, what exactly this is for would be helpful. I also read `red actions` first and was really confused 😅

rust/src/openvasd/api_tests/mod.rs
```diff
@@ -0,0 +1,464 @@
...
+            .body::<Status>()
+            .snapshot_if(write_status_snapshot, "status");
+        check_host_info(&status);
+        scan.get_results().await.body::<Vec<models::Result>>()
+    }
+
+    // I am leaving this in here for feature parity with the
+    // hurl smoketests for now but if this ever becomes flaky
+    // or annoying to maintain we should replace it with better
+    // assertions
```
Kraemii: nit:
```suggestion
    // This is here for feature parity with the
    // hurl smoketests for now but if this ever becomes flaky
    // or annoying to maintain we should replace it with better
    // assertions
```

rust/src/openvasd/api_tests/mod.rs
```diff
@@ -0,0 +1,464 @@
...
+        assert_eq!(host_info.dead, 0);
+        assert_eq!(host_info.alive, host_info.all);
+        assert_eq!(host_info.queued, 0);
+        assert_eq!(host_info.finished, host_info.all);
+    }
+
+    // I am leaving this in here for feature parity with the
+    // hurl smoketests for now but if this ever becomes flaky
+    // or annoying to maintain we should replace it with better
+    // assertions
```
Kraemii: nit:
```suggestion
    // This is here for feature parity with the
    // hurl smoketests for now but if this ever becomes flaky
    // or annoying to maintain we should replace it with better
    // assertions
```

rust/src/openvasd/api_tests/test_builder.rs
Tehforsch: It's a thing that's snapshot-able, not a Snapshot-Table. I'll add a description to make it clearer

Kraemii (dismissed):
Nice, thanks for the work here!

Kraemii (approved):

PR #2288, opened 2026-07-18
Fix: keep script stats timestamps ordered

### LINKED ISSUE #2242: `report_scripts` stats JSON: `start` and `stop` values are swapped for per-script entries
> **AI Disclaimer:** This bug was found and this report was written with the help of an AI agent (Claude). The code analysis and the proposed fix have NOT been FULLY reviewed by a human before submission.

## Summary

In the per-host script statistics file written when the `report_scripts` option is set (`<report_scripts>/<scan_id>-stats.json`), the `start` and `stop` values of every per-script entry are swapped: `start` contains the script's *finish* time and `stop` contains its *launch* time. As a result, `start > stop` for every script entry, and computing `stop - start` yields negative durations.

The top-level `scan_time` entry is **not** affected — its `start`/`stop` values are in the correct order.

## Affected versions

The feature was introduced in #1774 (first released in v23.14.1) and the values have been swapped since the initial implementation. The bug is still present in current `main`.

## Steps to reproduce

1. In the scanner preferences set:
   - `log_whole_attack = yes`
   - `report_scripts = /some/writable/path`
2. Run any scan.
3. Inspect `/some/writable/path/<scan_id>-stats.json`.

**Expected:** for each script entry, `start` < `stop` (millisecond timestamps of launch and finish).

**Actual:** for each script entry, `start` > `stop`; `start` is the finish time and `stop` is the launch time.

Example of actual output (note `start` > `stop`):

```json
{"1.3.6.1.4.1.25623.1.0.XXXXXX": {"start": 1767000012345, "stop": 1767000010123}}
```

## Root cause

The producer and the consumer of the `general/script_stats` KB entry disagree on the field order.

When a plugin finishes, the stats string is pushed to the KB as `OID/<finish_ms>/<launch_ms>` — finish time first ([`src/pluginlaunch.c#L198-L204`](https://github.com/greenbone/openvas-scanner/blob/main/src/pluginlaunch.c#L198-L204)):

```c
char msg[2048];
g_snprintf (msg, sizeof (msg), "%s/%ld/%ld", oid,
            (long) ((now.tv_sec * 1000)
                    + (long) (now.tv_usec / 1000)),          /* finish time */
            (long) (processes[i].start.tv_sec * 1000
                    + processes[i].start.tv_usec / 1000));   /* launch time */
kb_item_push_str (kb, "general/script_stats", msg);
```

But when the JSON is generated, the second field is labelled `start` and the third one `stop` ([`src/utils.c#L330-L331`](https://github.com/greenbone/openvas-scanner/blob/main/src/utils.c#L330-L331)):

```c
buf = g_strdup_printf ("{\"%s\": {\"start\": %s, \"stop\": %s}}", spl[0],
                       spl[1], spl[2]);
```

So `spl[1]` (the finish time) is written as `"start"` and `spl[2]` (the launch time) as `"stop"`.

## Proposed fix

Swap the two time arguments in `src/pluginlaunch.c` so that the KB string is `OID/<launch_ms>/<finish_ms>`:

```c
g_snprintf (msg, sizeof (msg), "%s/%ld/%ld", oid,
            (long) (processes[i].start.tv_sec * 1000
                    + processes[i].start.tv_usec / 1000),
            (long) ((now.tv_sec * 1000)
                    + (long) (now.tv_usec / 1000)));
```

(Alternatively, swap `spl[1]`/`spl[2]` in `src/utils.c`.)

### Related minor issue

A few lines above, the duration calculation mutates `processes[i].start` when borrowing microseconds ([`src/pluginlaunch.c#L181-L185`](https://github.com/greenbone/openvas-scanner/blob/main/src/pluginlaunch.c#L181-L185)):

```c
if (now.tv_usec < processes[i].start.tv_usec)
  {
    processes[i].start.tv_sec++;
    now.tv_usec += 1000000;
  }
```

Since this runs before the stats string is built, the recorded launch time can be off by up to one second. Using local copies for the duration calculation would avoid this.

BharatDeva:
**What**:

Fixes the per-script timestamps written to `general/script_stats` so the generated `report_scripts` JSON records `start` before `stop` for each script entry.

The change also keeps the original process start timestamp untouched while calculating the human-readable duration. Previously the microsecond borrow path incremented `processes[i].start.tv_sec`, which could shift the recorded start time before the stats entry was written.

**Why**:

`write_host_stats()` labels the second stats field as `start` and the third as `stop`. `pluginlaunch.c` was pushing those fields in the opposite order, so per-script entries could show `start > stop` and produce negative durations for consumers of the stats JSON.

Fixes #2242

**How**:

Validated locally in WSL Ubuntu 24.04:

- `git diff --check`
- `clang-format --dry-run --Werror src/pluginlaunch.c`
- `cmake -S . -B build-pr-2242 -G Ninja`
  - Configure finds the standard scanner dependencies installed locally (`glib-2.0`, `json-glib-1.0`, `gnutls`, `libcurl`, `mit-krb5`).
  - Configure stops at `libgvm_base>=22.4`, which is provided by Greenbone's CI `gvm-libs` image and is not installed in this local WSL environment.

**Checklist**:

- [x] Tests

### DIFF
--- a/src/pluginlaunch.c
+++ b/src/pluginlaunch.c
@@ -176,12 +176,15 @@ update_running_processes (kb_t main_kb, kb_t kb)
                 }
               else
                 {
-                  struct timeval old_now = now;
+                  long duration_sec =
+                    (long) (now.tv_sec - processes[i].start.tv_sec);
+                  long duration_usec =
+                    (long) (now.tv_usec - processes[i].start.tv_usec);
                   int e;
-                  if (now.tv_usec < processes[i].start.tv_usec)
+                  if (duration_usec < 0)
                     {
-                      processes[i].start.tv_sec++;
-                      now.tv_usec += 1000000;
+                      duration_sec--;
+                      duration_usec += 1000000;
                     }
 
                   if (log_whole)
@@ -189,21 +192,18 @@ update_running_processes (kb_t main_kb, kb_t kb)
                       char *name = nvticache_get_filename (oid);
                       g_message (
                         "%s (%s) [%d] finished its job in %ld.%.3ld seconds",
-                        name, oid, processes[i].pid,
-                        (long) (now.tv_sec - processes[i].start.tv_sec),
-                        (long) ((now.tv_usec - processes[i].start.tv_usec)
-                                / 1000));
+                        name, oid, processes[i].pid, duration_sec,
+                        duration_usec / 1000);
                       g_free (name);
 
                       char msg[2048];
                       g_snprintf (msg, sizeof (msg), "%s/%ld/%ld", oid,
-                                  (long) ((now.tv_sec * 1000)
-                                          + (long) (now.tv_usec / 1000)),
                                   (long) (processes[i].start.tv_sec * 1000
-                                          + processes[i].start.tv_usec / 1000));
+                                          + processes[i].start.tv_usec / 1000),
+                                  (long) ((now.tv_sec * 1000)
+                                          + (long) (now.tv_usec / 1000)));
                       kb_item_push_str (kb, "general/script_stats", msg);
                     }
-                  now = old_now;
                   do
                     {
                       e = waitpid (processes[i].pid, NULL, 0);


Tehforsch:
Closing this due to failure to check the pull request template.

BharatDeva:
Thanks for pointing this out. I opened a replacement PR that follows the repository template, including the RELICENSE file and required disclosure: #2290

PR #2289, opened 2026-07-20
Add: Support Skiron within openvas-scanner

Kraemii:
This will introduce a new preference `notus_route` and deprecate `openvasd_server`. The reason for this is, that the path of the notus endpoint of skiron is different than in openvasd and might change in the future. In this case the path can be adjusted with this config, instead of hardcoding it.

SC-1622


jjnicola (approved):
Tested against openvasd and works as expected ( with both new and deprecated preferences).

Issue #1710, opened 2024-09-03
openvas-scanner errors out at compilation in Arch GNU/Linux


r3k2:
OS: Arch GNU/Linux
Medium: AUR package
Doc: Follow the Arch wiki
I have read and did not find anything related before posting this.
also posted here: https://aur.archlinux.org/packages/openvas-scanner

## openvas-scanner will not compile :(

```
-- Found Doxygen: /usr/bin/doxygen (found version "1.12.0") found components: doxygen missing components: dot
-- Configuring done (0.8s)
-- Generating done (0.0s)
-- Build files have been written to: /home/rek2/.cache/paru/clone/openvas-scanner/src/build
make: Entering directory '/home/rek2/.cache/paru/clone/openvas-scanner/src/build'
[  1%] Building C object misc/CMakeFiles/openvas_misc_shared.dir/bpf_share.c.o
<command-line>: error: "_FORTIFY_SOURCE" redefined [-Werror]
<command-line>: note: this is the location of the previous definition
cc1: all warnings being treated as errors
make[2]: *** [misc/CMakeFiles/openvas_misc_shared.dir/build.make:76: misc/CMakeFiles/openvas_misc_shared.dir/bpf_share.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:229: misc/CMakeFiles/openvas_misc_shared.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
make: Leaving directory '/home/rek2/.cache/paru/clone/openvas-scanner/src/build'
==> ERROR: A failure occurred in build().
    Aborting...
error: failed to build 'openvas-scanner-23.8.5-2': 
error: packages failed to build: openvas-scanner-23.8.5-2
```

I did this manually as well with out using the AUR package and same issue.
make -C build 

Sorry if this is not the right place, looks like a bug on dependencies or versions

jjnicola:
Hi @r3k2 
It seems it is fixed and working now. https://aur.archlinux.org/packages/openvas-scanner#comment-989338
Please, close the issue if it is working for you as well.
Best regards

r3k2:
@jjnicola Hello, thank you. Yes I saw today, I got out of work a bit earlier and have re-tried with no luck.. :( 
same error, I deleted the old cache package to force "paru" to repull the source code. :man_shrugging:  :crying_cat_face:

Kraemii:
[...]

PR #2300, opened 2026-07-24
Fix: respect existing fortify source flags

BharatDeva:
## Summary

OpenVAS currently appends `-D_FORTIFY_SOURCE=2` to the Release hardening flags unconditionally. Some distributions, including Arch, may already provide a different `_FORTIFY_SOURCE` value through their compiler flags. When warnings are treated as errors, that duplicate definition can stop the build.

This keeps the existing default hardening behavior, but only adds `_FORTIFY_SOURCE=2` when neither `CMAKE_C_FLAGS` nor `CMAKE_C_FLAGS_RELEASE` already contains a `_FORTIFY_SOURCE` definition.

## Testing

Configured two WSL Ubuntu CMake builds with `SKIP_SRC=ON` to verify the flag logic without needing the full scanner dependency stack.

Default configuration still adds the project default:

```sh
cmake -S . -B build-fortify-trace-default -DSKIP_SRC=ON -DCMAKE_BUILD_TYPE=Release --trace-expand --trace-source=CMakeLists.txt
```

Relevant trace:

```text
set(HARDENING_FLAGS -Wformat -Wformat-security -fstack-protector )
set(HARDENING_FLAGS -Wformat -Wformat-security -fstack-protector -D_FORTIFY_SOURCE=2 )
```

Configuration with an existing distro/user value does not append a duplicate:

```sh
cmake -S . -B build-fortify-trace-existing -DSKIP_SRC=ON -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_FLAGS='-D_FORTIFY_SOURCE=3' --trace-expand --trace-source=CMakeLists.txt
```

Relevant trace:

```text
set(HARDENING_FLAGS -Wformat -Wformat-security -fstack-protector )
```

Also checked whitespace:

```sh
git diff --check
```

## Generative AI disclosure

I used OpenAI Codex while preparing this patch and reviewed the change and test output before submission.

Refs #1710

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -201,7 +201,11 @@ if (ENABLE_COVERAGE)
   set (COVERAGE_FLAGS "--coverage")
 endif (ENABLE_COVERAGE)
 
-set (HARDENING_FLAGS            "-Wformat -Wformat-security -D_FORTIFY_SOURCE=2 -fstack-protector")
+set (HARDENING_FLAGS            "-Wformat -Wformat-security -fstack-protector")
+if (NOT CMAKE_C_FLAGS MATCHES "(^|[ ;])-D_FORTIFY_SOURCE="
+    AND NOT CMAKE_C_FLAGS_RELEASE MATCHES "(^|[ ;])-D_FORTIFY_SOURCE=")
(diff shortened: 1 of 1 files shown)


Tehforsch:
Closing as we don't see this as an issue on our end. If you run into problems because of this, please consider opening an issue instead.

PR #2306, opened 2026-07-29
Refactor: split greenbone-scanner-framework and move into openvasd

tsurai:
The `greenbone-scanner-framework` crate is being split into two parts. A `models` crate containing shared data used among the workspace and the API related code only used by `openvasd`. The later is moved directly into `openvasd` as a module.

This is an intermediate step in preperation of a major API rework towards axum.


Tehforsch (commented):
I think a cleaner approach is to move `models` into scannerlib entirely and to expose it to openvasd. 

The reason why `greenbone-scanner-framework` was a separate crate to begin with is that once upon a time we thought about using it in entirely different projects. This is not necessary anymore. The remaining advantage of the crate-split is incremental compile times, but we have decided some time ago that we would rather have a monolithic lib crate for the simplicity that it brings and sacrifice a bit of compile time in the process. I'd prefer if we could stick with this approach here too.

(Side note for the openvasd historians among us: This is the eternal cycle of life: `models` used to be a workspace crate, was turned into a scannerlib module, then moved into gsf and has now - although only temporarily - become a workspace crate again before (in my opinion hopefully) reaching its true form as a scannerlib module again)

Tehforsch:
On another note: I saw that `src/openvasd/greenbone_scanner_framework` exists. The naming really doesn't make sense anymore now that this isn't a separate crate usable by other projects. I figure you'll probably address this in a future PR and it might get deleted entirely, but just wanted to make sure you're aware.

tsurai:
I've pulled `models` back into `scannerlib` as you suggested.

> On another note: I saw that `src/openvasd/greenbone_scanner_framework` exists. The naming really doesn't make sense anymore now that this isn't a separate crate usable by other projects. I figure you'll probably address this in a future PR and it might get deleted entirely, but just wanted to make sure you're aware.

Yea, it was my intention to rename it to just `api` but decided to delay it until the `axum` PR since it's responsibilties are still a bit murky in its current state.

rust/src/openvasd/greenbone_scanner_framework/time.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    #[test]
+    fn something_else() {
+        let example = "Fri Feb 10 16:09:30 2023 +0100";
+        assert_eq!(example.as_timestamp(), Some(1676041770));
+        let example = "Fri, 10 Feb 2023 16:09:30 +0100";
+        assert_eq!(example.as_timestamp(), Some(1676041770));
+    }
+}
```
Tehforsch: Delete, unused

rust/src/openvasd/greenbone_scanner_framework/snapshots/openvasd__greenbone_scanner_framework__get_scans__tests__get_scans.snap
```diff
@@ -0,0 +1,5 @@
+---
+source: crates/greenbone-scanner-framework/src/get_scans.rs
+expression: resp
+---
+[""]
```
Tehforsch, line 6: Delete, unused

Tehforsch (changes requested):
Thank you - Looks good for the most part, but please check what happened to the snapshots.

If I run `cargo insta --unreferenced=delete` on this PR, I find three unused snapshots. One of them was created, not moved, but I haven't looked into what happened with the other two. I checked and found some missing tests, on `main` I get 689 tests in total, on this branch I get 686. This fits with the three unused snapshots.

Also linting still fails due to unused stuff.

rust/src/openvasd/greenbone_scanner_framework/snapshots/openvasd__greenbone_scanner_framework__get_scans__tests__get_scans.snap
tsurai: I get a test error when I remove this snap file. Do you mean `greenbone_scanner_framework__get_scans__tests__get_scans.snap` because that doesn't seem to be used?

Tehforsch: Ah yes, my bad. Didn't mentally parse the diff between the two file names :D

tsurai:
> If I run `cargo insta --unreferenced=delete` on this PR, I find three unused snapshots. One of them was created, not moved, but I haven't looked into what happened with the other two. I checked and found some missing tests, on `main` I get 689 tests in total, on this branch I get 686. This fits with the three unused snapshots.

One of those was a copy-paste fail the other two are also unreferenced on `main` and I cannot trace back to when they actually last had running tests. The difference in numbers might be due to the `time` module now being unused and its three tests were skipped?

Tehforsch:
> > If I run `cargo insta --unreferenced=delete` on this PR, I find three unused snapshots. One of them was created, not moved, but I haven't looked into what happened with the other two. I checked and found some missing tests, on `main` I get 689 tests in total, on this branch I get 686. This fits with the three unused snapshots.
> 
> One of those was a copy-paste fail the other two are also unreferenced on `main` and I cannot trace back to when they actually last had running tests.

Haven't looked in detail but I dont get any unused snapshots running `cargo insta test --review --unreferenced=delete` on current main. 

> The difference in numbers might be due to the time module now being unused and its three tests were skipped?

But the time module still exists? It just temporarily existed twice, unless I'm missing something

tsurai:
> Haven't looked in detail but I dont get any unused snapshots running `cargo insta test --review --unreferenced=delete` on current main.

For whatever reason running `cargo insta test --review --unreferenced=delete` doesn't seem to actually pickup the tests in `rust/crates/greenbone-scanner-framework` unless you run it from that directory.

So when run from `/rust` or `/rust/src/openvasd` I get no unused snapshots, but I do get them when run from `/rust/crates/greenbone-scanner-framework` on `main`

Tehforsch:
> > 
> 
> For whatever reason running `cargo insta test --review --unreferenced=delete` doesn't seem to actually pickup the tests in `rust/crates/greenbone-scanner-framework` unless you run it from that directory.
> 
> So when run from `/rust` or `/rust/src/openvasd` I get no unused snapshots, but I do get them when run from `/rust/crates/greenbone-scanner-framework` on `main`

Yeah `--unreferenced` is a little shaky at times, depending on what subset of tests is run - I think it just checks for `.new` files that it didn't encounter during that one test run, so running from a subcrate will probably only run the tests in that crate and then for some reason pick up `.new` files somewhere else (not exactly sure why - maybe it starts from the git root) and then report them as unreferenced. This also frequently happens in crates with feature gates for example. It's not the most reliable command

tsurai:
The remaining two linter errors are false positives since the imports are used in tests and they will become obsolete soon anyway

Tehforsch:
> The remaining two linter errors are false positives since the imports are used in tests

That doesn't make them false positives, you can mark the imports themselves as `#[cfg(test)]` or use qualified paths in the tests themselves. Otherwise, if we merge this, everyone has to deal with the broken CI

Tehforsch (approved):
Thank you!

PR #2308, opened 2026-07-29
Fix: orchestrator ignoring feed update error.

Tehforsch:
Jira: SC-1686

The orchestrator marked the feed as synced first then performed the actual update. It also ignored any errors that occurred during the update.

This led to the following wonderful chain of events:
1. We start a openvasd instance, it quickly does something, finishes and cancels the feed sync before it finishes
2. We start another openvasd instance with the same redis storage, it considers the feed synced and works on an outdated feed.

### DIFF
--- a/rust/src/openvasd/vts/mod.rs
+++ b/rust/src/openvasd/vts/mod.rs
@@ -157,10 +157,17 @@ pub async fn init(
 }
 
 pub trait PluginStorer {
+    fn prepare_feed(&self, hash: &FeedHash) -> Promise<Result<(), WorkerError>>;
     fn store_hash(&self, hash: &FeedHash) -> Promise<Result<(), WorkerError>>;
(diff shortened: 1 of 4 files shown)


rust/src/openvasd/vts/mod.rs
```diff
@@ -211,12 +218,18 @@ where
     match feed_hash.typus {
         FeedType::Products => tracing::debug!(?feed_hash.typus, "Not supported, ignoring."),
         FeedType::Advisories => {
+            let path = feed_hash.path.clone();
+            let hash = feed_hash.hash.clone();
+            ps.prepare_feed(&feed_hash).await?;
+            synchronize_advisories(ps, path, hash, signature_check).await?;
             ps.store_hash(&feed_hash).await?;
-            synchronize_advisories(ps, feed_hash.path, feed_hash.hash, signature_check).await?
```
Tehforsch, line 215: Note how we store the hash first, then start the sync

jjnicola (approved):

PR #2316, opened 2026-07-31
Fix: flatten plugin struct to store right serialized data

jjnicola:
Fix: flatten plugin struct to store right serialized data
Each VT was stored under the object name item which later triggered an error during deserializing

SC-1710

**IMPORTANT**

IMPORTANT: If you have encountered a bug or have a specific feature request, please strongly consider opening an issue for it instead of a PR. It is often easier for maintainers to implement a fix or feature than it is to review a pull request for it. This is especially true for verbose AI generated pull requests. A well-written issue allows maintainers to decide what to do about it instead of presenting us with a finished solution that we can either accept or reject.

Failure to stick to these requirements will get the PR closed IMMEDIATELY.

### DIFF
--- a/rust/src/scanner/running_scan.rs
+++ b/rust/src/scanner/running_scan.rs
@@ -82,11 +82,16 @@ where
     }
 
     async fn run(self) -> Result<(), Error> {
-        let runner = self.make_runner().await?;
+        let runner = match self.make_runner().await {
+            Ok(r) => r,
+            Err(e) => {
+                tracing::warn!("Something went wrong: {}", e.to_string());
+                return Err(e);
+            }
+        };
         self.update_status_at_beginning_of_run(runner.host_info())
             .await;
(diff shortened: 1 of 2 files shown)


rust/src/scanner/running_scan.rs
```diff
@@ -82,11 +82,16 @@ where
     }
 
     async fn run(self) -> Result<(), Error> {
-        let runner = self.make_runner().await?;
+        let runner = match self.make_runner().await {
+            Ok(r) => r,
+            Err(e) => {
+                tracing::warn!("Something went wrong: {}", e.to_string());
```
Tehforsch: I think at this point we might as well just trace the error, i.e. `tracing::warn!("{}", e)`. Also, raise the log level? Warn seems a little low for a real error?

Tehforsch (changes requested):

Tehforsch (approved):

PR #2318, opened 2026-08-03
Remove unused items, mark others as `unused`

Tehforsch:
Part of the monthly clean up sessions, removing some unused items and marking others as `#[allow(unused)]` or `#[cfg(test)]`.

Critical items here: `Product.version` is never used - is this intended?

Jira: SC-1716

### DIFF
--- a/rust/src/feed/mod.rs
+++ b/rust/src/feed/mod.rs
@@ -30,5 +30,4 @@ pub use verify::HashSumFileItem;
 pub use verify::HashSumNameLoader;
 pub use verify::Hasher;
 pub use verify::NoVerifier;
-pub use verify::SignatureChecker;
 pub use verify::check_signature;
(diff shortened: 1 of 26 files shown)


rust/crates/greenbone-scanner-framework/src/models/product.rs
```diff
@@ -10,6 +10,7 @@ use serde::Deserialize;
 #[derive(Deserialize, Debug)]
 pub struct Product {
     /// Version of the file, some version might not be supported by notus
+    #[allow(unused)] // TODO: Figure out when this should be used
```
Kraemii: This would be relevant, if the notus file specification changes and either we are not yet ready for the new specification or our parser is incompatible with older versions of the spec. But as the rust implementation will be replaced by the Skiron implementation, I guess this will never have any use ¯\\\_(ツ)\_/¯

Kraemii (dismissed):

Kraemii (approved):

PR #2320, opened 2026-08-03
Be explicit about which directories need write access in docs

Tehforsch:
Jira: SC-1705


rust/src/openvasd/README.md
```diff
@@ -190,16 +196,64 @@ Options:
...
+          This option will enable table driven local security Checks (LSC). This means gathered packages are sent to a specialized scanner. This is far more efficient than doing checks via NASL. [possible values: true, false]
+      --dry_run <dry_run>
+          A dry run is a simulated scan, with no actual host scanned. This mode is useful for automated testing and also to check up, if the setup is actually working. [possible values: true, false]
+      --results_per_host <results_per_host>
+          Amount of fake results generated per each host in the target list for a dry run scan.
+      --max_mem_kb <max_mem_kb>
+          Maximum amount of memory (in MB) allowed to use for a single script. If this value is set, the amount of memory put into redis is tracked for every Script. If the amount of memory exceeds this limit, the script is not able to set more kb items. The tracked the value written into redis is only estimated, as it does not check, if a value was replaced or appended. The size of the key is also not tracked. If this value is not set or <= 0, the maximum amount is unlimited (Default).
   -h, --help
           Print help
 ```
```
Tehforsch, line 259: I don't massively see the point of having this in the config in the first place (who simultaneously uses the CLI  but does not know to use `--help`? Who is this documentation for?) and would prefer to get rid of this entirely.

However, if we keep it, it should at least be up to date, which is why I updated all of it in this PR, even though I only made minor changes.

Kraemii: I completely agree

Kraemii (approved):

jjnicola (approved):

rust/src/openvasd/README.md
Tehforsch: opened a ticket for this

PR #2323, opened 2026-08-04
Rework ctx

Tehforsch:
Jira: SC-1724

It's a big PR, but a lot of the changes in there are completely mechanical. 
I put some effort into making it reviewable commit by commit, and I recommend you review it this way.
Commit messages also have some explanation of what is going on and why at times.

The main point of this PR is to clearly separate things that belong to the entire scan and things that belong to the current script (VT, NVT, Plugin, who knows!).

The old structure:
Everything except for alive state / multicast groups and snmp_next belongs to the ScriptCtx

The new structure:
Belongs to the ScanCtx:
    ScanId
    The full list of targets
    Storage, Loader, Executor
    Open Sockets (this is something we have to think about in the future, can always be moved)
    scan preferences
    alive test methods
    Notus"Ctx"

Belongs to the ScriptCtx:
    Alive State
    Multicast groups
    snpm_next
    vt_data field (crucial, this is where description functions dump their data)
    target id (this is a single target, uniquely identified by its id, which is an index into the target array in ScanCtx)

This allows two things:
1. It makes the whole logic work in the first place. Previously, we would create a single `ScanCtx` per script. Aside from the fact that this would result in tons of cloning and general ugliness, it also makes no sense. Things that should be shared over the lifetime of a scan would not be shared.
2. It makes feed update a little less awkward (still awkward btw, just not as awkward). Prior to these changes, the only reason feed update would work is because the `Drop` impl of ScanCtx would silently "dispatch" the `vt` field in `ScanCtx` (?? I hope its clear how little sense this makes) in the storage. This was already really finicky and probably buggy in a lot of scenarios because `dispatch` is an async function and `Drop` cannot be async, so we did some awkward workarounds to make this compile (and sort of magically work most of the time). Now this drop impl just doesn't exist anymore, instead it is the job of whoever creates the `ScriptCtx` to decide what to do with the collected `VTData`. This makes it obvious that the only caller that actually cares about this field currently is the feed update. There, we explicitly take the vt out of the ScriptCtx and write it into the storage. 

As a side note, I still don't like a lot of how this works and find it very confusing. The `vt` field currently serves two purposes: 1. Gathering info in feed update/feed transform/description runs and 2. Providing info about the current VT during actual scans. I think a much nicer thing to do would be to just store the VT metadata in the storage directly. No magic initialization with fake data needs to take place (we simply leave unset fields unset). Feed update can read the storage (as it currently does) after the full run. During a normal scan, we pre-populate the storage from the vt-metadata.json once, then all scripts have access to the data. No need to move and clone the VTData field around, initialize it awkwardly in the interpreter  (who really shouldn't care about this kind of thing) just to put it in the storage anyways. 
However, in order not to anger the refactoring gods, I left it as it currently is, but would like to talk about this in the future.


jjnicola (approved):
LGTM and test where successful (same behavior as before the patch).
Also, I really appreciate this kind of PRs split in many commits with explanations.
Thanks!

PR #2328, opened 2026-08-04
Replace scanner tests with API tests

Tehforsch:
This replaces the tests in `src/scanner/tests.rs` with API tests that are not only a lot less verbose but easier to understand and less dependent on scanner internals as well.

The intention of these tests is to check if `required_keys`, `mandatory_keys`, `required_ports` settings are properly respected by the scanner.

The new tests simply create a test feed which contains a bunch of small nasl scripts.

For the kb keys, there is a `key_producer.nasl` script which sets a key in the KB. 
Other kb-related scripts depend on this script and then generate a result. By checking the results we see from the scan, we can see which scripts ran and which ones didn't in a simple way.

Ports work the same way.

Jira: SC-1730


Kraemii (approved):
Nice one,

PR #2330, opened 2026-08-05
Fix: handling of `table_driven_lsc` option

Kraemii:
There were two issues regarding this option:

1. For direct notus calls via the `notus` NASL function, this option was completely ignored
2. For the indirect and automatic notus implementation this option did not default to "yes"

I adjusted the printed messages, as well as the correct handling of the option.

To test this, start a small scan with the following scan config: (remember to change username and password accordingly)

```json
{
  "target": {
    "hosts": [
      "127.0.0.1"
    ],
    "ports": [
      {
        "range": [
          {
            "start": 1,
            "end": 65535
          }
        ]
      }
    ],
    "credentials": [
      {
        "service": "ssh",
        "port": 22,
        "up": {
          "username": "user",
          "password": "pw"
        }
      }
    ]
  },
  "vts": [
    {
      "oid": "1.3.6.1.4.1.25623.1.0.50282"
    }
  ]
}
```

Jira: SC-1731

### DIFF
--- a/src/openvas.c
+++ b/src/openvas.c
@@ -410,7 +410,9 @@ static int
 attack_network_init (struct scan_globals *globals, const gchar *config_file)
 {
   const char *mqtt_server_uri;
+  const char *openvasd_server;
   const char *notus_url;
+  const char *table_driven_lsc;
 
   set_default_openvas_prefs ();
   prefs_config (config_file);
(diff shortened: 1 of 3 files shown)


Tehforsch (dismissed):
Thanks!

Would be nice to have an API test for this (i.e., if i disable table_driven_lsc, can I reach the notus endpoint?), but I think due to the fact that the API tests currently use a central openvas instance, we can't really set the options on it just for a single test. I think it might be worth opening a new ticket for but we can merge this PR as it is.

src/openvas.c
```diff
@@ -431,60 +433,74 @@ attack_network_init (struct scan_globals *globals, const gchar *config_file)
...
+      if (openvasd_server)
         {
-          g_message ("%s: LSC via HTTP(S)", __func__);
+          gchar *full_notus_route;
+          g_warning (
+            "%s: option openvasd_server is deprecated and will be removed "
+            "in the next major release. Please use notus_route instead.",
+            __func__);
+          full_notus_route = g_strconcat (openvasd_server, "/notus/", NULL);
+          prefs_set ("notus_route", full_notus_route);
```
jjnicola: shouldn't be `notus_url` here?
```suggestion
          prefs_set ("notus_url", full_notus_route);
```

jjnicola (changes requested):

src/openvas.c
```diff
@@ -431,60 +433,74 @@ attack_network_init (struct scan_globals *globals, const gchar *config_file)
...
-      notus_url = prefs_get ("notus_url");
-      if (notus_url)
+      openvasd_server = prefs_get ("openvasd_server");
+      if (openvasd_server)
         {
-          g_message ("%s: LSC via HTTP(S)", __func__);
+          gchar *full_notus_route;
+          g_warning (
+            "%s: option openvasd_server is deprecated and will be removed "
+            "in the next major release. Please use notus_route instead.",
```
jjnicola: Also here
```suggestion
            "in the next major release. Please use notus_url instead.",
```

jjnicola (changes requested):

jjnicola (approved):

PR #2339, opened 2026-08-10
Refactor: axum based API migration

tsurai:
Replaces the old HTTP API with one based on axum for better maintainability. While this does require rather major refactoring within `openvasd` the changes are kept minimal to keep the PR scope reviewable. This leads to some awkward code that will be refactored in follow up PRs.

### Behavioral changes
* Routes that accept a `JSON` encoded payload **MUST** set `Content-Type: application/json`.
* `HEAD` routes are now RFC compliant, meaning they do the same as their `GET` counterpart but with the result body removed.
* Routes with a trailing `/` are considered a separate route now, resulting in 404s.
* The `enable-get-scans` option under `[endpoints]` was previously ignored, defaulting to true. It's working properly now and defaults to **false** as documented.

Jira: SC-1610


rust/src/openvasd/api/response/mod.rs
```diff
@@ -1,20 +1,25 @@
...
+use axum::{
+    body::Bytes,
+    http::{StatusCode, header},
+    response::{IntoResponse, Response},
+};
 use futures::{Stream, StreamExt, stream};
 use http_body::{Body, Frame, SizeHint};
-use hyper::{StatusCode, body::Bytes};
 
 pub struct BodyKind {
```
Tehforsch, line 17: I realize this is not a new addition in this migration, just writing it down because it crossed my mind:

I find it very weird that this struct (!) is called `BodyKind` and a cursory glance at its usages throughout the codebase suggests to me it should be renamed to `Response` (admittedly the overlap in naming with the axum type might be annoying, open to also call it `ApiResponse` or whatever), since it clearly contains more than a `Body` and is not a `Kind` in any sense of the word

rust/src/openvasd/api/response/mod.rs
```diff
@@ -1,20 +1,25 @@
...
-    pub status_code: StatusCode,
+    status_code: StatusCode,
     pub content: BodyKindContent,
 }
 
-/// Implements the hyper http body types
+/// Implements the http body types
 pub enum BodyKindContent {
```
Tehforsch, line 23: And, to continue from above - having renamed `BodyKind` to `Response`, this would become `Body` or maybe `BodyKind`, since it actually represents a 
response body.

rust/src/openvasd/api/response/mod.rs
```diff
@@ -178,3 +183,18 @@ impl Body for BodyKindContent {
         }
     }
 }
+
+impl IntoResponse for BodyKind {
+    fn into_response(self) -> Response {
+        let mut resp = Response::new(axum::body::Body::new(self.content));
+        *resp.status_mut() = self.status_code;
+
+        // this struct is only used for streamed results anymore which are all JSON encoded
```
Tehforsch, line 192: So does this mean we can `assert_matches!(self.content, BodyKindContent::BinaryStream)`?

rust/src/openvasd/api/response/mod.rs
```diff
@@ -178,3 +183,18 @@ impl Body for BodyKindContent {
...
+        resp.headers_mut().insert(
+            header::CONTENT_TYPE,
+            "application/json".parse().expect("valid HTTP header value"),
```
Tehforsch, line 195: Nit: As a proud member of the `expect`-dislike club, I would `unwrap()` here. Unless axum is fundamentally broken, this will never fail and - at least to me - the `expect` implies that we are upholding some sort of contract, when in reality we're just using one of the allowed strings

rust/src/openvasd/api/routes/notus.rs
```diff
@@ -0,0 +1,60 @@
...
+use scannerlib::notus::Notus;
+use tokio::sync::RwLock;
+
+use std::sync::Arc;
+
+use crate::api::error::ApiError;
+
+/// Access point for all `/notus` prefixed routes.
+///
+/// Used by [crate::api::router::create_router] to build the HTTP API.
```
Tehforsch, line 17: I'd get rid of the second part of the comment here. Nobody will notice and update this comment when the caller gets renamed and we have LSP references for this purpose

rust/src/openvasd/main.rs
```diff
@@ -63,60 +59,56 @@ async fn setup_sqlite(config: &Config) -> Result<SqlitePool> {
...
-        (None, None) => {
-            // ok no TLS
+    let feed_state = Arc::new(std::sync::RwLock::new(FeedState::Unknown));
+    let (sender, feed) = vts::init(pool.clone(), &config, feed_state.clone()).await;
+    let scanner = scans::init(pool.clone(), &config, sender).await?;
+    let image_scanner =
+        container_image_scanner::init(products.clone(), config.container_image_scanner).await?;
+
+    let auth_method = match (
+        config.tls.client_certs.is_some() && config.tls.certs.is_some() && config.tls.key.is_some(),
```
Tehforsch, line 73: Maybe I'm missing something, but it seems to me that if a user sets one of the three fields, they seem to be trying to set up TLS and we should at least give them a warning or even fail with a misconfiguration if they only supply one or two and miss the required second/third. Maybe we could extract this into a method

```
match (config.tls(), config.endpoints.key.is_some()) {
     ...
}
```

where

```
pub struct TlsConfig {
    pub certs: PathBuf,
    pub key: PathBuf,
    pub client_certs: PathBuf,
}

impl Config { 
    fn tls(&self) -> Result<TlsConfig> { // or Option<TlsConfig>
        // Return Ok(TlsConfig) if all fields on self.tls are set
        // otherwise either return None (and print a warning) or Err(...), depending
        // on how we want to handle this.
    }
}
```

or something of the sort. This seems especially useful, since the invariant that all three have to be set can be carried further downstream into `ApiConfig` by giving it a full `TlsConfig` field instead of the three individual options.

rust/src/openvasd/main.rs
```diff
@@ -63,60 +59,56 @@ async fn setup_sqlite(config: &Config) -> Result<SqlitePool> {
...
-        }
-        (None, None) => {
-            // ok no TLS
+    let feed_state = Arc::new(std::sync::RwLock::new(FeedState::Unknown));
+    let (sender, feed) = vts::init(pool.clone(), &config, feed_state.clone()).await;
+    let scanner = scans::init(pool.clone(), &config, sender).await?;
+    let image_scanner =
+        container_image_scanner::init(products.clone(), config.container_image_scanner).await?;
+
+    let auth_method = match (
```
Tehforsch, line 72: Super minor nitpick but I'd find this easier as

```rust
    if tls {
        Authentication::MTLS
    } else if config.endpoints.key.is_some() {
        Authentication::ApiKey
    } else {
        Authentication::None
    }
```

rust/src/openvasd/vts/redis.rs
```diff
@@ -66,7 +64,7 @@ impl RedisPluginHandler {
 }
 
 fn redis_error_to_worker_error(error: DbError) -> WorkerError {
-    WorkerError::Sync(GetVTsError::External(Box::new(error)))
+    WorkerError::Sync(Box::new(error))
```
Tehforsch, line 67: Writing this down here, doesn't have to be done in this PR:

I think we should switch to a consistent error handling theme at least per-crate. I'd prefer using anyhow consistently within `openvasd` rather than a mix of anyhow and custom error types.

rust/src/openvasd/vts/orchestrator.rs
```diff
@@ -134,7 +133,7 @@ pub enum WorkerError {
     #[error(transparent)]
     Calculation(#[from] feed::VerifyError),
     #[error(transparent)]
-    Sync(#[from] GetVTsError),
+    Sync(#[from] Box<dyn std::error::Error + Send + Sync + 'static>),
```
Tehforsch, line 136: I think replacing this with `Sync(#[from] anyhow::Error)` is simpler

rust/src/openvasd/api/routes/vts.rs
```diff
@@ -0,0 +1,25 @@
+//! `/vts` routes.
+use axum::{
+    Router, extract::State, http::status::StatusCode, response::IntoResponse, routing::get,
+};
+
+use crate::api::{error::ApiError, response::BodyKind, states::Feed};
+
+/// Access point for all `/vts` prefixed routes.
+///
+/// Used by [crate::api::router::create_router] to build the HTTP API.
```
Tehforsch, line 10: Same as with the notus one

rust/src/openvasd/api/routes/container_image_scanner.rs
```diff
@@ -0,0 +1,509 @@
...
+        id: "registry_allow_insecure",
+        name: "Allow HTTP when HTTPS is unavailable",
+        default: PreferenceValue::Bool(false),
+        description: "This allows falling back to unencrypted communication with a registry when HTTPS is unavailable.",
+    },
+];
+
+/// Access point for all `/container-image-scanner` prefixed routes.
+///
+/// Used by [crate::api::router::create_router] to build the HTTP API.
```
Tehforsch, line 42: Same as in notus, remove second part of docstr

rust/src/openvasd/api/routes/scans.rs
```diff
@@ -0,0 +1,537 @@
...
+    auth::{AuthLayer, ClientId},
+    error::ApiError,
+    response::BodyKind,
+    states::ScannerBridge,
+};
+use scannerlib::models;
+
+/// Access point for all `/scans` prefixed routes.
+///
+/// Used by [crate::api::router::create_router] to build the HTTP API.
```
Tehforsch, line 26: Same as in notus, remove this second part of the docstr

rust/src/openvasd/api/auth.rs
```diff
@@ -0,0 +1,246 @@
...
+//!                .route("/foo", get("foo"))
+//!                .route("/bar", get("bar")))
+//!     .nest(
+//!         "/unprotected",
+//!         Router::new()
+//!             .route("/", get("lorem"))
+//!             .route("/", get("ipsum"))
+//!             // enable authentication for all routes in this router: /unprotected/lorem and /unprotected/ipsum
+//!             .layer(crate::auth::AuthLayer::new(Authentication::ApiKey, Arc::new(vec!["api-keys"]))))
+//! ```
```
Tehforsch, line 20: Personal preference: I Would remove the example here - This isn't public API but rather a helper for building up two routers that need it and one can look at the callsites instead.

rust/src/openvasd/api/auth.rs
```diff
@@ -0,0 +1,246 @@
...
+use std::sync::Arc;
+use std::task;
+use tower::{Layer, Service};
+
+/// Type for [`axum`] extension extraction
+pub type ClientId = String;
+
+/// Enabled authentication method.
+#[derive(Copy, Clone, Debug)]
+#[allow(unused)]
```
Tehforsch, line 35: Remove

rust/src/openvasd/api/auth.rs
```diff
@@ -0,0 +1,246 @@
...
+
+    /// Tests that the used api-key is set as the client id.
+    #[tokio::test]
+    async fn auth_api_key_id() -> anyhow::Result<()> {
+        // simulates a route handler
+        let inner = service_fn(|req: Request<Body>| async move {
+            let client_id = req.extensions().get::<ClientId>();
+            assert_eq!(client_id, Some(&"secret2".to_string()));
+            Ok::<Response, Infallible>(Response::new(Body::empty()))
+        });
```
Tehforsch, line 198: Nit: If you write this as something like

```rust
fn check_client_id_route_handler(expected_client_id: &str) -> impl Fn(...) -> ... {
      service_fn(|req: Request<Body>| async move {
          let client_id = req.extensions().get::<ClientId>();
          assert_eq!(client_id, Some(&expected_client_id.to_string()));
          Ok::<Response, Infallible>(Response::new(Body::empty()))
      });
}
```

and use it at both duplication sites the tests will be slightly easier to read imo

rust/src/openvasd/api/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+    let app = router::create_router(cfg);
+
+    match (cfg.server_cert_path.as_ref(), cfg.server_key_file.as_ref()) {
+        // Enable TLS if a certificate and key is present
+        (Some(cert_path), Some(key_path)) => {
+            let server_cert = CertificateDer::from_pem_file(cert_path)?;
+            let server_key = PrivateKeyDer::from_pem_file(key_path)?;
+
+            if CryptoProvider::get_default().is_none() {
+                let _ = rustls::crypto::aws_lc_rs::default_provider().install_default();
```
Tehforsch, line 87: Why is the result ignored here? Haven't looked in detail but I assume that `install_default` returns an error if we have already called it before in this process? Then I think unwrapping is better than silently ignoring any error

rust/src/openvasd/api/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+                    if let Ok(certs) = CertificateDer::pem_file_iter(cert_file.path()) {
+                        roots.add_parsable_certificates(certs.collect::<Result<Vec<_>, _>>()?);
+                    }
+                }
+
+                let verifier = rustls::server::WebPkiClientVerifier::builder(Arc::new(roots))
+                    // Allow connections without client certificates.
+                    // This pushes the authentication to the AuthService.
+                    .allow_unauthenticated()
+                    .build()?;
```
Tehforsch, line 108: Nit: extract this into a separate function? This fn is a little long as it is

rust/src/openvasd/api/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+            axum_server::bind(cfg.address)
+                .acceptor(acceptor)
+                .handle(shutdown_handle)
+                .serve(app.into_make_service())
+                .await?;
+        }
+        (None, None) => {
+            tracing::info!("listening on http://{}", cfg.address);
+
+            // Start without tls and HTTP1.1
```
Tehforsch, line 138: Hyper-Nit: This reads a little weird, maybe
`// Start with HTTP1.1 and without TLS`
or
`// Start without TLS and with HTTP1.1`

rust/src/openvasd/api/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+            axum_server::bind(cfg.address)
+                .handle(shutdown_handle)
+                .serve(app.into_make_service())
+                .await?;
+        }
+        _ => {
+            panic!("both a server cert and key file are required");
```
Tehforsch, line 145: I don't like panicking on errors since error messages are ugly and async threads tend to hang, but the good thing is if we validate the TLS config once on the `Config` struct as suggested above(or below, lost track) then we know that TLS is properly configured here and don't have to do any error handling at all

rust/src/openvasd/api/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+        }
+    };
+
+    println!("waiting for signal");
```
Tehforsch, line 149: Won't this be a little out of place in the usual openvasd output? Maybe I'm missing something but either getting rid of this message or replacing it with a `info!(...)` seems clearer to me

rust/src/openvasd/api/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+                continue
+            }
+        };
+
+        // Send the triggered signal back to the main thread to set the correct exit code
+        tx.send(signal).expect("shutdown signal oneshot channel");
+
+        break;
+    }
+}
```
Tehforsch, line 221: Nitpick

```suggestion rust
async fn shutdown_signals(tx: oneshot::Sender<i32>) {
    let listen_for = |tokio_signal, signal_code| async move {
        signal::unix::signal(tokio_signal)
            .expect("failed to install signal handler")
            .recv()
            .await;
        tracing::info!(signal = signal_code, "Exit based on signal.");
        signal_code
    };
    let signal = loop {
        tokio::select! {
            val = listen_for(SignalKind::interrupt(), SIGINT) => break val,
            val = listen_for(SignalKind::terminate(), SIGTERM) => break val,
            val = listen_for(SignalKind::quit(), SIGQUIT) => break val,
            // Ignore sighup and keep running
            _ = listen_for(SignalKind::hangup(), SIGHUP) => {
                tracing::info!("Ignoring SIGHUP signal.");
            }
        };
    };
    // Send the triggered signal back to the main thread to set the correct exit code
    tx.send(signal).expect("shutdown signal oneshot channel");
}
```

rust/src/openvasd/api/mtls.rs
```diff
@@ -0,0 +1,130 @@
...
+            // call the default RustlsAcceptor which also performs the mTLS handshake
+            let (stream, service) = inner.accept(stream, service).await?;
+
+            // hash the clients peer certificate if one was given
+            let hash = match stream.get_ref().1.peer_certificates() {
+                Some(certs) if !certs.is_empty() => {
+                    MtlsCertHash(Some(Self::hash(certs.first().expect("peer certificate"))))
+                }
+                _ => MtlsCertHash(None),
+            };
```
Tehforsch, line 84: Nit:
```suggestion rust
            let hash = MtlsCertHash(
                stream
                    .get_ref()
                    .1
                    .peer_certificates()
                    .and_then(|certs| certs.first().map(Self::hash)),
            );
```

rust/src/openvasd/api/states.rs
```diff
@@ -0,0 +1,255 @@
...
+use super::StreamResult;
+use futures::{StreamExt, TryStreamExt};
+use scannerlib::models::{self, FeedState, Phase, VTData};
+use tokio::sync::mpsc::Sender;
+
+/// Feed access point.
+#[derive(Clone)]
+pub struct Feed {
+    /// Storage independent plugin fetcher.
+    fetcher: Arc<Box<dyn PluginFetcher + Send + Sync + 'static>>,
```
Tehforsch, line 26: Nit: How is this storage independent? Independent as in "we dont use a storage generic here"? Would prefer getting rid of the docstr

rust/src/openvasd/api/states.rs
```diff
@@ -0,0 +1,255 @@
...
+                    .map(|x| x.map_err(ApiError::VtsError)),
+            ))
+        } else {
+            Err(ApiError::FeedNotSynced)
+        }
+    }
+
+    /// Reads the current feed state.
+    pub fn state(&self) -> RwLockReadGuard<'_, FeedState> {
+        self.feed_state.read().expect("failed to read feed state")
```
Tehforsch, line 70: I think "failed to read feed state" will potentially be a very confusing error message, since it suggests that reading the feed state is where the application failed, when really what happened is that the lock is poisoned due to another task panicking. I prefer `unwrap()` here

rust/src/openvasd/api/states.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    /// Posts a given scan without starting it.
+    pub async fn post_scan(&self, client_id: &str, scan: &models::Scan) -> Result<(), DAOError> {
+        if let Some(ref crypter) = self.crypter {
+            ScanDB::new(&self.pool, (crypter.as_ref(), client_id, scan))
+                .exec()
+                .await
+                .map(|_| ())
+        } else {
+            DBScan::new(&self.pool, (client_id, scan)).exec().await
```
Tehforsch, line 106: Confusing naming `ScanDB` and `DBScan`? If this is part of what will be cleaned up in a future PR we can leave it in, but otherwise I'd prefer renaming

rust/src/openvasd/api/states.rs
```diff
@@ -0,0 +1,255 @@
...
+
+        match msg {
+            Ok(msg) => {
+                if let Some(ref scheduler) = self.scheduler {
+                    scheduler.send(msg).await.map_err(|e| e.into())
+                } else {
+                    DBScan::new(&self.pool, (id, action))
+                        .retry_exec()
+                        .await
+                        .map_err(|e| e.into())
```
Tehforsch, line 203: I'm probably just confused we differentiate between DBScan and ScanDB in most places except here (and one time below). Why?

rust/src/openvasd/api/routes/scans.rs
```diff
@@ -0,0 +1,537 @@
...
+///
+/// Authenticated: yes
+///
+/// Returns `JSON` encoded list of all available scan preferences, which can be set within the scan
+/// configuration via the POST /scans endpoint.
+///
+/// ## Errors
+/// * 401: unauthorized access
+async fn get_scans_preferences() -> impl IntoResponse {
+    axum::Json::from(scannerlib::scanner::preferences::preference::PREFERENCES)
```
Tehforsch, line 68: nit: replace qualified path with use

rust/src/openvasd/api/routes/scans.rs
```diff
@@ -0,0 +1,537 @@
...
+    client_id: Extension<ClientId>,
+    Path(scan_id): Path<String>,
+    State(scanner): State<ScannerBridge>,
+    action: Result<Json<models::ScanAction>, JsonRejection>,
+) -> Result<impl IntoResponse, ApiError> {
+    scanner
+        .schedule_scan(&client_id, &scan_id, action?.action)
+        .await?;
+
+    // maybe change to 202 Accepted instead
```
Tehforsch, line 181: Add a `TODO` to the comment? Or remove

rust/src/openvasd/api_tests/mod.rs
```diff
@@ -75,6 +75,141 @@ async fn get_scans_preferences() {
...
+        .request(GET, "/container-image-scanner/scans/preferences")
+        .await
+        .assert_status(StatusCode::OK)
+        .body::<Vec<BTreeMap<String, Value>>>();
+
+    // The full response body looks ugly, so we extract
+    // it as a map to make the snapshot more readable
+    // Then we sort by id to have some sort of order
+    body.sort_by_key(|entry| entry["id"].to_string());
+    body.snapshot("body");
```
Tehforsch, line 210: Nit: Maybe introduce a `ScanPreferences` that impls `Snapshottable` so we can reduce the duplication.

rust/src/openvasd/api_tests/mod.rs
```diff
@@ -291,8 +426,10 @@ async fn container_image_scanner_deadlock() {
     };
 
     let scan = t.create_container_image_scan(scan).await;
+
     scan.start().await;
-    scan.wait_for(Phase::Failed.with_timeout(Duration::from_secs(5)))
+
+    scan.wait_for(Phase::Failed.with_timeout(Duration::from_secs(10)))
```
Tehforsch, line 432: nothing to address now (not at all) but the fact we have to manually tune the timeout here annoys me. it feels too flaky. not this PRs fault at all though, we'll have to try to come up with a better solution to wait for scan statuses

rust/src/openvasd/config/mod.rs
```diff
@@ -497,16 +496,7 @@ impl Config {
...
-            )
-            .arg(
-                clap::Arg::new("enable-get-performance")
-                    .env("ENABLE_GET_PERFORMANCE")
-                    .long("enable-get-performance")
-                    .num_args(0..=1)
-                    .value_parser(clap::builder::BoolValueParser::new())
-                    .default_missing_value("false")
-                    .help("enable get performance endpoint. Default 'false'."),
+                    .help("enable get scans endpoint. Default 'false'."),
```
Tehforsch, line 499: I probably missed the discussion on this - what happened to the old name, is it deprecated?

rust/src/openvasd/container_image_scanner/mod.rs
```diff
@@ -42,11 +34,674 @@ pub async fn init(
...
-        SqlPluginStorage::from(vt_pool),
-        feed_state,
-        Some(scan.prefix()),
-    );
-    Ok((scan, vts))
+    Ok(ScannerBridge::new(pool, None, None))
+}
+
+#[cfg(test)]
+pub mod scans_utils {
```
Tehforsch, line 41: Nothing for this PR, but we should think about merging these tests with the API tests at some point

rust/src/openvasd/scans/mod.rs
```diff
@@ -1,240 +1,10 @@
...
-    E: Send + Sync,
-{
-    fn get_scans(&self, client_id: String) -> StreamResult<String, GetScansError> {
-        Box::pin(
-            ScanDB::new(&self.pool, client_id)
-                .stream_fetch()
-                .map_err(into_external_error),
-        )
-    }
-}
```
Tehforsch, line 97: :D

rust/src/openvasd/scans/mod.rs
```diff
@@ -1,240 +1,10 @@
...
-
-            // everything else should have ON DELETE CASCADE
-            ScanDB::new(&self.pool, id)
-                .exec()
-                .await
-                .map_err(|e| DeleteScansIDError::External(Box::new(e)))
-                .map(|_| ())
-        })
-    }
-}
```
Tehforsch, line 237: :D number 2

Tehforsch (changes requested):
Looks great! Thank you!

I have some initial comments just from looking at the code. I'll also look at the larger structure a little bit now, but if anything comes up I'll do it in a second review

rust/src/openvasd/api/mod.rs
```diff
@@ -0,0 +1,257 @@
...
+        shutdown_signals(signal_tx).await;
+        handle.graceful_shutdown(None);
+    });
+
+    let app = router::create_router(cfg);
+
+    match (cfg.server_cert_path.as_ref(), cfg.server_key_file.as_ref()) {
+        // Enable TLS if a certificate and key is present
+        (Some(cert_path), Some(key_path)) => {
+            let server_cert = CertificateDer::from_pem_file(cert_path)?;
```
Tehforsch, line 83: This only gets the first cert from the pem file, the previous impl used `pem_file_iter` to iterate over all the certificates and passed them all to the server config. Is there any reason to change the behavior or is this a regression?

rust/src/openvasd/api/routes/vts.rs
```diff
@@ -0,0 +1,25 @@
...
+///
+/// Authenticated: no
+///
+/// Returns a streamed response containing a `JSON` encoded array of `OIDs`.
+///
+/// ## Errors
+/// * 503: feed state is unknown or unsynced
+async fn get_vts(State(feed): State<Feed>) -> Result<impl IntoResponse, ApiError> {
+    Ok(BodyKind::from_result_stream(StatusCode::OK, feed.get_oids()?).await)
+}
```
Tehforsch, line 25: This endpoint does not take the `information` parameter into account.

API test to reproduce:
```rust
impl Snapshottable for Vec<models::VTData> {
    fn prepare(&mut self) {
        self.sort_by_key(|vt| vt.oid.clone());
    }

    fn redactions() -> Vec<String> {
        vec![]
    }
}

#[tokio::test]
async fn get_vts_with_information() {
    let t = Test::new("get_vts_with_information")
        .config("basic_feed")
        .await;

    t.request(GET, &format!("/vts?information=1"))
        .wait_for(
            StatusCode::OK
                .with_timeout(Duration::from_millis(100))
                .with_intermediate_status(StatusCode::SERVICE_UNAVAILABLE),
        )
        .await
        .assert_status(StatusCode::OK)
        .body::<Vec<models::VTData>>()
        .snapshot("vts");
}
```

On `main` this yields a detailed list of the VTs. On this branch, I still get only the OIDs

rust/src/openvasd/api/routes/container_image_scanner.rs
```diff
@@ -0,0 +1,509 @@
...
+            },
+        )
+        .route("/", post(post_scans))
+        .route("/preferences", get(get_scans_preferences))
+        .route("/{id}", get(get_scans_id))
+        .route("/{id}", post(post_scans_id))
+        .route("/{id}", delete(delete_scans_id))
+        .route("/{id}/results", get(get_scans_id_results))
+        .route("/{id}/results/{rid}", get(get_scans_id_results_rid))
+        .route("/{id}/status", get(get_scans_id_status))
```
Tehforsch, line 67: The old CIS route had `/vts` as a separate endpoint as well. I am not sure if this is supposed to be kept. If we decide so, here is an API test to ensure it exists/works
(requires `Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord` derives for `Oid`)
```rust
impl Snapshottable for Vec<Oid> { 
    fn prepare(&mut self) {
        self.sort();
    }

    fn redactions() -> Vec<String> {
        vec![]
    }
}

#[tokio::test]
async fn get_vts_container_image_scanner() {
    let t = Test::new("get_container_image_scanner_vts")
        .config("basic_feed")
        .await;
    t.request(GET, "/container-image-scanner/vts")
        .await
        .assert_status(StatusCode::OK)
        .body::<Vec<Oid>>()
        .snapshot("vts");
}
```

rust/src/openvasd/api/routes/scans.rs
```diff
@@ -0,0 +1,537 @@
...
+            && let Ok(end) = end.parse::<usize>()
+        {
+            Ok((start, end))
+        } else {
+            Err(ApiError::InvalidInput(
+                "\"{range}\" is not a valid range".to_string(),
+            ))
+        }
+    } else if let Ok(idx) = range.parse::<usize>() {
+        Ok((idx, idx))
```
Tehforsch, line 341: In the old implementation, an omitted end of the range used to mean the open range `start..`, i.e. every result starting from `start`. Here it means `= start`. Fix and add a test?

rust/src/openvasd/api/error.rs
```diff
@@ -0,0 +1,156 @@
...
+            ApiError::Database(e) => match e {
+                DAOError::DBViolation(_) => StatusCode::CONFLICT,
+                DAOError::NotFound => StatusCode::NOT_FOUND,
+                DAOError::Corrupt | DAOError::Infrastructure(_) => {
+                    StatusCode::INTERNAL_SERVER_ERROR
+                }
+            },
+            ApiError::Notus(e) => match e {
+                NotusError::UnknownProduct(_) => StatusCode::NOT_FOUND,
+                _ => StatusCode::INTERNAL_SERVER_ERROR,
```
Tehforsch, line 58: A malformed notus package is turned into an internal server error. This used to be a `BAD_REQUEST`.

```rust
#[tokio::test]
async fn notus_malformed_package() {
    let t = Test::new("notus_malformed_package").config("notus").await;
    // Very professional creation of a malformed package
    let package = "malformed_input~";
    t.request(POST, "/notus/test")
        .json(vec![package])
        .await
        .assert_status(StatusCode::BAD_REQUEST);
}
```

rust/src/openvasd/api/routes/notus.rs
```diff
@@ -0,0 +1,60 @@
...
+/// * 500: internal error
+async fn post_notus(
+    Path(os): Path<String>,
+    State(notus): State<Arc<RwLock<scannerlib::notus::Notus>>>,
+    packages: Result<Json<Vec<String>>, JsonRejection>,
+) -> Result<impl IntoResponse, ApiError> {
+    let mut notus = notus.write_owned().await;
+
+    let result = notus.scan(&os, &packages?)?;
+    Ok(Json(result))
```
Tehforsch, line 59: The previous impl used `tokio::spawn_blocking` here to avoid blocking the runtime since `Notus::scan` is very much not async (it should be, but for now it isnt). I suggest being defensive and still using `spawn_blocking` here and then to eventually make the method properly async. 

Feel free to add a comment that this can be removed when it becomes async

rust/src/openvasd/api/auth.rs
```diff
@@ -0,0 +1,246 @@
...
+    Mtls,
+    ApiKey,
+    Disabled,
+}
+
+impl Display for Authentication {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::Mtls => write!(f, "mTLS"),
+            Self::ApiKey => write!(f, "apikey"),
```
Tehforsch, line 46: This used to be `api-key`?

Can be tested by
```rust
#[tokio::test]
async fn api_key_authentication_header() {
    let t = Test::new("api_key_authentication_header")
        .config("api_key")
        .await;

    t.request(HEAD, "/health/alive")
        .await
        .assert_status(StatusCode::OK)
        .assert_header("authentication", "api-key");
}
```

with `api_key.toml`:

```toml
[scanner]
type = "openvasd"

[endpoints]
enable_get_scans = false # on main
key = "test-api-key"
```

Tehforsch (changes requested):

rust/src/openvasd/api/states.rs
tsurai: Independent as in it doesn't matter whether the data is coming from redis or sqlite

rust/src/openvasd/api/states.rs
tsurai: The scanner and container image scanner currently both have their own database access which are `ScanDB` and `DBScan`. Will be unified with the database refactor.

rust/src/openvasd/api/states.rs
tsurai: The scanner and cis use different mechanisms to schedule a scan.

The scanner uses a scheduling channel while the container image scanner polls the database for new scan jobs in a loop.

rust/src/openvasd/config/mod.rs
tsurai: `enable_get_scans` was suppose to be `false` by default but `true` which wasn't consistantly documented in all places either.

As for `enable_get_performance` the option does nothing since the actual logic had been removed long ago. I asked in a daily and was told to just remove it.

rust/src/openvasd/api/mod.rs
tsurai: `install_default` can only succeed once per thread and "fails" safely otherwise. The actual return value is even Result<(), Arc<Self>>. Even if it is `Err` it is still successful.

Doing it this way is the recommended way of handling it by the authors.

rust/src/openvasd/api/mod.rs
tsurai: The leaf certificate is suppose to be the first section of a `.pem` file which is what `from_pem_file` reads.

There is no need to iterate over it unless you either create a non-standard `.pem` file or bundle both private key and certificate in a single file. To me both sound like a bad configuration but I don't know if this is something we explicitly support.

rust/src/openvasd/main.rs
Kraemii: I thing the current implementation looks better, and also does a nice job logging what's going on

rust/src/openvasd/vts/orchestrator.rs
tsurai: I actually tried this and it lead to an internal compiler error of the rust compiler soooo I will keep it like this for now :D

rust/src/openvasd/api/auth.rs
tsurai: This would require the expected_client_id to be `Arc` and then cloned multiple times and so on. I rather have a little bit of duplication instead

rust/src/openvasd/api/mod.rs
Tehforsch: Ok nvm then, let's leave it as it is

rust/src/openvasd/api/routes/container_image_scanner.rs
tsurai: It seems rather illogical since the same route doesn't exist under `/scan/vts` either, and instead sits independent at the root `/vts`.

But if this was specified by the customer and can't be changed I will add it again.

rust/src/openvasd/api/states.rs
```diff
@@ -0,0 +1,256 @@
...
+        } else {
+            Err(ApiError::FeedNotSynced)
+        }
+    }
+
+    /// Reads the current feed state.
+    pub fn state(&self) -> RwLockReadGuard<'_, FeedState> {
+        self.feed_state
+            .read()
+            .expect("failed acquire feed state read lock")
```
Tehforsch, line 71:
```suggestion rust
            .expect("failed to acquire feed state read lock")
```

Tehforsch (approved):

rust/src/openvasd/api/states.rs
Tehforsch: I was going to not block this huge PR on this pathetically small suggestion, but since it needs rebasing anyways, I might as well :D

Tehforsch:
Thank you, looks great!

ArnoStiefvater (changes requested):
Just a blocker comment to circumvent accidental merge: Will be merged after next GOS release.

PR #2346, opened 2026-08-12
Fix deadlock on slow db connection.

Tehforsch:
Fix deadlock on slow db connection.

This fixes a specific bug in which a streamed response would own
its own database connection, which led to problems if only a single db
connection was allowed at a time, since the stream might be read slowly,
blocking all other async tasks from making progress.

Jira: SC-1714

### DIFF
--- a/rust/src/openvasd/database/sqlite/scans.rs
+++ b/rust/src/openvasd/database/sqlite/scans.rs
@@ -1,4 +1,4 @@
-use futures::StreamExt;
+use futures::{StreamExt, stream};
 use scannerlib::models::{self, AliveTestMethods};
 use sqlx::{Connection, Row, Sqlite, SqlitePool, query, query_scalar, sqlite::SqliteRow};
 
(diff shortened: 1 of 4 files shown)


rust/src/openvasd/database/sqlite/scans.rs
```diff
@@ -503,17 +503,29 @@ where
 {
     fn stream_fetch(self) -> DAOStreamer<String> {
         let (db, client_id) = self.inner();
-        let result = query(
-            r#"
+        let pool = db.clone();
+        // The code below is a temporary fix for a problem we encountered where a response being
+        // read slowly if sqlite was configured to only a single connection at a time would lead to
+        // a deadlock.
+        // The implementation is not great, since it fetches all entries from the DB and
```
jjnicola, line 510: I don't think it is serious, since the amount of entries that are kept in memory for the stream are just a few ones (IIUC it is just the amount of stored scans, running, stopped, queued). So keeping this in memory while streaming should not be a problem. 
It would be different if we would have to keep the whole feed in memory for streaming it.

jjnicola (approved):

PR #2348, opened 2026-08-13
Add: Integrate AliveTest (Boreas) to openvasd scanner

jjnicola:
Perform AliveTest (Boreas) before starting a scan, so only the a live host are attacked.
SC-1527

### DIFF
--- a/rust/src/alive_test/alive_test.rs
+++ b/rust/src/alive_test/alive_test.rs
@@ -392,7 +392,9 @@ impl Scanner {
                 println!("{t} via {}", AliveTestMethods::ConsiderAlive);
                 if let Some(tx) = &notify {
                     // if there is no receiver, we just stop here.
-                    tx.send(t.clone()).await.unwrap();
+                    if tx.send(t.clone()).await.is_err() {
+                        break;
+                    }
                 }
             }
             return Ok(alive);
@@ -413,29 +415,28 @@ impl Scanner {
         let send_handle = tokio::spawn(send_task(methods_c, trgt, timeout, tx_ctl));
 
         while let Some(alivehost) = rx_msg.recv().await {
-            let (newly_found, newly_found_method) = if self.target.contains(&alivehost.ip)
+            let newly_found = if self.target.contains(&alivehost.ip)
                 && !alive.contains(&alivehost.ip)
             {
-                (
-                    Some(alivehost.ip.clone()),
-                    Some(alivehost.detection_method.clone()),
-                )
+                Some((alivehost.ip.clone(), alivehost.detection_method.clone()))
             } else if let Some(Ok(dst)) = get_host_discovery_ipv6_net(&self.methods, &self.target)
-                && dst.contains(&alivehost.ip.parse::<std::net::Ipv6Addr>().unwrap())
-            {
-                (
-                    Some(alivehost.ip.clone()),
-                    Some(alivehost.detection_method.clone()),
+                && dst.contains(
+                    &alivehost
+                        .ip
+                        .parse::<std::net::Ipv6Addr>()
+                        .expect("IPv6 address"),
                 )
+            {
+                Some((alivehost.ip.clone(), alivehost.detection_method.clone()))
             } else {
-                (None, None)
+                None
             };
 
-            if let Some(host) = newly_found {
+            if let Some((host, method)) = newly_found {
                 if let Some(tx) = &notify {
                     tx.send(host).await.unwrap();
                 } else {
-                    println!("{} via {:?}", &host, &newly_found_method.unwrap());
+                    println!("{} via {:?}", &host, &method);
                 }
                 alive.insert(alivehost.ip.clone());
             }
--- a/rust/src/alive_test/alive_test.rs
+++ b/rust/src/alive_test/alive_test.rs
@@ -52,7 +52,7 @@ struct AliveTestCtlStop;
 #[derive(Debug, Clone)]
 struct AliveHostInfo {
     ip: String,
-    detectihttp_method: AliveTestMethods,
+    detection_method: AliveTestMethods,
 }
 
 struct PktCodec;
@@ -117,7 +117,7 @@ fn process_ipv4_packet(packet: &[u8]) -> Result<Option<AliveHostInfo>, AliveTest
         {
             return Ok(Some(AliveHostInfo {
                 ip: pkt.get_source().to_string(),
-                detectihttp_method: AliveTestMethods::Icmp,
+                detection_method: AliveTestMethods::Icmp,
             }));
         }
     }
@@ -130,7 +130,7 @@ fn process_ipv4_packet(packet: &[u8]) -> Result<Option<AliveHostInfo>, AliveTest
         if tcp_packet.get_destination() == FILTER_PORT {
             return Ok(Some(AliveHostInfo {
                 ip: pkt.get_source().to_string(),
-                detectihttp_method: AliveTestMethods::TcpSyn,
+                detection_method: AliveTestMethods::TcpSyn,
             }));
         }
     }
@@ -149,7 +149,7 @@ fn process_ipv6_packet(packet: &[u8]) -> Result<Option<AliveHostInfo>, AliveTest
         let make_alive_host_ctl = |pkt: Ipv6Packet<'_>, method| {
             Ok(Some(AliveHostInfo {
                 ip: pkt.get_source().to_string(),
-                detectihttp_method: method,
+                detection_method: method,
             }))
         };
 
@@ -169,7 +169,7 @@ fn process_ipv6_packet(packet: &[u8]) -> Result<Option<AliveHostInfo>, AliveTest
         if tcp_packet.get_destination() == FILTER_PORT {
             return Ok(Some(AliveHostInfo {
                 ip: pkt.get_source().to_string(),
-                detectihttp_method: AliveTestMethods::TcpSyn,
+                detection_method: AliveTestMethods::TcpSyn,
             }));
         }
     }
@@ -183,7 +183,7 @@ fn process_arp_frame(frame: &[u8]) -> Result<Option<AliveHostInfo>, AliveTestErr
     if arp.get_operation() == ArpOperations::Reply {
         return Ok(Some(AliveHostInfo {
             ip: arp.get_sender_proto_addr().to_string(),
-            detectihttp_method: AliveTestMethods::Arp,
+            detection_method: AliveTestMethods::Arp,
         }));
     }
     Ok(None)
@@ -406,12 +406,12 @@ impl Scanner {
         while let Some(alivehost) = rx_msg.recv().await {
             if self.target.contains(&alivehost.ip) && !alive.contains(&alivehost.ip) {
                 alive.insert(alivehost.ip.clone());
-                println!("{} via {:?}", &alivehost.ip, &alivehost.detectihttp_method);
+                println!("{} via {:?}", &alivehost.ip, &alivehost.detection_method);
             } else if let Some(Ok(dst)) = get_host_discovery_ipv6_net(&self.methods, &self.target)
                 && dst.contains(&alivehost.ip.parse::<std::net::Ipv6Addr>().unwrap())
             {
                 alive.insert(alivehost.ip.clone());
-                println!("{} via {:?}", &alivehost.ip, &alivehost.detectihttp_method);
+                println!("{} via {:?}", &alivehost.ip, &alivehost.detection_method);
             }
         }
 
--- a/rust/src/alive_test/alive_test.rs
+++ b/rust/src/alive_test/alive_test.rs
@@ -378,13 +378,25 @@ impl Scanner {
     }
 
     pub async fn run_alive_test(&self) -> Result<HashSet<String>, AliveTestError> {
+        self.run_alive_test_streaming(None).await
+    }
+
+    pub async fn run_alive_test_streaming(
+        &self,
+        notify: Option<Sender<Host>>,
+    ) -> Result<HashSet<String>, AliveTestError> {
         // TODO: Replace with a Storage type to store the alive host list
         let mut alive = HashSet::<String>::new();
 
         if self.methods.contains(&AliveTestMethods::ConsiderAlive) {
             for t in self.target.iter() {
                 alive.insert(t.clone());
-                println!("{t} via {}", AliveTestMethods::ConsiderAlive)
+                println!("{t} via {}", AliveTestMethods::ConsiderAlive);
+                if let Some(tx) = &notify {
+                    // TODO: check what happens if there is no receiver (scan stop)
+                    // We should break the alive test too.
+                    let _ = tx.send(t.clone()).await;
+                }
             }
             return Ok(alive);
         };
@@ -404,14 +416,22 @@ impl Scanner {
         let send_handle = tokio::spawn(send_task(methods_c, trgt, timeout, tx_ctl));
 
         while let Some(alivehost) = rx_msg.recv().await {
+            let mut newly_found = None;
             if self.target.contains(&alivehost.ip) && !alive.contains(&alivehost.ip) {
                 alive.insert(alivehost.ip.clone());
                 println!("{} via {:?}", &alivehost.ip, &alivehost.detection_method);
+                newly_found = Some(alivehost.ip.clone());
             } else if let Some(Ok(dst)) = get_host_discovery_ipv6_net(&self.methods, &self.target)
                 && dst.contains(&alivehost.ip.parse::<std::net::Ipv6Addr>().unwrap())
             {
                 alive.insert(alivehost.ip.clone());
                 println!("{} via {:?}", &alivehost.ip, &alivehost.detection_method);
+                newly_found = Some(alivehost.ip.clone());
+            }
+            if let Some(host) = newly_found
+                && let Some(tx) = &notify
+            {
+                let _ = tx.send(host).await;
             }
         }
 
--- a/rust/src/alive_test/alive_test.rs
+++ b/rust/src/alive_test/alive_test.rs
@@ -391,9 +391,8 @@ impl Scanner {
                 alive.insert(t.clone());
                 println!("{t} via {}", AliveTestMethods::ConsiderAlive);
                 if let Some(tx) = &notify {
-                    // TODO: check what happens if there is no receiver (scan stop)
-                    // We should break the alive test too.
-                    let _ = tx.send(t.clone()).await;
+                    // if there is no receiver, we just stop here.
+                    tx.send(t.clone()).await.unwrap();
                 }
             }
             return Ok(alive);
@@ -414,20 +413,27 @@ impl Scanner {
         let send_handle = tokio::spawn(send_task(methods_c, trgt, timeout, tx_ctl));
 
         while let Some(alivehost) = rx_msg.recv().await {
-            let mut newly_found = None;
-            let mut newly_found_method: Option<AliveTestMethods> = None;
-            if self.target.contains(&alivehost.ip) && !alive.contains(&alivehost.ip) {
-                newly_found = Some(alivehost.ip.clone());
-                newly_found_method = Some(alivehost.detection_method.clone());
+            let (newly_found, newly_found_method) = if self.target.contains(&alivehost.ip)
+                && !alive.contains(&alivehost.ip)
+            {
+                (
+                    Some(alivehost.ip.clone()),
+                    Some(alivehost.detection_method.clone()),
+                )
             } else if let Some(Ok(dst)) = get_host_discovery_ipv6_net(&self.methods, &self.target)
                 && dst.contains(&alivehost.ip.parse::<std::net::Ipv6Addr>().unwrap())
             {
-                newly_found = Some(alivehost.ip.clone());
-                newly_found_method = Some(alivehost.detection_method.clone());
-            }
+                (
+                    Some(alivehost.ip.clone()),
+                    Some(alivehost.detection_method.clone()),
+                )
+            } else {
+                (None, None)
+            };
+
             if let Some(host) = newly_found {
                 if let Some(tx) = &notify {
-                    let _ = tx.send(host).await;
+                    tx.send(host).await.unwrap();
                 } else {
                     println!("{} via {:?}", &host, &newly_found_method.unwrap());
                 }
--- a/rust/src/alive_test/alive_test.rs
+++ b/rust/src/alive_test/alive_test.rs
@@ -385,10 +385,8 @@ impl Scanner {
         &self,
         notify: Option<Sender<Host>>,
     ) -> Result<HashSet<String>, AliveTestError> {
-        // TODO: Replace with a Storage type to store the alive host list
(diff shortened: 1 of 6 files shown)


rust/src/alive_test/alive_test.rs
```diff
@@ -378,13 +378,23 @@ impl Scanner {
...
+        let mut alive = HashSet::<String>::new();
+        if self.methods.contains(&AliveTestMethods::ConsiderAlive) || self.methods.is_empty() {
             for t in self.target.iter() {
                 alive.insert(t.clone());
-                println!("{t} via {}", AliveTestMethods::ConsiderAlive)
+                println!("{t} via {}", AliveTestMethods::ConsiderAlive);
+                if let Some(tx) = &notify {
+                    // TODO: check what happens if there is no receiver (scan stop)
+                    // We should break the alive test too.
+                    let _ = tx.send(t.clone()).await;
```
Tehforsch: So this says the sender will be `None` both in the normal alive test and for a stopped scanner, right? Maybe we should be able to differentiate these modes by using a custom enum or something
```
enum NotifyMode {
    Scanner(Sender), AliveTest, StoppedScan
}
```


Just writing this down, could very well turn out to be a bad idea as I go on reviewing

Tehforsch: Please use `.unwrap()` here instead of ignoring the result, otherwise this async task might continue when the corresponding scanner has panicked which isn't helpful and can make it harder to debug stuff

rust/src/alive_test/alive_test.rs
```diff
@@ -404,14 +414,24 @@ impl Scanner {
...
+                newly_found_method = Some(alivehost.detection_method.clone());
             } else if let Some(Ok(dst)) = get_host_discovery_ipv6_net(&self.methods, &self.target)
                 && dst.contains(&alivehost.ip.parse::<std::net::Ipv6Addr>().unwrap())
             {
+                newly_found = Some(alivehost.ip.clone());
+                newly_found_method = Some(alivehost.detection_method.clone());
+            }
+            if let Some(host) = newly_found {
+                if let Some(tx) = &notify {
+                    let _ = tx.send(host).await;
```
Tehforsch: Same as above, `.unwrap()` the result

rust/src/alive_test/alive_test.rs
```diff
@@ -404,14 +414,24 @@ impl Scanner {
...
-                alive.insert(alivehost.ip.clone());
-                println!("{} via {:?}", &alivehost.ip, &alivehost.detectihttp_method);
+                newly_found = Some(alivehost.ip.clone());
+                newly_found_method = Some(alivehost.detection_method.clone());
             } else if let Some(Ok(dst)) = get_host_discovery_ipv6_net(&self.methods, &self.target)
                 && dst.contains(&alivehost.ip.parse::<std::net::Ipv6Addr>().unwrap())
             {
+                newly_found = Some(alivehost.ip.clone());
+                newly_found_method = Some(alivehost.detection_method.clone());
+            }
```
Tehforsch: Personal preference: I would find this a bit harder to read if it was a large assignment like

```rust
let (newly_found, newly_found_method) = if ... { } else if { ... } else { ... }`
```

instead of the mutated variables, which I personally find a bit confusing

rust/src/alive_test/alive_test.rs
```diff
@@ -404,14 +414,24 @@ impl Scanner {
...
 
         while let Some(alivehost) = rx_msg.recv().await {
+            let mut newly_found = None;
+            let mut newly_found_method: Option<AliveTestMethods> = None;
             if self.target.contains(&alivehost.ip) && !alive.contains(&alivehost.ip) {
-                alive.insert(alivehost.ip.clone());
-                println!("{} via {:?}", &alivehost.ip, &alivehost.detectihttp_method);
+                newly_found = Some(alivehost.ip.clone());
+                newly_found_method = Some(alivehost.detection_method.clone());
             } else if let Some(Ok(dst)) = get_host_discovery_ipv6_net(&self.methods, &self.target)
```
Tehforsch, line 422: This else if means we ignore any result from `get_host_discovery_ipv6_net` - assuming that is intended but just wanted to clarify.

If that is intended, it seems more natural to make the method return something other than a result, but thats nitpicking

rust/src/scanner/running_scan.rs
```diff
@@ -35,6 +38,10 @@ pub struct RunningScan<S> {
     keep_running: Arc<AtomicBool>,
     status: Arc<RwLock<Status>>,
     notus: Option<NotusCtx>,
+    /// When set, hosts are attacked as they arrive on this channel
```
Tehforsch: Terminology nitpick: Can we rename this to "scanned"? I dont think we use "attacked" anywhere else

rust/src/scanner/running_scan.rs
```diff
@@ -35,6 +38,10 @@ pub struct RunningScan<S> {
...
+    /// (e.g. as they are confirmed alive by a concurrently running alive
+    /// test) instead of scanning `scan.targets` all at once.
```
Tehforsch: I'd get rid of the "instead of scanning `scan.targets` all at once", that seems like a reference to a status quo that will never be relevant anymore

rust/src/scanner/running_scan.rs
```diff
@@ -81,8 +89,82 @@ where
...
+            .iter()
+            .map(|t| (t.ip_addr().to_string(), t.clone()))
+            .collect();
+        let host_set: HashSet<Host> = host_by_ip.keys().cloned().collect();
+        let methods = scan.alive_test_methods.clone();
+        let capacity = host_by_ip.len().max(1);
+
+        // This channel is for sending a target to the running scan.
+        let (tx_target, rx_target) = mpsc::channel::<Target>(capacity);
+        // This channel receives alive host from the boreas
```
Tehforsch: I'd prefer getting rid of the "boreas" name since it is synonymous with "alive test" and we already use that a lot more. I think we now have the chance to at least make this consistent within the codebase before it entrenches itself. We do have some scanner preferences which contain boreas but well ...

rust/src/scanner/running_scan.rs
```diff
@@ -81,8 +89,82 @@ where
...
+        notus: Option<NotusCtx>,
+    ) -> RunningScanHandle {
+        let keep_running: Arc<AtomicBool> = Arc::new(true.into());
+        let status = Arc::new(RwLock::new(Status {
+            ..Default::default()
+        }));
+
+        let host_by_ip: HashMap<String, Target> = scan
+            .targets
+            .iter()
```
Tehforsch: I haven't thought this through in detail but it seems to me that it is a little confusing that we now have two ways to run the scanner, either we directly run `RunningScan` or we start a `RunningScan` and drive it from the outside using the code in this method - maybe we could make `host_feed` just `Sender<...>` instead of an `Option` - runs without alive test then just simply send all of `scan.targets` through the channel immediately and the code proceeds in the same way.

I think this is important because finding out where exactly the async tasks are that actually do the work in a running program is not super trivial and simplifying this type of code as much as possible makes it a lot easier.

rust/src/scanner/running_scan.rs
```diff
@@ -81,8 +89,82 @@ where
...
+            .map(|t| (t.ip_addr().to_string(), t.clone()))
+            .collect();
+        let host_set: HashSet<Host> = host_by_ip.keys().cloned().collect();
+        let methods = scan.alive_test_methods.clone();
+        let capacity = host_by_ip.len().max(1);
```
Tehforsch, line 76: Why the `.max(1)`? This will effectively cause backpressure on the sending end meaning that it has to wait for the receiver before sending another target / host. Is that what we want? If so we should maybe add a comment here, explaining why

rust/src/scanner/running_scan.rs
```diff
@@ -169,6 +255,28 @@ where
...
+/// Marks the given hosts as dead in `status`, once its `host_info` has
+/// been populated by [`RunningScan::update_status_at_beginning_of_run`].
+/// The alive test and the scheduling of the scan (which determines when
+/// `host_info` becomes available) run concurrently, so this waits (with
+/// a bounded number of retries) until `host_info` is set.
+async fn mark_hosts_dead_when_available(status: &Arc<RwLock<Status>>, dead_hosts: &[String]) {
+    const MAX_ATTEMPTS: usize = 200;
+    const RETRY_DELAY: std::time::Duration = std::time::Duration::from_millis(25);
+
+    for _ in 0..MAX_ATTEMPTS {
```
Tehforsch, line 231: I don't understand these max_attempts, what are we retrying? Is this for waiting until `host_info` becomes `Some(...)`?

Tehforsch (changes requested):

rust/src/alive_test/alive_test.rs
Tehforsch: See below, I think we shouldn't have this enum at all and just use a `Sender<...>` directly, without the surrounding `Option`

jjnicola: We can run alive tests via scannerctl, which doesn't requires a Sender. Therefore it is the option of None. We can leave as it is or remove the Option<> and introduce a channel for scannerctl as well.

rust/src/scanner/running_scan.rs
jjnicola: As mentioned in another comment, scannerctl uses the Alive Test scanner as well, but it doesn't initiates a mspc channel. Of course, we can simplify the scanner part and do it always via the alive test scanner even if no method is given and every hosts will be treat as alive.

rust/src/scanner/running_scan.rs
jjnicola: from the docu of mpsc::channel() 
```
/// The channel will buffer up to the provided number of messages.  Once the
/// buffer is full, attempts to send new messages will wait until a message is
/// received from the channel. The provided buffer capacity must be at least 1.
```

jjnicola: So, if for some reason there is no host in the target list (e.g. none has been resolved) we don't panic passing 0 capacity to the channel

rust/src/scanner/running_scan.rs
jjnicola: Yes, we must wait for status.host_info to be initialized by the concurrent task before we can write the dead hosts.

rust/src/scanner/running_scan.rs
Tehforsch: What I meant is, I think the code becomes slightly easier if we always use a sender. That doesn't mean running the alive test scanner even on scannerctl runs, but simply initializing a sender and dumping the entire `scan.targets` into it.

This way the scanner itself doesn't have to differentiate between these two modes, which fulfills separation of concerns better.

rust/src/scanner/running_scan.rs
Tehforsch: Oh of course, my bad - I still always confuse what `.max` does for some reason, confused it with `.min`.

rust/src/alive_test/alive_test.rs
```diff
@@ -404,14 +413,31 @@ impl Scanner {
         let send_handle = tokio::spawn(send_task(methods_c, trgt, timeout, tx_ctl));
 
         while let Some(alivehost) = rx_msg.recv().await {
-            if self.target.contains(&alivehost.ip) && !alive.contains(&alivehost.ip) {
-                alive.insert(alivehost.ip.clone());
-                println!("{} via {:?}", &alivehost.ip, &alivehost.detectihttp_method);
+            let (newly_found, newly_found_method) = if self.target.contains(&alivehost.ip)
```
Tehforsch: Irrelevant nit, feel free to ignore:

Slightly cleaner if you write return `Option<(A, B)>` instead of `(Option<A>, Option<B>)` here

rust/src/scanner/running_scan.rs
Tehforsch: This seems extremely fragile to me. Sure it's probably going to work most of the time, but if it doesn't it will be very annoying to debug. Why can we not initialize this field immediately? We have write access so we could do it here, or is there any special information that the other task has that is needed for initialization?

rust/src/scanner/mod.rs
```diff
@@ -76,8 +76,11 @@ where
         let loader = self.loader.clone();
         let function_executor = self.function_executor.clone();
         let id = scan.scan_id.clone();
+        // TODO: An empty test method should never happen, it should default to icmp.
+        // Is this already happening? check if this is dead code
```
Tehforsch: Outdated comment?

rust/src/scanner/running_scan.rs
```diff
@@ -35,6 +38,10 @@ pub struct RunningScan<S> {
     keep_running: Arc<AtomicBool>,
     status: Arc<RwLock<Status>>,
     notus: Option<NotusCtx>,
+    /// When set, hosts are scanned as they arrive on this channel
+    /// (e.g. as they are confirmed alive by a concurrently running alive
+    /// test).
```
Tehforsch: Rephrase now that this is always set?

rust/src/scanner/running_scan.rs
```diff
@@ -163,8 +132,12 @@ where
     }
 
     async fn run(mut self) -> Result<(), Error> {
-        let host_feed = self.host_feed.take();
-        let runner = match self.make_runner(host_feed).await {
+        // to avoid the option envelop for the host_feed receiver
+        // we need to replace the receiver with dummy one.
+        let (_dummy_tx, dummy_rx) = mpsc::channel::<Target>(1);
+        let rx = std::mem::replace(&mut self.host_feed, dummy_rx);
```
Tehforsch: This seems sketchy. Why are we doing this here instead of initializing the `host_feed` with the proper sender immediately? This is the proper scanner path, so there can't be a dummy sender here? Ignoring the sending end of a channel is bound to stall the receiving end

Tehforsch: Ok I just understood this, the point is to move the `host_feed` out of `self` so we can move it into the `ScanRunner`. But `self.host_feed` is never used except for excactly this purpose, so I think removing it from the `RunningScan` struct and passing it directly into the `run` method will make things simpler.

rust/src/scanner/scan_runner.rs
```diff
@@ -46,19 +49,25 @@ pub struct ScanRunner<'a, S> {
...
+    // Otherwise, uses scan.target.
+    host_feed: Receiver<Target>,
 }
 
 impl<'a, S> ScanRunner<'a, S>
 where
     S: ContextStorage,
 {
+    /// This method uses the alive host received via the channel in host_feed
+    /// in the mean they are found alive instead of the targets in scan.target
```
Tehforsch: I don't understand this comment

rust/src/scanner/scan_runner.rs
```diff
@@ -85,47 +95,61 @@ where
...
         // If this is changed, make sure to uphold the scheduling requirements in the
         // new implementation.
-        stream::unfold(data, move |mut data| async move {
-            match data.next() {
-                Some((stage, vt, param, host, ports, scan_id, notus)) => {
+        stream::unfold(state, move |(mut host_feed, mut queue)| {
+            // Cloning here (a cheap borrow of the outer captured
+            // `concurrent_vts`) instead of moving it directly into the
+            // `async move` block below keeps `concurrent_vts` available
+            // across repeated calls of this `FnMut` closure.
```
Tehforsch: Thanks copilot :D Can you compress this comment a bit. This is pretty standard practice before closures, so in my opinion it can also go entirely

rust/src/scanner/scan_runner.rs
```diff
@@ -85,47 +95,61 @@ where
...
+            async move {
+                loop {
+                    if queue.is_empty() {
+                        match host_feed.recv().await {
+                            Some(host) => {
+                                enqueue_host(&mut queue, &host, &concurrent_vts);
+                                // A host may end up with no VTs to run (e.g.
+                                // empty schedule); keep pulling hosts until we
+                                // either have work to do or the feed is
+                                // exhausted.
```
Tehforsch: I don't understand this comment - if I understand the code correcly we simply await new hosts and if `recv().await` returns None the channel is closed, meaning there are no more hosts incoming.

Tehforsch: Ok nvm got it now, but this still seems overly complicated? Why dont we do something like

```
loop { 
  if queue.is_empty() {
    // check receiver, queue new hosts
  }
  if let Some(host) = queue.pop_front() {
    // do the rest
  }
}
```
Then we can save ourselves the non-linear control flow of the continue and this entire comment

rust/src/scanner/tests.rs
```diff
@@ -309,8 +310,10 @@ async fn run(
 
     let scheduler = Scheduler::new(storage.clone());
     let schedule = scheduler.execution_plan(&scan.vts).await?;
-    let interpreter: ScanRunner<Arc<InMemoryStorage>> =
-        ScanRunner::new(&storage, &loader, &executor, schedule, &scan, &None)?;
+    let (_dummy_tx, dummy_rx) = mpsc::channel::<Target>(1);
+    let interpreter: ScanRunner<Arc<InMemoryStorage>> = ScanRunner::new(
+        &storage, &loader, &executor, schedule, &scan, &None, dummy_rx,
+    )?;
```
Tehforsch: I don't think this works, but it's irrelevant because this test has been removed on current main

rust/src/scannerctl/execute/mod.rs
```diff
@@ -137,8 +138,13 @@ async fn scan(args: ScanArgs) -> Result<(), CliError> {
                 ProductLoader::new(false, Loader::from_feed_path(path)),
             )))),
         });
-        let runner: ScanRunner<Arc<InMemoryStorage>> =
-            ScanRunner::new(&storage, &loader, &executor, schedule, &scan, &notus).unwrap();
+
+        let (_dummy_tx, dummy_rx) = mpsc::channel::<Target>(1);
```
Tehforsch: Why is this unused? We pass in a dummy_rx but we should actually send the targets through, otherwise nothing should happen

Tehforsch (changes requested):

rust/src/alive_test/alive_test.rs
jjnicola: I will leave it as it for now. I probably have to do some changes in the filters here. but will be done in a next PR

rust/src/alive_test/alive_test.rs
jjnicola: done

rust/src/scanner/running_scan.rs
jjnicola: Added a TODO. I will leave it as it is for now

rust/src/scanner/scan_runner.rs
```diff
@@ -85,47 +93,52 @@ where
...
-                        &notus,
-                    )
-                    .await;
-                    Some((result, data))
+        stream::unfold(state, move |(mut host_feed, mut queue)| {
+            let concurrent_vts = concurrent_vts.clone();
+            async move {
+                loop {
+                    if queue.is_empty() {
+                        if let Some(host) = host_feed.recv().await {
```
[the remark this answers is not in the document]
jjnicola: Done, Now it is inside an Arc<>

rust/src/scannerctl/execute/mod.rs
```diff
@@ -137,8 +140,51 @@ async fn scan(args: ScanArgs) -> Result<(), CliError> {
...
+            // Dropping `tx_target` here closes the channel, signalling to
+            // the running scan that no further hosts will arrive.
+        });
+
+        tokio::spawn(async move {
+            let alive_scanner = BoreasScanner::new(host_set.clone(), methods, None);
+            let alive = alive_scanner
+                .run_alive_test_streaming(Some(tx_host))
+                .await
+                .unwrap_or_default();
```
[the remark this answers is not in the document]
jjnicola: If fails, it returns an empty Hashset per default. It is ok so.

rust/src/models/host_info.rs
```diff
@@ -59,6 +59,16 @@ impl HostInfo {
         self.remaining_vts_per_host.clear();
     }
 
+    // Used by openvasd scanner when using Boreas.
+    pub fn mark_hosts_dead<'a>(&mut self, hosts: impl Iterator<Item = &'a str>) {
+        for host in hosts {
+            if self.remaining_vts_per_host.remove(host).is_some() {
+                self.dead += 1;
+                self.queued -= 1;
+            }
```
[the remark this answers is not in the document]
jjnicola: fixed in another PR

rust/src/alive_test/alive_test.rs
```diff
@@ -404,14 +413,25 @@ impl Scanner {
...
-            if self.target.contains(&alivehost.ip) && !alive.contains(&alivehost.ip) {
-                alive.insert(alivehost.ip.clone());
-                println!("{} via {:?}", &alivehost.ip, &alivehost.detectihttp_method);
+            let newly_found = if self.target.contains(&alivehost.ip)
+                && !alive.contains(&alivehost.ip)
+            {
+                Some((alivehost.ip.clone(), alivehost.detection_method.clone()))
             } else if let Some(Ok(dst)) = get_host_discovery_ipv6_net(&self.methods, &self.target)
                 && dst.contains(&alivehost.ip.parse::<std::net::Ipv6Addr>().unwrap())
             {
```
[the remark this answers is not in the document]
jjnicola: Will be done in another PR

rust/src/alive_test/alive_test.rs
```diff
@@ -378,13 +378,22 @@ impl Scanner {
...
+        let mut alive = HashSet::<String>::new();
+        if self.methods.contains(&AliveTestMethods::ConsiderAlive) || self.methods.is_empty() {
             for t in self.target.iter() {
                 alive.insert(t.clone());
-                println!("{t} via {}", AliveTestMethods::ConsiderAlive)
+                println!("{t} via {}", AliveTestMethods::ConsiderAlive);
+                if let Some(tx) = &notify {
+                    // if there is no receiver, we just stop here.
+                    tx.send(t.clone()).await.unwrap();
+                }
```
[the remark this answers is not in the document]
jjnicola: Done

rust/src/scanner/running_scan.rs
jjnicola: Done

rust/src/scanner/scan_runner.rs
jjnicola: done

rust/src/scannerctl/execute/mod.rs
jjnicola: Implement the right communication, but the scanner in scannerctl didn't work before. So, I added a TODO, since it must be fixed. The original issue in the scanner doesn't belong to this PR

rust/src/scanner/tests.rs
jjnicola: test doesn't exist anymore

Tehforsch (approved):
I didn't mean to say that scannerctl should do its own alive test checking, but if that's what we want then its alright

PR #2349, opened 2026-08-17
Handle malformed notus entries more gracefully.

Tehforsch:
Jira: SC-1701

### DIFF
--- a/rust/src/notus/tests.rs
+++ b/rust/src/notus/tests.rs
@@ -5,7 +5,7 @@
 use std::path::{Path, PathBuf};
 
 use crate::{
-    models::{FixedPackage, FixedVersion, Specifier},
+    models::{FixedVersion, Specifier},
     nasl::Loader,
     notus::{ProductLoader, error::Error, notus::Notus},
 };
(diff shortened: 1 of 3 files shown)


rust/src/notus/tests.rs
```diff
@@ -89,32 +89,15 @@ fn test_err_package_parse_error() {
...
-            && module.is_none()
-    ));
-    assert!(
-        err.to_string().contains(
-            "malformed entry in vulnerability data file debian_10_product_parse_err.notus"
-        )
-    );
+    let vulnerable = &results["1.3.6.1.4.1.25623.1.1.7.2.2023.10089729899100"];
+    assert_eq!(vulnerable.len(), 1);
+    assert_eq!(vulnerable[0].name, "gitlab-ce");
```
Kraemii, line 102: In general I am not sure about this test. I does not test, if an invalid product entry is skipped, but only if the file in question finds the the requested vulnerability. But as the notus part will be deprecated anyways and the fix looks good we can leave it as it is.

Kraemii (approved):
Tested the fix manually, looks got. Only a nit about the test :)

rust/src/notus/tests.rs
Tehforsch: Yeah you're right, I wasn't sure what else to do without adding a lot of machinery and this at least tests the "external" behavior that we want - other vulnerabilities are still found. So yeah, not fully convinced either, but probably OK for a soon-to-be-deprecated feature

PR #2353, opened 2026-08-18
Fix: queued counter in host_info.

jjnicola:
Fix: queued counter in host_info.

When there is no host left in the queued, the sql query will store the old value, because of the NULLIF(?,0) if the value is equal to zero, what is what happends when the scan finishes and there is no more host in the queue to be scanned.

SC-1747

### DIFF
--- a/rust/src/openvasd/database/sqlite/state_change.rs
+++ b/rust/src/openvasd/database/sqlite/state_change.rs
@@ -182,7 +182,7 @@ impl ScanStateController {
         end_time      = COALESCE(?, end_time),
         host_dead     = COALESCE(NULLIF(?, 0), host_dead),
         host_alive    = COALESCE(NULLIF(?, 0), host_alive),
-        host_queued   = COALESCE(NULLIF(?, 0), host_queued),
+        host_queued   = ?,
         host_excluded = COALESCE(NULLIF(?, 0), host_excluded),
         host_all      = COALESCE(NULLIF(?, 0), host_all),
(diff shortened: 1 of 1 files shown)


rust/src/openvasd/database/sqlite/state_change.rs
```diff
@@ -182,7 +182,7 @@ impl ScanStateController {
         end_time      = COALESCE(?, end_time),
         host_dead     = COALESCE(NULLIF(?, 0), host_dead),
         host_alive    = COALESCE(NULLIF(?, 0), host_alive),
-        host_queued   = COALESCE(NULLIF(?, 0), host_queued),
+        host_queued   = ?,
         host_excluded = COALESCE(NULLIF(?, 0), host_excluded),
         host_all      = COALESCE(NULLIF(?, 0), host_all),
         status        = COALESCE(NULLIF(NULLIF(?, 'stored'), 'requested'), status)
```
Tehforsch, line 188: Am a little confused about the logic here - why is it OK for the other fields to remain as `COALESCE(NULLIF(?, 0), ...)` but not for queued?

Tehforsch (approved):

rust/src/openvasd/database/sqlite/state_change.rs
jjnicola: All fields will increase their values, but not queued. `queued` is the only one which decrease their value and reaches zero.

Tehforsch: That makes sense, alright

PR #2359, opened 2026-08-24
Fix: Send compact port list to openvas

### LINKED ISSUE #2223: openvasd expands port ranges into per-port list → nmap E2BIG ("Argument list too long") on full 1-65535 scans

Kraemii:
As we use the `nmap` executable for scanning ports, the list of ports is given to it. `openvasd` generates a large comma-separated port list for `openvas`. This leads to the issue, that the argument list is too long and nmap fails to execute. This results in all ports being unscanned. In combination with the setting unscanned_closed this leads to all ports being closed.

To test this:

Use the following scan config:
```
{
  "target": {
    "hosts": [
      "127.0.0.1"
    ],
    "ports": [
      {
        "protocol": "tcp",
        "range": [
          {
            "start": 1,
            "end": 65535
          }
        ]
      }
    ]
  },
  "vts": [
    {
      "oid": "1.3.6.1.4.1.25623.1.0.10394"
    }
  ]
}
```

Run a scan with this scan config via openvas before and after the patch. Before the patch the following log should be visible in the openvasd logs:
```
:WARNING:2026-08-24 09h39.55 utc:24265:e83368e1-1236-4b0d-8e45-734da79e508d: nasl_pread: Failed to execute child process “nmap” (Argument list too long)
```
After the patch this log should not be shown anymore.

Closes https://github.com/greenbone/openvas-scanner/issues/2223
Jira: SC-1753, SC-1756

### DIFF
--- a/rust/src/models/port.rs
+++ b/rust/src/models/port.rs
@@ -247,4 +247,30 @@ mod tests {
             Some("T:22-32,80,1000,2000-2004,U:22,24-40,1000,5060".to_string())
         );
     }
+
+    #[test]
+    fn test_very_deranged_ports() {
+        let mut ports = vec![];
+        for i in 1..=1000 {
+            ports.push(Port {
+                protocol: Some(Protocol::TCP),
+                range: vec![PortRange {
+                    start: i,
+                    end: None,
+                }],
+            });
+            ports.push(Port {
+                protocol: Some(Protocol::TCP),
+                range: vec![PortRange {
+                    start: 65535 - i + 1,
+                    end: None,
+                }],
+            });
+        }
+
+        assert_eq!(
+            ports_to_openvas_port_list(ports),
+            Some("T:1-1000,64536-65535".to_string())
+        );
+    }
 }
--- a/rust/src/models/port.rs
+++ b/rust/src/models/port.rs
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: 2023 Greenbone AG
 //
 // SPDX-License-Identifier: GPL-2.0-or-later WITH x11vnc-openssl-exception
-use std::{fmt::Display, str::FromStr};
+use std::{collections::BTreeSet, fmt::Display, str::FromStr};
 
 /// Represents a port representation for scanning.
 #[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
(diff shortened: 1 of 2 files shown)


ArnoStiefvater (commented):
I feel like a few more unit tests with some more exotic port lists could be good.

Kraemii:
Added some overlapping ports. Not sure what exactly is meant with exotic tbh

rust/src/models/port.rs
```diff
@@ -94,58 +94,73 @@ impl AsRef<str> for Protocol {
...
-        // Add range
-        if let Some(end) = end {
-            for p in start..=end {
-                list.push_str(p.to_string().as_str());
-                list.push(',');
+    // Collect all ports per protocol into a sorted, deduplicated set so that
+    // contiguous ports, including many single ports, can be compacted into
+    // ranges. This keeps the resulting string as short as possible. This is
+    // necessary because of the call to nmap, which fails to execute with too
+    // many arguments given.
```
Tehforsch: "The call to nmap" is not very clear in a codebase that never calls nmap ... Clarify that this is an issue in our default feed?

rust/src/models/port.rs
```diff
@@ -94,58 +94,73 @@ impl AsRef<str> for Protocol {
...
+            prev = Some(port);
+        }
+        if let (Some(s), Some(e)) = (start, prev) {
+            parts.push(if s == e {
+                s.to_string()
+            } else {
+                format!("{s}-{e}")
+            });
         }
+        parts.join(",")
```
Tehforsch, line 130: If we're already slightly overengineering this, why not do it "properly" so it actually gives the most compact set? This doesn't cover the case where the user specifies ports in a truly deranged way like `1,65535,2,65534,...`

I also find it a little awkward that this returns a string when we have `Vec<PortRange>` available which does exactly this.

Tehforsch (changes requested):

rust/src/models/port.rs
Kraemii: I changed the comment to be as clear as possible, why we actually do something overengineered like this

rust/src/models/port.rs
Kraemii: 1. I added a test with a very deranged port list.

2. I still see an issue if the user is giving a port list with every other port like `1, 3, 5, ...`, as this cannot be compacted at all an will still lead to a failure in nmap. Not sure if we even want to handle that, like giving a response that the port list contains too many single ports.

3. The string is especially created to put it into redis. Afterwards it is directly used by NASL to call nmap with this string

Tehforsch: Yeah, number 2 is going to be hard to solve, let's just hope this unlikely case doesnt happen for now.

For number 3: I know, but I think its nicer to do this in a pure way without going into string land. But then again this is one-off logic and its probably fine as it is.

rust/src/models/port.rs
Tehforsch: i dont see it

Tehforsch (approved):
Thanks!

